PR #191, opened 2015-03-11
Fixed plot permissions never working.

romanalexander:
There is no better way to do this safely without reflection.


boy0001:
Have you verified that this is the root of the problem rather than the permission caching?

romanalexander:
I didn't check, but logically I'd imagine the behavior is still bugged due to the way most servers set up their permission inheritance.

Considering the following player's permissions:
* plots.plot.1
* plots.plot.2
* plots.plot.3

and that the player has claimed no plots yet.

The bugged behavior will return 1, the legacy behavior should return 3.

boy0001:
Yes, I am aware that it doesn't return the maximum number of plots a player is allowed. It will return the first permission they have above their current number of plots. Shouldn't this still work for plot claiming?

PR #275, opened 2015-04-15
Allow falling block check during piston extend to be configurable

Byteflux:
The default behavior seems to be interfering with a particular design of TNT cannons which use cobwebs on pistons, with sand caught in the cobwebs as a method of launching large amounts of sand entities at the same time.

~~This patch allows pistons to be extended even if there are falling block entities directly in front of the piston as long as there is also a cobweb. I didn't fully grasp the scope of the original fix so if I am messing something up please let me know.~~ Updated to use a configuration option instead, this seemed like the better choice since there could be other edge cases that have yet to pop up. Cannoning is the primary purpose of my plot server so there's bound to be lots of cases of pistons being used with falling blocks.


Byteflux:
This is the type of cannon I'm referring to: http://i.imgur.com/dCsqTzA.gif 

Pretty much any TNT cannon that uses pistons and cobwebs to shoot sand.

PR #305, opened 2015-05-09
Get cached PlotPlayer

Byteflux:
I ran into some issues with PlotPlayer's metadata always being empty. Turns out it was because `BukkitUtil.getPlayer(Player)` was always returning a new PlotPlayer object even for the same player.


boy0001:
Thanks for noticing things.

PR #389, opened 2015-07-11
30 secs is 600 ticks. Don't lie.

LaxWasHere:


Sauilitired:
>:[ What if we like lying? Anyways, merged and all ^_^

PR #758, opened 2015-11-28
fix players being counted for entity caps

manuelgu:
There might be a better way. I've changed the general `countEntities` method as it is used to calculate the amount of entities on a plot.

**REASON**
Players do not bypass the entity cap limit for a plot


boy0001:
The counting occurs here:
https://github.com/manuelgu/PlotSquared/blob/master/src/main/java/com/plotsquared/bukkit/util/BukkitChunkManager.java#L1018
where it should be ignoring players.

That part above is just for deciding which algorithm to use.

PR #785, opened 2015-12-13
Fix schematic pasting for plots with offsets

UltraX3:
Might be bad code but it works. Not really familiar with all the classes, I'm sure there is a better Configuration class that has methods for accessing plot heights.


boy0001:
Depending on the generator, that value may not be present and it will throw and error.
e.g. AdvPlots (GridPlotWorld), IslandPlots (SquarePlotWorld), PlotSquaredMG (SquarePlotWorld).

You would need to cast to the type of plot world to get the height.
```Java
PlotWorld pw = plot.getWorld();
if (pw instanceof ClassicPlotWorld) {
    y_offset = ((ClassicPlotWorld) pw).PLOT_HEIGHT;
} else {
    y_offset = MainUtil.getHeighestBlock(plot.world, region.minX + 1, region.minZ + 1);
}
```

PR #849, opened 2016-01-28
Properly organised the layout and modulation.

SilverCory:
This project is really badly organised.

This still won't compile but I'll look into it, and I hope someone who knows the structure of the project properly does too.

The lines of separation between API, Bukkit and Sponge within this project have become blurred, resulting in classes that should be in the API referencing directly to Bukkit and Bukkit references to API so this is where the problem is at.


Sauilitired:
The API separation is indeed quite blurred, but the API isn't as simple as just being an API, and should imo be referred to as core instead. Just a minor thing.

SilverCory:
Yeah I agree, just changed that.

MattBDev:
There are a few issues in the Pom files such as the api dependency needs to be changed to core. There are also duplicated dependencies. I am not sure if the dependency issues are related to optional but I think removing the optionals for now would be best so this can be merged as soon as possible. More of my code cleanup work can continue as soon as this is merged. I'll update this PR myself if I don't have a reply soon.

Bukkit/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+                <exclusion>
+                    <groupId>org.bukkit</groupId>
+                    <artifactId>bukkit</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        
+        <dependency>
+			<groupId>com.intellectualcrafters</groupId>
+			<artifactId>api</artifactId>
```
MattBDev: This should be changed to core

MattBDev: This should be changed to core

SilverCory:
Good good, my bad.
As for the dupe dependencies, I left them in for reference, thought I cleaned up the master pom.
Great to see others help organise. It really annoyed me. 
On 5 Feb 2016 5:13 pm, Matt <notifications@github.com> wrote:There are a few issues in the Pom files such as the api dependency needs to be changed to core. There are also duplicated dependencies. I am not sure if the dependency issues are related to optional but I think removing the optionals for now would be best so this can be merged as soon as possible. More of my code cleanup work can continue as soon as this is merged. I'll update this PR myself if I don't have a reply soon.

MattBDev:
I've looked it over throughly and I think once those changes are made a merge can happen. Cleaning up the project is something I am focused on at the moment. This reorganization would also allow for more Sponge integration to continue which is important for further cleanup. 
@boy0001 and @Sauilitired I encourage you to merge this right away as soon as the final changes are made.

Sauilitired:
@MattBDev: Sure thing. Actually, thanks for being helpful rather than holding a grudge against us, I really appreciate it.

SilverCory:
Just a note, this won't compile still, too much non-api stuff in the api and vice versa.

MattBDev:
I think some input from @Sauilitired and @boy0001 is needed for this PR to continue. A lot of conflicts and to resolve them it requires very critical decisions.

SilverCory:
:+1:

MattBDev:
I have been working on abstracting things myself for the time being and have reduced the number of conflicts in half.

MattBDev:
The best way to make the Sponge port work though is to make it dependent on SpongeForge instead of all implementations of Sponge.(at least until spigot updates their block id support)

boy0001:
Separation is done with packages. Initially with just the `com.intellectualcrafters` package, now there's also `com.plotsquared`. I believe @Sauilitired started moving everything to the `com.plotsquared` package, but that never happened since it would break compatibility with any plugin using the PlotAPI class, or some other reasons.

The current layout is `com.plotsquared.bukkit` has the bukkit only classes, `com.plotsquared.sponge` has the sponge classes, and everything else "should" be core stuff.

A potential benefit of having only package separation is refactoring / cleanup is possibly simpler with it in the same source.

On the note of this / any other refactor / cleanup breaking compatibility, it may as well if it means moving this project forward.

SilverCory:
The benefits of separating it into different projects is if something gets screwed up to sponge then bukkit won't be affected. Cleaner builds and a defined separation between the packages. 
Also the ability to skip building of a single project if nessacairy.
It also resolves the unorthodox way that you're compiling the different jars. And makes actually using them as a developer within maven,  a hell of a lot easier. 
The separation via packages has failed, there's bukkit code in core.
On 10 Feb 2016 7:10 pm, Jesse Boyd <notifications@github.com> wrote:Separation is done with packages. Initially with just the com.intellectualcrafters package, now there's also com.plotsquared. I believe @Sauilitired started moving everything to the com.plotsquared package, but that never happened since it would break compatibility with any plugin using the PlotAPI class, or some other reasons.

The current layout is com.plotsquared.bukkit has the bukkit only classes, com.plotsquared.sponge has the sponge classes, and everything else "should" be core stuff.

A potential benefit of having only package separation is refactoring / cleanup is possibly simpler with it in the same source.

On the note of this / any other refactor / cleanup breaking compatibility, it may as well if it means moving this project forward.

boy0001:
Well I did some cleaning that moves all the bukkit code to the correct package. (ignoring the PlotAPI class)
c83378a91b52237e677c63d768c54c0f8d23ff00
893c2b30884285b9aa1f01f645bc132519c2c260

As far as it being in the core, it's always been in the core, I had only moved out the stuff which sponge had implemented at the time. The rest wouldn't have done anything given that sponge version didn't use / support those classes. Now that sponge is in beta, I guess it's possible to port the whole plugin rather than just a subset to the platform.

MattBDev:
It may even be worth switching to gradle so we have access to forge on the sponge port and can directly manipulate the NBT for downloads and schematics. Worldedit is doing this for their sponge port.

boy0001:
WorldEdit ported to Forge long before sponge even started. Since Sponge(Forge) is just a forge plugin, WorldEdit didn't need to be ported.

I'm not sure if a forge jar is necessary. For the bukkit mass block changes it uses reflection to stay compatible between minor version bumps. It also uses net.minecraft.server which if I'm not mistaken would be present on both platforms albeit with a slightly different import.

MattBDev:
Fixed with the switch to gradle.

PR #869, opened 2016-02-13
Fix mismatch of brackets

manuelgu:
This could potentially cause problems because the permission check is not correct


boy0001:
Nope

manuelgu:
Care to elaborate? All the other classes do not have these extra brackets around the owner check.

boy0001:
The brackets are redundant, it compiles to the same thing either way.

manuelgu:
Why not merge then when they are "redundant"?

boy0001:
mostly in the same way an extra space somewhere is redundant. cbs

PR #874, opened 2016-02-15
Gradle

### LINKED ISSUE #864: Migrate Project to Gradle
Gradle will provide cleaner and faster builds and allow easier and simpler compatibility with a Sponge port.

MattBDev:
Fixes #864 and #849
Allows for easier compatibility with Sponge and cleaner directory structure.
The plugin compiles again, even the Sponge version (but a lot is broken in Sponge).

Waiting for the okay from @Sauilitired and @boy0001 before merging. 
The project can be compiled using the command "gradle build" if Gradle is installed locally.
It also has plugins for Eclipse and IntelliJ workspaces.


Byteflux:
Please no. The Bukkit community likes Maven. You don't need Gradle to be compatible with Sponge. Many have already expressed discontent for switching to Gradle in #PlotSquared as soon as we saw these commits/PR getting posted in chat.

MattBDev:
@Byteflux Unless you plan on modifying the source code I don't see a big issue with switching to Gradle. The project would still be compatible with maven so people who use maven for add-ons can still do so. The benefit of switching to Gradle is not needing any unnecessary reflection on the Sponge version. Being able to directly access minecraft code which is not "NMS" is helpful.
Gradle is much neater and cleaner when it comes to multi-project builds when compared to maven as well. I am happy to discuss this further in IRC or here.

Sauilitired:
I'm completely fine with this :)

Sauilitired:
Actually, angry as I am; if this was to make the bukkit community angry, then I'd see that as a perk ;)

boy0001:
I'd like to leave this open for a while so past / potential contributors may express their view. We can still work on the sponge port before this is merged.
@MattBDev 
https://github.com/IntellectualSites/PlotSquared/pulls?q=is%3Apr+author%3AByteflux+is%3Aclosed
Also, It's only the block changes that use nms. It would probably be easier reusing the same nms code for both. That means a bit of reflection, which is fine. For 1.8, this is done at the chunk level, so the overhead is insignificant.

@Sauilitired 
I would not view that as a perk. fyi, PlotSquared is a bukkit plugin.

Sauilitired:
@boy0001 So I'm <painfully> are.

Sauilitired:
@boy0001 So I'm *painfully* aware.

MattBDev:
@Sauilitired I feel your pain. :laughing:

minecrafter:
> Allows for easier compatibility with Sponge
Yes, Gradle is the standard for Sponge. No, it's not required to be compatible with Sponge.

> cleaner directory structure.
The _last_ thing that needs to happen to get a cleaner directory structure for the project is moving to Gradle.

Sauilitired:
@minecrafter It won't really affect the directory structure a whole lot, and if it does make sponge development any easier, I'd say that it's worth it. It honestly doesn't affect us all that much ;D

MattBDev:
@minecrafter It makes the structure a lot neater actually.

MattBDev:
Although reflection is an option, I don't think we really want to use it. Using gradle allows us to avoid that while keeping the jar small. The current way of handling numeric block id's increases the jar file size. We can keep that down by using gradle.
There are no major issues with gradle.
If no one has any real reason not to switch, I am going to update this and merge it soon.

minecrafter:
@MattBDev what's wrong with Maven modules?

My complaints revolve around "why?" The three reasons provided don't convince me.

MattBDev:
@minecrafter The two major reasons of switching to gradle are: Easier Sponge Compatibility (This is extremely important since Spigot API SUCKS and Sponge Servers are expected to increase once more plugins are ready). It's much easier to structure the gradle project than the maven multi-module project. Instead of 150 line pom.xml we can have a 26 line build.gradle file for the Bukkit Module.
It's easier to manipulate minecraft code directly than with reflection for Sponge. We have that ability when we switch to gradle. There is a whole page about Maven VS Gradle on http://gradle.org/maven_vs_gradle/ . Granted we won't use everything there but there are advantages to having gradle. In the long run, people making add-ons can still use maven to depend on the project.

minecrafter:
> Easier Sponge Compatibility (This is extremely important since Spigot API SUCKS and Sponge Servers are expected to increase once more plugins are ready).

Hogwash.

> It's much easier to structure the gradle project than the maven multi-module project. Instead of 150 line pom.xml we can have a 26 line build.gradle file for the Bukkit Module.

You do have a point here.

> It's easier to manipulate minecraft code directly than with reflection for Sponge. We have that ability when we switch to gradle.

You already have it with Maven?!?

MattBDev:
@minecrafter We don't have that already with maven.

minecrafter:
I'm trying to say that it was possible with Maven. Apologies.

MattBDev:
@minecrafter It actually isn't already possible with maven. We only have access to the SpongeAPI and not access to minecraft  code. we need the access to the MC code.

minecrafter:
In that case you do have a point. My concerns are dropped.

MattBDev:
@Byteflux I still haven't heard any comments from you regarding the switch. I haven't merged yet because of your disapproval. I'd love to hear more from you.

PhanaticD:
+1 for maven

MattBDev:
@PhanaticD That isn't a convincing argument.

RoboMWM:
While I do see the access to MC code being a valid point, the point of moving to Gradle "because it's better with sponge" is clearly subjective regarding this project. [I've already outlined my points elsewhere why I won't be moving to Sponge anytime soon.](https://gist.github.com/RoboMWM/03a95b50db1e1d6c32db)

MattBDev:
@RoboMWM Sponge is not ready and everyone is aware of that. WorldEdit has two implementations going to support Sponge natively. They use a feature in gradle to support SpongeForge. WorldEdit also used to be a maven project before they switched to Gradle and how much easier it is to maintain. For developers working on Linux (like I do a lot and so does the worldedit devs) gradle is much easier to install than maven is. This is not a debate about switching to sponge, this a debate on why we should switch.
Add-on developers are NOT impacted by this switch and will not even notice a difference!

MattBDev:
I'd like to point out the following code:
```java
private static void initBlockCache() {
        try {
            PS.debug("Caching block id/data: Please wait...");
            stateArray = new BlockState[Character.MAX_VALUE];
stateMap = new HashMap<>();
Class<?> classBlock = Class.forName("net.minecraft.block.Block");
            Class<?> classBlockState = Class.forName("net.minecraft.block.state.IBlockState");
            Method[] blockMethods = classBlock.getDeclaredMethods();
            Method methodGetByCombinedId = null;
            for (Method method : blockMethods) {
                Class<?> result = method.getReturnType();
                Class<?>[] param = method.getParameterTypes();
                int paramCount = param.length;
                boolean isStatic = Modifier.isStatic(method.getModifiers());
                if (methodGetByCombinedId == null) {
                    if (isStatic && result == classBlockState && paramCount == 1 && param[0] == int.class) {
                        methodGetByCombinedId = method;
                        continue;
                    }
                }
            }
            for (int i = 0; i < Character.MAX_VALUE; i++) {
                try {
                    BlockState state = (BlockState) methodGetByCombinedId.invoke(null, i);
                    if (state.getType() == BlockTypes.AIR) {
                        continue;
                    }
                    int id = i & 0xFFF;
                    int data = i >> 12 & 0xF;
                    PlotBlock plotBlock = new PlotBlock((short) id, (byte) data);
                    stateArray[i] = state;
                    stateMap.put(state, plotBlock);
                } catch (Throwable e) {}
            }
            PS.debug("Done!");
        } catch (Throwable e) {
            e.printStackTrace();
        }
    }
```
Not the prettiest code. It is also unreliable due to the injection Sponge performs on the minecraft code. They perform modifications at bytecode level. Granted it may work for one version of Sponge but it may change. Sponge won't have everything implemented right away and by having access to net.minecraft we can implement things like per player time.

MattBDev:
I will be pulling this tonight if there are no more objections and if I don't hear from @Byteflux

RoboMWM:
So even as per discussion in #intellectualcrafters the above wouldn't be better just because of gradle (and granted, that is all a reason centered around Sponge). Therefore - not seeing the reason other than that Sponge plugins use gradle (even though it's not required) and you guys (via comments in this thread) prefer to align this with Sponge as much as possible - which is the sole reason I see so far in favor of gradle.

Sauilitired:
@MattBDev I am all in favour for this. The only reason I've seen against it, is that.... it's... not needed? Well, surely, that might be true, in some cases. However, I still feel like it's a proper change.

manuelgu:
> The only reason I've seen against it, is that.... it's... not needed? 

Agree.
Matt, you are making sense with what you are saying though I don't see too much of need of actually switching..

Byteflux:
PlotSquared is first and foremost a Bukkit plugin. 99% of PlotSquared users are Bukkit users. The Bukkit community prefers to use Maven. Let's just make sure we all understand that. Yes, I understand some of you have strong distaste for Bukkit.

So far the argument for Gradle is that it increases compatibility with Sponge by eliminating the need to use reflection to access internals. This is not a flaw inherent to Maven. Switching to Gradle isn't the only solution to solving this problem. Maven is fully capable of setting up a multi-module project (albeit with more lines than you would with Gradle). If it ever came down to reflection being a pain to use, then I vote switching to a multi-module Maven project with any necessary plugins that would help make the process easier.

The Sponge userbase for PlotSquared is in the extreme minority. By that fact alone, in situations like this, I think it's important to appeal to the wishes of the larger community, especially when the change being proposed doesn't actually introduce any significant benefits that can't be provided by Maven.

I've seen PlotSquared's POM. It's not pretty. I think a lot of the confusion about Maven and using reflection is due to nobody here (who's advocating Gradle) actually having ever set up a proper multi-module Maven project.

I'm still against switching to Gradle. I've been a long time Maven user, I see no reason to switch. The argument about reflection doesn't hold up for reasons I've already stated. Since there is no real argument here, other than people being a fan of Sponge (again, I've stated these people are an extreme minority of the PlotSquared community) and Gradle, I propose that we continue to use Maven, and if reflection becomes a problem, switch to a multi-module project structure like every other cross-platform plugin that uses Maven.

MattBDev:
@Byteflux I think the arguement that not a lot of people use Sponge isn't relevant. As Sponge approaches an official stable release this year, more plugins will be developing for it. A platform that's in beta (like Sponge) isn't likely to have a lot of users until it is stable and there are plugins available. I have also setup a multi-module maven project before (nothing public so don't bother looking for it) and I have found it is a bit more complex than a multi-module gradle project.

minecrafter:
@MattBDev Having set up multi-module projects with Maven before, I do agree that it requires quite a bit of boilerplate. That doesn't mean you should switch to Gradle.

Again, I would still expect to see Sponge users in the minority even if Sponge does a full stable release this year. Sponge will need time to build user adoption.

MattBDev:
I don't think the userbase of Sponge is relevant in this discussion.
I think @Sauilitired should make a decision based on this discussion. I prefer to not decide myself due to bias.

Byteflux:
@MattBDev I disagree. I think it's very relevant. We're talking about replacing an entire build system that the larger community is already established and comfortable with, with one that we're less comfortable using.

I have set up multi-module projects, both private and public and as an avid Maven user, I find them to be easy if you do it once or twice.

RoboMWM:
> I don't think the userbase of Sponge is releveant in this dicussion
Yet you keep bringing it up, and seems to be the only "solid" reason for switching.

Byteflux:
> I think @Sauilitired should make a decision based on this discussion. I prefer to not decide myself due to bias of creating the PR.

Perhaps @boy0001 should be the one making the decision, as @Sauilitired has also demonstrated a clear bias towards Sponge.

MattBDev:
@Byteflux or they can both chat with each other. I just know that it isn't my call.

MattBDev:
@Byteflux I am also curious if you took a look at the build.gradle files though?

Byteflux:
Yep, I've seen the Gradle files. I get that they have a nice fancy DSL for doing things. I understand that the build file is smaller. To me though, that's not really important. Could be that I've spent too much time with my head in POM files, I might be desensitized to all the XML. :)

Sauilitired:
@Byteflux Just a correction, I am not a huge fan of bukkit. I am even less of a fan of Sponge. The whole reason why I wanted to make the core abstract, has nothing to do with either of those projects. My distaste towards bukkit is mainly because of the disgusting state the community is moving towards, that is the community in general, so no one here should take offense. I just don't feel like it's worth sacrificing either cleanliness nor usability to please a community of which I am not too found. End of that.

@MattBDev I am not very active here these days. With that said, PlotSquared is still like a baby to me. 

@boy0001 In reality, this is your call. Do as you wish. However, if we don't move to gradle, we have to at least use a multi-module setup (I'm capable to set it up myself, if so is needed).

SilverCory:
TL;DR: Migrated to gradle because SpongeAPI isn't complete.

Issue #124, opened 2015-01-31
New flags


boy0001:
Some other basic worldguard flags would be nice:
 - ice-melt

PR #909, opened 2016-02-29
Implements the ice-melt flag

zombachu:
Implements #124

Hopefully you guys are fine with where I put this... I have no experience with sponge so I left that alone.


boy0001:
The flag also needs to be registered with the FlagManager, I can do that though.

zombachu:
I believe that it's [already being registered](https://github.com/IntellectualSites/PlotSquared/blob/01710e3ddb3f24ddbf4afe838139bf405cea2c0a/Core/src/main/java/com/intellectualcrafters/plot/PS.java#L1939), but thanks for accepting my pull request!

boy0001:
Neat.

PR #986, opened 2016-03-20
A small change to /p weanywhere's behavior

zombachu:
Allows players with worldedit toggled to use the wand anywhere to set positions.


boy0001:
Since WorldEdit cancels the block break event, this could also be achieved by changing the priority to highest and adding ignorecancelled to the eventhandler annotation.

zombachu:
@boy0001 I wasn't sure if changing the event properties would break something so I decided to play it safe. Either way, I think it would be a nice change to have.

PR #1012, opened 2016-03-28
Update Plugin command to show correct versions

manuelgu:
- Update the `/plot plugin` command to show correct versions for current and newest

Not using `PS.get().update` but rather reading from the GitHub releases. Can be changed if needed/wanted.


MattBDev:
I approve. Thanks!

PR #1095, opened 2016-05-04
Ratings fix

diecode:


MattBDev:
I am gonna look at this myself actually. I won't accept this PR but I will leave it open until I fix it myself. If it is just an issue with the ratings map being null, it can be fixed deeper in the code.

PR #1107, opened 2016-05-11
Update to 1.9.4 and List<> -> ArrayList<>

Heddo:


MattBDev:
There is already a 3.3.4 branch open. 3.4.0 isn't even released or ready to be so it does not make sense to jump to 3.4.1

PR #1112, opened 2016-05-13
Method renaming, util class and logic change

manuelgu:

- Rename `doDeathTask` to `doRespawnTask` as it is being called on player respawn
- Add `HttpUtil` class (more stuff to be added) to read contents of url
- Fetch the latest version when executing `/plot update` and not use the version at the time of the last restart


boy0001:
Have you tested these changes?

PR #1137, opened 2016-05-27
Fix forcefield flag

manuelgu:

- Fixes `forcefield` flag not working properly

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/ForceFieldListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/ForceFieldListener.java
@@ -5,15 +5,14 @@ import com.intellectualcrafters.plot.flag.Flags;
 import com.intellectualcrafters.plot.object.Location;
 import com.intellectualcrafters.plot.object.Plot;
 import com.intellectualcrafters.plot.object.PlotPlayer;
-import com.plotsquared.bukkit.events.PlayerEnterPlotEvent;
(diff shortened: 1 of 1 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listeners/ForceFieldListener.java
```diff
@@ -84,29 +83,27 @@ public Vector calculateVelocity(PlotPlayer pp, PlotPlayer e) {
     }
 
     @EventHandler
-    public void onPlotEntry(PlayerEnterPlotEvent event) {
+    public void onPlotEntry(PlayerMoveEvent event) {
```
MattBDev: why was the event changed to PlayerMoveEvent?

manuelgu: Forgot to rename it to `onPlayerMove` but since @boy0001 wanted to move it to a different class anyway, it doesn't matter for now.

It was moved to the PlayerMoveEvent because what the code does is, it checks there is a player nearby who is permitted to be on the plot (owner, trusted or owner) and the player that moves is not permitted. If that is the case, it pushes the player who is not permitted back as soon as their distance is not further than 5 blocks from each other.

As of right now, when the code was only called in the PlayerEnterPlotEvent, it would have only affected a player entering the plot when there is a permitted player not further than 5 blocks from the border away (where the event location is). So it was kinda useless because once you entered the plot the event would never call again and the pushing back wouldn't work.

PR #1141, opened 2016-05-29
Fix some grow flags

### LINKED ISSUE #1140: All block spread flags are broken
**Debug paste link**: 
http://hastebin.com/dipuqizaha

The three block spreading flags, grass-grow, mycel-grow, and vine-grow don't work. I believe this is caused by [line 784](https://github.com/IntellectualSites/PlotSquared/blob/85055320c22b93b08747a21a4d4cf39b2ded5b37/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java#L784), which is actually checking the block material that is being altered, not the one it's going to turn to. A fix should be changing the expression to event.getSource().getType().

**How to replicate**:
Try one of the flags listed above.

Make sure you've completed the following steps (put an X between of brackets):
- [X] Include `/plot debugpaste`
- [X] Made sure there aren't duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [X] Made sure you're using an updated version of PlotSquared
- [X] Made sure the bug/error isn't caused by any other plugin

manuelgu:

- Fix #1140

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -780,6 +780,9 @@ public class PlayerEvents extends PlotListener implements Listener {
             return;
         }
         Plot plot = area.getOwnedPlot(location);
+        if (plot == null) {
+            return;
+        }
         switch (event.getSource().getType()) {
             case GRASS:
                 if (Flags.GRASS_GROW.isFalse(plot)) {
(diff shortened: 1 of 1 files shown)


manuelgu:
// TODO resolve NPE http://hastebin.com/unesozabeb.avrasm
Check whether the flag is actually present before checking `#isFalse( )`

boy0001:
> // TODO resolve NPE http://hastebin.com/unesozabeb.avrasm
Check whether the flag is actually present before checking #isFalse( )

.isFalse() will return false if the flag isn't present. Your issue is the plot object is null. 

> flag.isFalse(null)

throws the NPE

Issue #1291, opened 2016-08-06
Fences Gate are not in good way when debugroadregen


jewome62:
**Debug paste link**: http://hastebin.com/rogokedizo (World Edit 6.1.3)
and world config for this world : http://hastebin.com/bidilemugu.sm

**Description of the problem:**
The fences gates aren't copy well for roadregen

**How to replicate**:
First i create schematic with /plot createroadschematic
![2016-08-06_14 51 39](https://cloud.githubusercontent.com/assets/472429/17456854/74ac66ae-5be5-11e6-8361-d578f3517f6e.png)

Secondly I go to another plot (with ugly stone) and make /plot debugroadregen
![2016-08-06_14 51 58](https://cloud.githubusercontent.com/assets/472429/17456855/781d058c-5be5-11e6-9c84-2d179c1569a6.png)

Make sure you've completed the following steps (put an X between of brackets):
- [x] Include `/plot debugpaste` **in head of topic**
- [x] Made sure there aren't duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue) **search road or fence keyword**
- [x] Made sure you're using an updated version of PlotSquared **Yes, 3.4.4-Snapshot**
- [x] Made sure the bug/error isn't caused by any other plugin **For a copy problem, maybe World edit ? but it is updated to last version**

jewome62:
I retry with version 3.4.5-SNAPSHOT-c99dd1e and I have same problem

jewome62:
I verify the schematic, and it is good.
I think the block aren't rotate when he create road.

I dont understand, the id block of fence gate are in rotate function (block id : 107) : https://github.com/IntellectualSites/PlotSquared/blob/30d18c917d61c88ab924cbe8234309d56c98cc6c/Core/src/main/java/com/intellectualcrafters/plot/generator/HybridPlotWorld.java#L51

And the stairs are well rotated

jewome62:
I make some test with data id. My reference is the spruce stair because it's work with it !
And they are the same algorythm

![2016-08-06_16 37 13](https://cloud.githubusercontent.com/assets/472429/17457281/20377a6e-5bf4-11e6-85de-a763d64235fd.png)
![2016-08-06_16 37 16](https://cloud.githubusercontent.com/assets/472429/17457282/2156f21c-5bf4-11e6-8ce3-14a730ce6b02.png)
[...]

PR #1293, opened 2016-08-06
Bugfix fences gate rotation

jewome62:
The fences gates don't rotate as stair but as pumpkin or bed.

See #1291


jewome62:
Hum ... I try this but that don't work.
Do you have any idea ?

jewome62:
This work on my server

Issue #1300, opened 2016-08-08
An event when a player is removed from a plot


MisterErwin:
I was wondering why there is a `PlayerPlotTrustedEvent`, but no `PlayerPlotRemovedEvent`. (or a similar name)

And as a small help: the event [calling part in the Add command](https://github.com/IntellectualSites/PlotSquared/blob/30d18c917d61c88ab924cbe8234309d56c98cc6c/Core/src/main/java/com/intellectualcrafters/plot/commands/Add.java#L74).

And the (not event calling part) in the [Remove command](https://github.com/IntellectualSites/PlotSquared/blob/30d18c917d61c88ab924cbe8234309d56c98cc6c/Core/src/main/java/com/intellectualcrafters/plot/commands/Remove.java#L79).

manuelgu:
Such an event should also store the previous role of that player (trusted, added, denied).

Feel free to PR :)

MisterErwin:
Working on it :)

The PlayerPlotTrustedEvent,... do have an added field. I am using this right now.

By doing so: Is there a special reason behind [this extra check](https://github.com/IntellectualSites/PlotSquared/blob/30d18c917d61c88ab924cbe8234309d56c98cc6c/Core/src/main/java/com/intellectualcrafters/plot/commands/Remove.java#L78)?

(It is not there for the "all-player" check)

manuelgu:
It's there so you can call the respective `removeTrusted`, `removeMember` and `removeDenied`

MisterErwin:
@manuelgu And Line [74](https://github.com/IntellectualSites/PlotSquared/blob/30d18c917d61c88ab924cbe8234309d56c98cc6c/Core/src/main/java/com/intellectualcrafters/plot/commands/Remove.java#L74)? :stuck_out_tongue:

manuelgu:
Because otherwise the players wouldn't be removed. `DBFunc.everyone` is * which means remove all trusted, all members and all denied.

MisterErwin:
Oh.. my bad. I did not look at the source correctly.

MisterErwin:
Just noticed that I forgot to close this, as the PR is merged

PR #1301, opened 2016-08-08
Calling events when players are removed from plots & correct calling of Member & Trusted events

MisterErwin:
Now calling the respective events when a user is removed from a plot.

And switched the event calling of adding Helpers and Trusted players. (Correct me if that was wrong)
#1300


Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitEventUtil.java
```diff
@@ -1,30 +1,9 @@
...
 
 import com.intellectualcrafters.plot.flag.Flag;
-import com.intellectualcrafters.plot.object.Location;
-import com.intellectualcrafters.plot.object.Plot;
-import com.intellectualcrafters.plot.object.PlotArea;
-import com.intellectualcrafters.plot.object.PlotCluster;
-import com.intellectualcrafters.plot.object.PlotId;
-import com.intellectualcrafters.plot.object.PlotPlayer;
-import com.intellectualcrafters.plot.object.Rating;
+import com.intellectualcrafters.plot.object.*;
```
manuelgu: Don't use * imports.

Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitEventUtil.java
```diff
@@ -1,30 +1,9 @@
...
-import com.plotsquared.bukkit.events.PlayerTeleportToPlotEvent;
-import com.plotsquared.bukkit.events.PlotClearEvent;
-import com.plotsquared.bukkit.events.PlotComponentSetEvent;
-import com.plotsquared.bukkit.events.PlotDeleteEvent;
-import com.plotsquared.bukkit.events.PlotFlagAddEvent;
-import com.plotsquared.bukkit.events.PlotFlagRemoveEvent;
-import com.plotsquared.bukkit.events.PlotMergeEvent;
-import com.plotsquared.bukkit.events.PlotRateEvent;
-import com.plotsquared.bukkit.events.PlotUnlinkEvent;
+import com.plotsquared.bukkit.events.*;
```
manuelgu: Same here.

boy0001:
@MisterErwin Legacy reasons, PlotSquared when the groups were different. You'll see other references to Trusted/Helpers instead of Members/Trusted. 

I see no problem merging this for the 3.5 update, unless @MattBDev has something to add to the conversation?

MattBDev:
As long as it is a non-breaking change, I'm fine with it.

MisterErwin:
Till now nothing broke on my test server. It appears to be fine on the "production server",

Issue #1351, opened 2016-09-12
Reloading plugins that add flags causes data corruption


MisterErwin:
(I know, you should never reload plugins, but well...)
When reloading only a plugin which adds external Flags to PS, some flags corrupt their data.

When we have the following data:
> Flags: test-flag: [a,b]   

The size is 2

And after reloading the plugin which adds __test-flag__
>Flags: test-flag [[a-b]]

The size is 1

(test-flag is a StringListFlag in this case)

[Line 172 of Flags](https://github.com/IntellectualSites/PlotSquared/blob/a95b68b73c87c465a6294d4e30cd47ee68f6fc76/Core/src/main/java/com/intellectualcrafters/plot/flag/Flags.java#L172) appears to be the cause for that
`flag.parseValue("" + remove)`

I am unsure if a fix is reasonable, but I will make a PR nonetheless.

P.S.: I am too lazy to add the debugpaste/version, as it's visible in the (at this time) last commit `a95b68b73c87c465a6294d4e30cd47ee68f6fc76`

MisterErwin:
See PR #1352

boy0001:
I'll merge with some modifications.

PR #1352, opened 2016-09-12
A (possible) fix to #1351

MisterErwin:
Tested in a local environment and it works.
Nothing broke due do that. (Nor do I know a plugin that tries to do what causes #1351 )
(Except if two plugins try to register the same flag)


boy0001:
If the collection implements a custom toString() method, it would work fine. I'll make some changes so that it only does this if toString isn't overridden.

PR #1424, opened 2016-11-12
translation

c7w:
translation of chinese


boy0001:
Note on translations: Chinese characters need to be escaped in order to be read properly.

manuelgu:
Language files are supposed to be in `Core/src/main/resources/``, not `/`.

boy0001:
that too.

PR #1570, opened 2017-04-02
Do not block the creation of items. Fixes #1506.

### LINKED ISSUE #1506: How do I allow players to drop Items?
This isn't a bug report, but rather a question. How do I allow players to drop items on the road without them instantly despawning?

ghacproductions:


boy0001:
Set `misc_spawn_unowned` to true in the `worlds.yml`.

PR #1583, opened 2017-04-16
Added useSSL=false to JDBC connection URL

manuelgu:
This is required according to MySQL 5.5.45+, 5.6.26+ and 5.7.6+.


Legoman99573:
I think best is to make it configure as some might have ssl set up. Setting ssl to false by default will work too ;)

PR #1750, opened 2017-10-30
Allow SSL to be enabled or disabled in storage.yml

Legoman99573:
Some people may have ssl set up on their MySQL Databases, so I don't see the problem allowing people to enable it IMO.

By default, it will be set to false and has a message for people not to enable it if they didn't set it up on their servers.


manuelgu:
👍 
Might want to consider implementing `verifyServerCertificate` too.

Legoman99573:
@manuelgu i forgot to add that to the check. Can set as a seperate config option. Thanks for reminding me ;)

PR #1757, opened 2017-11-06
Implement VehicleMoveEvent for Bukkit.

### LINKED ISSUE #1754: Vehicles cancel out PlotLeave and PlotJoin events
**Debug paste link:** 
https://hastebin.com/yelegamelo

When you ride a minecart off of a plot and onto another plot, it doesnt trigger PlotLeave or PlotJoin events until you hop out of a vehicle and move around on another plot.

**How to replicate:** 
Video shows how to reproduce, but its really easy to replicate.
https://youtu.be/LSPWk3AFM8k

*In before i get the allow killing road vehicles, i do have a plugin that requires it to be disabled.*

**Checklist**:

- [x] I included a `/plot debugpaste` link
- [x] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] I made sure I am using an up-to-date version of PlotSquared
- [x] I Made sure the bug/error is not caused by any other plugin

Sauilitired:
This fixes #1754. It also makes sure that you cannot use vehicles to enter plots that you're denied from.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -6,22 +6,8 @@ import com.intellectualcrafters.plot.config.C;
 import com.intellectualcrafters.plot.config.Settings;
 import com.intellectualcrafters.plot.flag.Flags;
 import com.intellectualcrafters.plot.flag.IntegerFlag;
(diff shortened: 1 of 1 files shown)


Legoman99573:
Thanks for adding this. It has been a long problem as custom plugins vehicles could bypass it. Should work on all vehicles and those who deny vehicles will despawn instead of staying spawned until they dismount.

Issue #1723, opened 2017-09-14
no wether update after "/p flag remove weather"


RedstoneFuture:
**Description of the problem:** 
After the command "/p flag remove weather" PlotSquared will not update the weather.

**How to replicate:** 
1. The world-weather is sun.
2. I set the plot weather with "/p flag set weather rain" -> the weather in the plot will update
3. I remove the weather on the plot with "/p flag remove weather"
--> If I fly over the plot or outside the plot the weather will not update. I have still rain.

**Checklist**:
- [X] I made sure there are no duplicates of this report
- [X] I made sure I am using an up-to-date version of PlotSquared
- [X] I Made sure the bug/error is not caused by any other plugin

Sauilitired:
Fixed in #1767

RedstoneFuture:
re-open: Empire has removed the fix trom Opachl after problems at many users, I think.

PR #1767, opened 2017-11-15
fix: no wether update after "/p flag remove weather"

Opachl:
no wether update after "/p flag remove weather" #1723


Opachl:
Are there any suggestions for this fix ? or why did you not apply it?

Opachl:
@Sauilitired Sorry it was ask as question not to force you. I know how much work and time intensive a project like this is.

Issue #1762, opened 2017-11-10
shulker-teleport in a other plot


RedstoneFuture:
Problem: **A user can teleport shulker from your plot to a next plot.**

**How to replicate:** 

1. spawn shulker on the border in your plot
2. remove the block under the shulker
--->  the shulker will teleport to a random position and thus in other plots
         (Currently the plugin can't remove the shulker, because ge down't go over the road.)

**Checklist**:

Serversoftware: Paper version git-Paper-1209 (MC: 1.12.2) (Implementing API version 1.12.2-R0.1-SNAPSHOT)
- [X] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [X] I made sure I am using an up-to-date version of PlotSquared
- [X] I Made sure the bug/error is not caused by any other plugin

Legoman99573:
What version of plotsquared are you using, plus forgot to paste the debug paste. That helps developers alot.

Use `/p debugpaste`

PR #1774, opened 2017-11-19
fix: shulker-teleport in a other plot

Opachl:
shulker-teleport in a other plot #1762 

As we dont have an shulker teleport event. This is a workaround.

We assign to each shulker a new metadata (ownerplot) with the plotid of the shulker stays on

in the runEntityTask we check the shulker if it sill is on the original plot as it was spawnd. if not we delete it

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -1263,7 +1263,8 @@ public class PlayerEvents extends PlotListener implements Listener {
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
     public void onBlockDispense(BlockDispenseEvent event) {
         Material type = event.getItem().getType();
-        if (type != Material.WATER_BUCKET && type != Material.LAVA_BUCKET) {
+        Material dispenserType = event.getBlock().getType();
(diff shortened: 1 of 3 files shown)


Opachl:

Sauilitired:
Honestly, I have not yet had time to test this on my own. Are you positive it works? (Both for the current version, but also previous versions)? The same applies to your other pull requests.

Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
```diff
@@ -1263,7 +1263,8 @@ public void onBlockPistonRetract(BlockPistonRetractEvent event) {
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
     public void onBlockDispense(BlockDispenseEvent event) {
         Material type = event.getItem().getType();
-        if (type != Material.WATER_BUCKET && type != Material.LAVA_BUCKET) {
+        Material dispenserType = event.getBlock().getType();
```
Sauilitired: So, great that this is fixed (I assume it's a bug fix as well?). 

However, for future reference I'd suggest that things like this would be a separate pull request. I'll let this one pass, because it's simple enough. :)

Opachl: yea sorry for that :)

Issue #1691, opened 2017-08-14
argument priorisation problem for /p tp


khlfkk:
**Bug report **

paste link:
https://pastebin.com/Wg27wNnq

Description of the problem:
when visit the player claim  /p tp 7005151678
it caused exeception:(  java.lang.NumberFormatException: For input string: "7005151678" )

boy0001:
Is that meant to be a player name?

samueleycw:
yup, it is a player name called "7005151678". May i know is it don't support full number IGN?

RedstoneFuture:
Yes. I get a error too. ^^
"An error occured: For input string: 005151678"

https://namemc.com/profile/7005151678.1

Opachl:
I will have a look to this. :)

Bug is reproducible. I try to find a Solution.

I will try change the code in that way that it always asumes this fallbackorder:
-> Playername -> Alias -> world -> id

Additional as the Player is unkown (means never played on the Server) will result assuming it as id. See Fallbackorder. For this special case we will catch the NumberFormatException and convert it to a normal console warning message.

PR #1778, opened 2017-11-22
fix: numeric Playernames thows NumberFormatException on Teleport

Opachl:
fix for #1691 

Refactoring of the Teleport Handling

Supports now:
     * p h xPlotPlayerGuy [PAGE] = Player  [numeric] [alphanumeric]
     * p h MyPlantage      [PAGE] = Alias    [numeric] [alphanumeric]
     * p h PlotworldLarge [PAGE] = World  [numeric] [alphanumeric]
     * p h 1:1		       [PAGE] = plotId  [numeric (1,1|1;2) ]
     * p h 4		                   = 4 th Plot of the Player if available  [numeric (1,1|1;2) ]
Check order: Player, Alias, World, PlotId, Page [PAGE]
[PAGE] is optional

### DIFF
--- a/Core/src/main/java/com/intellectualcrafters/plot/PS.java
+++ b/Core/src/main/java/com/intellectualcrafters/plot/PS.java
@@ -1968,17 +1968,14 @@ public class PS{
      * @param alias to search plots
      * @param worldname to filter alias to a specific world [optional] null means all worlds
      *
-     * @return Set<{@link Plot}> null if nothing found
+     * @return Set<{@link Plot}> empty if nothing found
      */
     public Set<Plot> getPlotsByAlias(String alias, String worldname) {
-    	Set<Plot> result = null;
+    	Set<Plot> result = new HashSet<>();
     	
     	if (alias != null) {
     		for (Plot plot : getPlots()) {
-    			if (alias.equals(plot.getAlias()) && (worldname == null || worldname.equals(plot.getWorldName()))) {
-    				if (result == null) {
-    					result = new HashSet<>();
-    				}
+    			if (alias.equals(plot.getAlias()) && (worldname == null || worldname.equals(plot.getWorldName()))) {    				
     				result.add(plot);    				
     			}
     		}
--- a/Core/src/main/java/com/intellectualcrafters/plot/PS.java
+++ b/Core/src/main/java/com/intellectualcrafters/plot/PS.java
@@ -1961,6 +1961,31 @@ public class PS{
             return null;
         }
     }
+    
+    /**
+     * Get Plots based on alias
+     *
+     * @param alias to search plots
+     * @param worldname to filter alias to a specific world [optional] null means all worlds
+     *
+     * @return Set<{@link Plot}> null if nothing found
+     */
+    public Set<Plot> getPlotsByAlias(String alias, String worldname) {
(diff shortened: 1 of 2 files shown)


Opachl:
Feel free to contact me for questions for this fix. :)

Core/src/main/java/com/intellectualcrafters/plot/PS.java
```diff
@@ -1961,6 +1961,31 @@ public PlotArea getPlotAreaByString(String search) {
...
         }
     }
+    
+    /**
+     * Get Plots based on alias
+     *
+     * @param alias to search plots
+     * @param worldname to filter alias to a specific world [optional] null means all worlds
+     *
+     * @return Set<{@link Plot}> null if nothing found
```
Sauilitired: This should return an empty set rather than null. Always avoid using null when it is possible (that's why we :heart: Optional<>)

Core/src/main/java/com/intellectualcrafters/plot/commands/Visit.java
```diff
@@ -31,6 +32,9 @@
         category = CommandCategory.TELEPORT)
 public class Visit extends Command {
 
+	private final int PageOutOfRange = -998899; // this is to flag Page argument is to long. Can occur if someone enters a large number (some player uses numeric names)
```
Sauilitired: Make this  `private static final int PAGE_OUT_OF_RANGE ...` :)

Core/src/main/java/com/intellectualcrafters/plot/commands/Visit.java
```diff
@@ -160,4 +158,62 @@ public void run() {
...
+     * @param argument to search plots
+     * <pre>
+     * Samples:
+     * p h xPlotPlayerGuy	= Player [numeric] [alphanumeric] [UUID]
+     * p h MyPlantage		= Alias  [numeric] [alphanumeric] [Set<Plot>]
+     * p h PlotworldLarge	= World  [numeric] [alphanumeric] [PlotArea]
+     * p h 1:1		= plotId [numeric] (1,1|1;2)      [PlotId]
+     * </pre>
+     * @param applicablePlotArea the area from the player invoked the command
+     * @return Collection<{@link Plot}> null if nothing found
```
Sauilitired: Same here! Then we can do `Collection#isEmpty` rather `than == null`

Core/src/main/java/com/intellectualcrafters/plot/PS.java
Opachl: okay normally i would do this too but i found a lot of code returning colletions null instead of empty so i thought this is your code style ;)

Sauilitired: @Opachl I know that the code base is a bit... messy. However, it is quite hard to change things at this point without losing backwards compatibility. I think it'd be smart to release a breaking version soon(~ish) and make the code pretty again. As you may have noticed, there code base also struggles with packages and whatnot.

So yeah, I see your point :laughing: But let's try to make things pretty moving forward.

Opachl:
@Sauilitired I changed the Code with your suggestions

Opachl:

Issue #1718, opened 2017-09-12
remove plot-alias with "/p alias"


RedstoneFuture:
At now I can't remove a plot-alias with the command "/p alias".
With "/p desc" I remove the descrition of a plot too.

Opachl:
suggestion:
change command to 
/p alias set xxxxx
Add a new command
/p alias remove

PR #1779, opened 2017-11-22
fix: remove plot-alias with "/p alias"

Opachl:
remove plot-alias with "/p alias" #1718


Bukkit/src/main/resources/plugin.yml
```diff
@@ -131,7 +131,8 @@ permissions:
       plots.toggle.chat: true
       plots.set.biome: true
       plots.set.home: true
-      plots.set.alias: true
```
Sauilitired: Is there any way to keep this backwards compatible? I think it would be better to avoid changing (~removing~) the permission. Thoughts?

Core/src/main/java/com/intellectualcrafters/plot/commands/Alias.java
```diff
@@ -2,27 +2,83 @@
...
 import com.intellectualcrafters.plot.object.Plot;
 import com.intellectualcrafters.plot.object.PlotPlayer;
 import com.intellectualcrafters.plot.object.StringWrapper;
 import com.intellectualcrafters.plot.util.MainUtil;
+import com.intellectualcrafters.plot.util.Permissions;
+import com.intellectualcrafters.plot.util.StringMan;
 import com.intellectualcrafters.plot.util.UUIDHandler;
 import com.plotsquared.general.commands.CommandDeclaration;
 
+import de.schlichtherle.util.Arrays;
```
Sauilitired: What is this? 😕

Bukkit/src/main/resources/plugin.yml
Opachl: @Sauilitired
what would be better for you addind the permisson again and use the 
plots.set.alias as obsolete permission and plots.set.alias as new permisson.
**or** 
leave the original permisson plots.set.alias and remove the new one plots.alias.set
Note: should we change the plots.alias.remove to plots.remove.alias or should we leave it as it was

Core/src/main/java/com/intellectualcrafters/plot/commands/Alias.java
Opachl: @Sauilitired
what would be better for you addind the permisson again and use the
plots.set.alias as obsolete permission and plots.set.alias as new permisson.
or
leave the original permisson plots.set.alias and remove the new one plots.alias.set
Note: should we change the plots.alias.remove to plots.remove.alias or should we leave it as it was

Bukkit/src/main/resources/plugin.yml
Opachl: @Sauilitired
what would be better for you addind the permisson again and use the
plots.set.alias as obsolete permission and plots.set.alias as new permisson.
or
leave the original permisson plots.set.alias and remove the new one plots.alias.set
Note: should we change the plots.alias.remove to plots.remove.alias or should we leave it as it was

Core/src/main/java/com/intellectualcrafters/plot/commands/Alias.java
Opachl: I will remove this that was a misstake

Opachl:
@Sauilitired I changed the Code with your suggestions

Opachl:

RedstoneFuture:
Yes, we have tested it and it is currently working with the latest and the old versions.

Sauilitired:
So, just a question. Does anyone else think it'd be worth having `plots.set.alias` as a fallback for the new permissions? I can just imagine people will get confused and file bug requests otherwise :thinking:

RedstoneFuture:
I and @Opachl think it's a bad fallback feature. Any good admin will check the chancelog and fix any config and permission inconsistency.

Sauilitired:
@RedstoneFuture I tend to agree.

Issue #1798, opened 2017-12-09
Improvement: Code refactoring (FLAG:misc-interact, frame-interact) (EVENT:interact-misc)


Opachl:
# flag misc-interact allows modify armor stands

**Description of the problem:** 
Final goal make the code more readable and clear.

The improvement is located here PlayerEvents_1_8.onInteract(PlayerInteractAtEntityEvent e)
Added in commit: https://github.com/IntellectualSites/PlotSquared/commit/93c48544541a7783c33b8972aebd73e12ccab3f0

Step 1:
Suggestion remove the special handling for AmorStand from the PlayerEvents_1_8.class and move it to the PlayerEvent.class

Step 2:
Refactoring the misc-interact flag and the interact-misc event.
After understanding the current code i came to the point. That the moving the handling for AmorStand to PlayerEvents makes most of the code obsolete since PlayerEvents.class handles all that stuff.

Step 3:
Remove unused code
The case INTERACT_MISC in EventUtil is unused -> no one set at any time the EventType for that
Remove the EventType INTERACT_MISC itself -> not used anymore

Step 4:
Add a new flag frame-interact for item frame this is for player that want seperate the interact between 
ItemFrame and AmorStand

Step 5:
Add / update the informations to the Wiki

This will be made by RedstoneFuture

PR #1799, opened 2017-12-09
Improvement: Code refactoring (FLAG:misc-interact, frame-interact) (EVENT:interact-misc)

Opachl:
Final goal make the code more readable and clear. #1798

The improvement is located here PlayerEvents_1_8.onInteract(PlayerInteractAtEntityEvent e)
Added in commit: 93c4854

Step 1:
Suggestion remove the special handling for AmorStand from the PlayerEvents_1_8.class and move it to the PlayerEvent.class

Step 2:
Refactoring the misc-interact flag and the interact-misc event.
After understanding the current code i came to the point. That the moving the handling for AmorStand to PlayerEvents makes most of the code obsolete since PlayerEvents.class handles all that stuff.

Step 3:
Remove unused code
The case INTERACT_MISC in EventUtil is unused -> no one set at any time the EventType for that
Remove the EventType INTERACT_MISC itself -> not used anymore

Step 4:
Add a new flag frame-interact for item frame this is for player that want seperate the interact between
ItemFrame and AmorStand

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -51,6 +51,7 @@ import org.bukkit.block.BlockState;
 import org.bukkit.command.PluginCommand;
 import org.bukkit.entity.Ageable;
 import org.bukkit.entity.Animals;
+import org.bukkit.entity.ArmorStand;
 import org.bukkit.entity.Arrow;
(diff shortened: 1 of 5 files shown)


boy0001:
PlotSquared must support previous versions of the game. ArmorStands were introduced in 1.8, so it must be handled in a separate class.

Sauilitired:
API backwards compatibility isn't a necessity in this case, but game version backwards compatibility is, which this PR would break.

PR #1957, opened 2018-05-21
Fix Command classes to be hierarchical

Caelin:
Reorganized and created new classes for the command handler system so all Command classes are hierarchical now.  This allows command suggestions to work properly now past the first tier and makes the execute and onCommand methods simpler for each Command class.  Also fixed issues with permissions in relation with Sponge since Sponge does not use the .* wildcard.


Sauilitired:
Was this ever tested? If so it might be worth merging before further changes are made to commands, as there are already quite a few conflicts.

N0tMyFaultOG:
I will carefully close this PR now, it is outdated and targets the legacy branch. If someone wants to implement this stuff still, they can target the breaking branch.

Issue #1768, opened 2017-11-15
Failed to tab complete


guuslieben:
**Debug paste link:**
https://hastebin.com/arogiteqas

**Description of the problem:** 
Whenever both in-game or console users try to tab complete anything within the PlotSquared provided commands, nothing happens except for this output in the console : https://gist.github.com/guuslieben/c847a2e9caead33d18fc3501351bb32a

Using PlotSquared-Sponge-17.11.14-734c4bb-1937.jar

**How to replicate:** 
Execute /p2, /plots, /plot and use [Tab] to autocomplete, error messages occur in console.

**Checklist**:

- [X] I included a `/plot debugpaste` link
- [X] I made sure there are no duplicates of this report [(Use Search)]
- [X] I made sure I am using an up-to-date version of PlotSquared
- [X] I Made sure the bug/error is not caused by any other plugin

Tollainmear:
I have the same problem with p2.
[gist](https://gist.github.com/Tollainmear/97e07b1fbf08eee2fca3474cc4952ebc)

Using version=17.12.09-59557cf-1945

lurr:
Please note: Your hastebin link is invalid.

guuslieben:
@lurr To be expected as I recieved no reponse for over 2 months. Here is an updated link https://hastebin.com/quwoqusocu

RedstoneFuture:
Hello. Have you still this problem with a currently PlotSquared version?

guuslieben:
@RedstoneFuture I am still experiencing this issue. We recently updated to 1.12.2 (coming from 1.10.2) yet the issue persists.

BrainStone:
I am also experiencing this issue.

StackTrace:

```
> [21:26:39] [Server thread/FATAL] [net.minecraft.server.MinecraftServer]: Error executing task
java.util.concurrent.ExecutionException: java.lang.RuntimeException: Error occured while tab completing 'plot visit worl'
at java.util.concurrent.FutureTask.report(FutureTask.java:122) ~[?:1.8.0_161]
at java.util.concurrent.FutureTask.get(FutureTask.java:192) ~[?:1.8.0_161]
at net.minecraft.util.Util.func_181617_a(SourceFile:47) [h.class:?]
at org.spongepowered.common.SpongeImplHooks.onUtilRunTask(SpongeImplHooks.java:277) [SpongeImplHooks.class:1.12.2-2655-7.1.0-BETA-3065]
[...]

PR #1986, opened 2018-06-08
Fix errors with tab completion on Sponge.

dualspiral:
In Sponge, `CommandCallable#getSuggestions` is [marked as `Nonnull` at a package level](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/command/package-info.java). As a result, our implementation of Sponge assumes that `getSuggestions` returns a list, even if it is empty, and processes tab completions based on that.

In PlotSquared, you're currently using `null` to indicate no choices, which is causing tab completion errors as shown in #1768 because we use the return value to create our own list. PlotSquared's Sponge commands should be returning an empty list instead - which is what this PR does.


boy0001:
thanks

PR #2003, opened 2018-06-25
Update PlayerEvents.java

Feerko:
After set flag "pve" to true and "animal-attack" to false not added players can attack animals. Can you fix this?

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -2491,10 +2491,18 @@ public class PlayerEvents extends PlotListener implements Listener {
                     return false;
                 }
             } else if (victim instanceof Creature) { // victim is animal
-                if (plot != null && (plot.getFlag(Flags.ANIMAL_ATTACK, false) || plot.getFlag(Flags.PVE, false) || plot
-                        .isAdded(plotPlayer.getUUID()))) {
-                    return true;
+                if (plot != null || plot.isAdded(plotPlayer.getUUID())) {
+                  return true; 
+                 }
+
+                if (plot.getFlag(Flags.ANIMAL_ATTACK, false)) {
+                  return false; 
+                 }
+
+                if (plot.getFlag(Flags.PVE, false)) { 
+                    return true; 
                 }
+                
                 if (!Permissions.hasPermission(plotPlayer, "plots.admin.pve." + stub)) {
                     MainUtil.sendMessage(plotPlayer, C.NO_PERMISSION_EVENT, "plots.admin.pve." + stub);
                     return false;


Platinteufel:
Same here: Because hostile mobs didn't attack players anymore I've set the pve flag as default for all plots. Now I got the problem that players can kill mobs even on foreign plots where they are not added.

Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
```diff
@@ -2491,10 +2491,18 @@ public boolean entityDamage(Entity damager, Entity victim) {
                     return false;
                 }
             } else if (victim instanceof Creature) { // victim is animal
-                if (plot != null && (plot.getFlag(Flags.ANIMAL_ATTACK, false) || plot.getFlag(Flags.PVE, false) || plot
-                        .isAdded(plotPlayer.getUUID()))) {
-                    return true;
+                if (plot != null || plot.isAdded(plotPlayer.getUUID())) {
```
Sauilitired: This changes the logic. The removed snippet requires that the plot isn't null AND that the player is added to the plot, whereas this line would terminate if the plot isn't null, if the plot is null it would check if the player is added and thus result in a NullPointerException.

Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
```diff
@@ -2491,10 +2491,18 @@ public boolean entityDamage(Entity damager, Entity victim) {
...
+                  return true; 
+                 }
+
+                if (plot.getFlag(Flags.ANIMAL_ATTACK, false)) {
+                  return false; 
+                 }
+
+                if (plot.getFlag(Flags.PVE, false)) {
```
Sauilitired: This could be merged with the if statement above (that returns true), as it also requires that the plot isn't null.

Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
```diff
@@ -2491,10 +2491,18 @@ public boolean entityDamage(Entity damager, Entity victim) {
...
                 }
             } else if (victim instanceof Creature) { // victim is animal
-                if (plot != null && (plot.getFlag(Flags.ANIMAL_ATTACK, false) || plot.getFlag(Flags.PVE, false) || plot
-                        .isAdded(plotPlayer.getUUID()))) {
-                    return true;
+                if (plot != null || plot.isAdded(plotPlayer.getUUID())) {
+                  return true; 
+                 }
+
+                if (plot.getFlag(Flags.ANIMAL_ATTACK, false)) {
```
Sauilitired: Results in a NullPointerException if the plot is null.

Platinteufel:
Looks like the author of this PR is inactive. Is it able to merge with the changes from @Sauilitired?

Sauilitired:
I'll implement this when I have a bit more time

Platinteufel:
Any news here? @Sauilitired 
Players are reporting weekly that someone killed their animals or villagers. It's a bit annoying. :/

Issue #1956, opened 2018-05-19
dispenser und dropper can't drop a item


RedstoneFuture:
Since the last updates dispenser and dropper can't drop someone item in the plot.

Maybe because: https://github.com/IntellectualSites/PlotSquared/issues/1573

**EDIT:** because 5cb94f3

**Checklist**:

- [X] I included a `/plot debugpaste` link
- [X] I made sure there are no duplicates of this report
- [X] I made sure I am using an up-to-date version of PlotSquared

N0tMyFaultOG:
If I place them on the plot they work, placing them on the edge like on the reference above doesn't work.
Running latest P2 build

RedstoneFuture:
@N0tMyFaultOG Yes. I use the last version. And no, it doesn't work. Only water or lava block will spawn or respawn, but the item-drop is stopped.

RedstoneFuture:
With the change #589 -> #590

RedstoneFuture:
With this fix:
https://github.com/IntellectualSites/PlotSquared/commit/5cb94f3e45f4555c424524e7a3a7e79b92b0379e

![2018-05-21_06 01 32](https://user-images.githubusercontent.com/4140635/40290142-72e9da5c-5cbc-11e8-876f-0dc6d0d7ced5.png)

IAlekseyI22:
Yes, when it will be fixed??

RedstoneFuture:
I think currently the devs have many to do. We must wait.

CraigStr:
I'm also having this issue, but considering that 1.13 is coming out soon, it'll probably be merged into that update. But hopefully it'll happen sooner than later.

RedstoneFuture:
@boy0001 
fast to fix: IF you undoing #5cb94f3 it will work I think.

dextonanderson:
Also having this issue

RedstoneFuture:
Fixed in the last version (tested in 1.13)

PR #2028, opened 2018-07-20
fix dispensers

PhanaticD:
previously dispensers were not working at all, this change blocks lava and water and allows dispensers to function normally again on servers with plotsquared. if you were trying to block dispensers launching items into the road however that still is not blocked by this

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -1322,12 +1322,13 @@ public class PlayerEvents extends PlotListener implements Listener {
             case WATER_BUCKET:
             case LAVA_BUCKET: {
                 if (event.getBlock().getType() == Material.DROPPER) return;
+                BlockFace targetFace = ((org.bukkit.material.Dispenser) event.getBlock().getState().getData()).getFacing();
+                Location location = BukkitUtil.getLocation(event.getBlock().getRelative(targetFace).getLocation());
+                if (location.isPlotRoad()) {
+                    event.setCancelled(true);
+                }
             }
-        }
-        Location location = BukkitUtil.getLocation(event.getVelocity().toLocation(event.getBlock().getWorld()));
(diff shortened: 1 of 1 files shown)


Speentie8081:
dispensers were working fine, at least on my server. it actually ended up being a problem where people would use dispensers to lag out the server by spawning tnt or mobs etc. so I ended up blocking the use of dispensers in the creative world in my plugin.

PhanaticD:
they were not working on mine due to this code, the changes fix it

Speentie8081:
from looking at your commit, it seems like all you did was make it so dispensers do not function if fired into the road. it doesn't seem like it does anything that would make the event not cancelled because even before your commit was had, it was only cancelling the event if fired directly into the road. are you sure you were running the absolute latest version of plotsquared before you made your commit? it could be that you were running an outdated plotsquared that had a bug like what your talking about and when you made this commit and compiled it for your server you now have that bug fix because your using the latest plotsquared. just my thoughts, because it just seems kindof odd that you just moved the if-statement which cancels the event inside the lava bucket switch-case statement and that supposedly fixes the issue. if this did truly solve the issue it might be something to do with the version of spigot you are running that is causing the getVelocity() method to return some inaccurate value that makes the if-statement think that it's a road when it's not because all the getLocation() method does is return a version of the location with rounded off coordinates. like I said, I never had issues with this so I might be missing context of the situation, but if this fixes your issue-- great!

PhanaticD:
I was not the only one having this issue: https://github.com/IntellectualSites/PlotSquared/issues/1956

the value being returned was not correct from the velocity check, and yes I did use the latest

Speentie8081:
ah yes, that's what I ended up assuming at the end of my last post, that the getVelocity() method was returning an inaccurate value. and after looking at that other issue you linked, it makes a lot more sense now, I looked at the other issue and I see that on may 16th there was an updated to plotsquared that caused this issue. I haven't used droppers or dispensers in plots since mid-2016 so that makes perfect sense now why I never had the issue and you did. good job fixing the issue good sir- have a nice day.

PR #2029, opened 2018-07-21
Add support for spigot 1.13 and spigot 1.13-pre7

Speentie8081:
These commits adds support for spigot 1.13-pre7 but chunk generation is a lot slower then normal because it's using the fallback block local queue as the 1.9 compatible one is not compatible with 1.13, this is mainly a quick fix so current server owners can run plotsquared on their 1.13-pre7 minecraft servers. In time, a proper solution should be put in place that offers full 1.13 compatibility, but for now I believe this is good enough.


Speentie8081:
@boy0001 I just fixed my code, it was messing up support for 1.9 through 1.12. You can see very clearly what this patch as a whole does to the project to support 1.13 clients from the files changed tab.

PR #2066, opened 2018-08-16
Try to fix triggering of wooden buttons by arrows from untrusted players

david-bla:
Hey PlotSquared Team, first I'd like to thank you for this awesome project - well done 👍 

Im quite new to java and spigot/bukkit plugin developing - please don't merge this pull request.
But i thought this is the best way to show what I've done.

As the subject claims, my changes try to fix the "unprivileged" use of wooden buttons (the same will be with pressure plates). If a untrusted player fires a arrow and hit wooden button or pressure plate it will 
set the button powered!
I think the expected behavior will be: ignoring arrows from untrusted - allow arrows from trusted.
Also your sources also say: 
https://github.com/IntellectualSites/PlotSquared/blob/4b8c434669da5b700f700416ddaaad7526ce4b3c/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java#L355-L368

Seems like redstone activation in this case is a bug.
I tryed to prevent redstone activation from projectileHitEvent but I was not able to cast to button like this: `Button button = (Button)hitBlock.getState();  button.setPowered(false);`
So I'm using this metadata way - that is for my point of view a ugly way to solve this. 
I did some testing and for some reason the fix does not work every time.
Sometimes the (german) "Redstone gestoppt" system out debug msg was logged to console after the projectile has triggerd the redstone...
And I also started to ignore already pressed buttons (to prevent that this could be abuse to unpower buttons) but anyway this seems to not make any difference.
But yes im realy new to java oop and spigot at all.

The same should be done for pressure plates - but keep in mind: they will also aktivated if someone drops items on it. 

Best regards & thank you in advance

DavidBla

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -7,6 +7,8 @@ import com.intellectualcrafters.plot.config.Settings;
 import com.intellectualcrafters.plot.flag.Flags;
 import com.intellectualcrafters.plot.flag.IntegerFlag;
 import com.intellectualcrafters.plot.object.*;
+import com.intellectualcrafters.plot.object.Location;
+import com.intellectualcrafters.plot.object.PlotPlayer;
 import com.intellectualcrafters.plot.util.*;
 import com.plotsquared.bukkit.BukkitMain;
 import com.plotsquared.bukkit.object.BukkitLazyBlock;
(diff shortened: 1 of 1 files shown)


david-bla:
I think this issue is already solved with some better (for sure) solution.
I was not able to find a specific commit but I checked the behavior on a newer version and it seems like it is fixed. I'll do some additional checks (tripwire / pressure plates etc) maybe I'll raise issues instead contributing directly because Id not get any response about half a year.

PR #2095, opened 2018-09-12
Fix /p deny not kicking players in merged plots

RickyB98:
As the title says, made the /p deny command look for equality with connected plots so as to kick the denied player to spawn even if they are in one of the merged (not main) plot.


N0tMyFaultOG:
Can someone test the legacy version and confirm the issue persists still? There is no issue related to this "fix".

N0tMyFaultOG:
I was unable to find an issue with denying players in merged plots. If you experience that issue still update plotsquared and open a new pull request following the template.

PR #2135, opened 2018-10-16
Update PlayerEvents.java

SrLeojaco:
Implement isCanceled to being able to censor a message using external plugin. Actually if the message is sent to plot chat, it is ignored the cancelation.


dordsor21:
You should also implement this for the Sponge and Nukkit versions.

SrLeojaco:
> You should also implement this for the Sponge and Nukkit versions.

As i only work with bukkit/spigot, i don't know events for sponge/nukkit

dordsor21:
#2142 includes nukkit/sponge

PR #2158, opened 2018-11-08
Upgrade Plotsquared Core to Worldedit 7.0

Narimm:
This upgrades the PlotSqaured Core dependency on WE to version 7

Also add the spigot required api-version to plugin.yml

The Nukkit and Sponge modules wont compile for me due to dependencies that I dont believe are 1.13.2 compatable. But I have limited experience with those servers.

I am aware this means that this branch probably cannot be merged.    However  I do believe that the core module would now server and 1.13 compatible.

### DIFF
--- a/Bukkit/build.gradle
+++ b/Bukkit/build.gradle
@@ -7,14 +7,14 @@ repositories {
 
 dependencies {
     compile project(':Core')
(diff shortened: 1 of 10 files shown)


Narimm:
Ok so this was premature...i didnt realize how much you had used byte values in the Bukkit module...

I will close this and resubmit if I get close otherwise ... goodluck!

PR #2238, opened 2019-01-16
Fixed plot merge keep road without permission

thenilsdev:
Fixes the problem that you do not need an explicit permission to merge two plots without removing the road. This allows you to break down the road and keep the resources.


thenilsdev:
```
*User* issued server command: /p merge auto removeroad
```
image

For wiki: https://github.com/IntellectualSites/PlotSquared/wiki/Commands-and-Permissions
- plots.merge.terrain: Keep terrain of road when doing "/p merge auto false"

Issue #2249, opened 2019-01-23
Bypassing the world border


theJ3H:
# Bug report template 

**Debug paste link:** 

https://incendo.org/paste/view/294b0057704d4c7fba19c7c73e3ab639
**Description of the problem:** 
A player can get past the world border by not running directly at it. Instead you have to fly agains the border with a certain angle or move forward while also moving to the side.

https://youtu.be/g63wq1iNfTw

**How to replicate:** 

Look at the video.
You can also attempt to get past the border by looking directly at the border and pressing W and A at the same time.
**Checklist**:

- [x] I included a `/plot debugpaste` link
- [x] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] I made sure I am using an up-to-date version of PlotSquared
- [x] I made sure the bug/error is not caused by any other plugin

MattBDev:
I feel like this is related to one of those old directional issues minecraft use to have with certain things like redstone.

AeSix:
Your debugpaste has been removed from hastebin.  This is why the devs have moved debugpaste to a new site, which they control.  Please update P2, test again.  If the issue perissts, update your original post with a new debugpaste and any additional info you can provide. 

If for some reason, an update manages to fix this issue, please update and close :)

AeSix:
I suggest using a newer build from https://ci.athion.net/view/Everything/job/PlotSquared/
Test. If it's still an issue, update the ticket so the devs can have more accurate information. 
Create a new debugpaste as well, please. 

If this fixes the issue, update and close the ticket please.

Be aware that the devs may not be able to resolve this issue as it may be a core issue with Minecraft 1.8.8.  You may want to test older builds, in the event that something changed with newer versions of Minecraft which had to be accounted for in newer builds of P2.

theJ3H:
This is still a problem in the latest dev build. I've updated the OP with the new debug paste.

In addition to the issue described in the OP it is also possible to bypass the border using enderpearls.

theJ3H:
https://github.com/theJ3H/PlotSquared/commit/48596dd7e2bf5b7843bd7f5d5dec35e1fa993220

Something like this seems to fix the issue shown in the video. But I can't figure out how to fix the enderpearl issue. A player can throw a pearl so that it will always land on a plot, but apparently the border is only checked if a player is moving over a road.

[...]

PR #2263, opened 2019-02-04
Players are able to bypass border

theJ3H:
Currently players are able to bypass the worldborder as described in #2249 
This can be done by  using sprint flight agains the border while flying at a specific angle. It is also possible to bypass the border at the edges. Another way is to use enderpearls.

This PR forces players back inside the border instead of just teleporting them a block back. (first commit)

The second commit is needed so that enderpearl teleports are properly detected. A few returns stopped the processing before the border would be checked, or stopped the processing before the z axis would be checked.

With these changes I was no longer able to bypass the border using creative flight, enderpearls and the minecraft tp command.


Sauilitired:
Someone did an ooosie. Good stuff, cheers mate =)

PR #2289, opened 2019-03-03
Add leaf-decay flag

AlecDusheck:
Add a plot flag called "leaf-decay" which disables leaves decaying on plots.


AlecDusheck:
Actually just noticed that this code only applies to the Bukkit version? Would you guys like this change to be added to the Sponge and Nukkit version as well?

dordsor21:
Yes, please add to sponge/nukkit as well.

N0tMyFaultOG:
This sounds like an interesting idea, unfortunately, there are issues with merging it and you didn't apply the requested changes. Next time you open a pull request, please ensure to follow its template.

PR #2350, opened 2019-05-02
Remove members when buying a plot + Fix /plot delete issue #2351

UnioDex:
Fixes that: https://github.com/IntellectualSites/PlotSquaredSuggestions/issues/48


MattBDev:
The database needs to be called to properly remove all these players.

UnioDex:
> The database needs to be called to properly remove all these players.

Changed the way to remove players. I think it will work now.

UnioDex:
I added some commits to fix #2351

MattBDev:
Squash the first two commits into one and do the same for the two that follow. Then I'll merge it.

UnioDex:
Here you go:
https://github.com/IntellectualSites/PlotSquared/pull/2353
 @MattBDev

Issue #2375, opened 2019-05-28
Plot merging/unmerging not working correctly


alexschrod:
**[REQUIRED] PlotSquared Version Number:** PlotSquared-4.277

**[REQUIRED] Spigot/Paper Version Number:** Paper #624 for MC 1.13.2

**[REQUIRED] Minecraft Version Number:** 1.13.2

**Links to worlds.yml file and settings.yml file:** 
* worlds.yml: https://pastebin.com/QP9jrDKR
* settings.yml: https://pastebin.com/T7n3Vb82

**[REQUIRED] Description of the problem:**
Roads and plot borders do not always get correctly restored when plots are merged and unmerged, the same goes for biomes. Biomes also seem to misbehave during plots being cleared.

**Any relevant console output or screenshots:**
Video showing merging and unmerging, claiming and clearing, with the issues at hand: https://youtu.be/CjJXNw6jfi4?t=125 (content before this time is setting up the world, can be watched if interested -- world uses all default values)

**Plugins being used on the server:**
* Advanced Portals
* Chestshop
* Dynmap
* DiscordSRV
* editable signs
* enjinminecraftplugin
* essentialsX
* essentialsXChat
* essentialsXSpawn
* FastLeafDecay
* LuckPerms-Bukkit
* Matrix
* Mycommand
* NuVotifier
* PlayerHeads
* plot2dynmap
* PlotSquared
* PluginConstructorAPI
* ProtocolLIB
* serverlistMOTD
* serverrestorer
* simple egg
* skript
* u-chat
* vault
* worledit
* Worldguard

**How to replicate:** (timestamps for point in video this is done)
* Create new plot world
* Claim a plot (/plot claim) 2:13
* Claim another plot 2:22
* Merge plots 2:28
* Unmerge plots 2:35
* _Observe road not being correctly restored_
* Merge again 2:44
* Unmerge again 2:46
* _Observe road being correctly restored_
* Merge again 2:50
* Unmerge again 2:56
* _Observe road not being correctly restored again_
* Claim another plot 3:08
* Set biome to jungle 3:16
* Clear plot 3:21
* _Observe plot's biome only being partially cleared_
* Set biome to jungle 3:33
* Merge jungle plot with cleared plot 3:39
* _Observe biome where road used to be not updating_
* Merge with first plot 3:49
* _Observe once more biome where road used to be not updating_
* Set biome to jungle 3:53
* _Observe whole plot taking on the correct biome_
[...]

PR #2378, opened 2019-05-30
Plot schematic fix

alexschrod:
These changes fixes various issues with using plot schematics and `paste-on-top` together, although it may also fix other bugs with a related cause. These issues were discovered while working on fixing other things mentioned in #2375, but aren't really shown or mentioned therein.

* Due to a conflict between using `BaseBlock` and `PlotBlock` at the same time in the same `BukkitLocalQueue`: https://github.com/IntellectualSites/PlotSquared/blob/ac5cdfae648ce3f76ed3d036a7444a1ec5ea6fef/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/util/block/BukkitLocalQueue.java#L70-L76 when using a schematic (which uses `BaseBlock`), all other block operations (which use `PlotBlock`) would be silently ignored. I solved this by converting `PlotBlock`s to `BaseBlock`s, since the latter has more features/is a superset of the former.
* Blocks above `WALL_HEIGHT` that were air in the schematic would not be removed/replaced when clearing a plot. This meant that any blocks in the schematic that were air, would remain whatever they were,  which is not great for a clear operation. Whatever blocks were not specified in the schematic are now turned into air: https://github.com/IntellectualSites/PlotSquared/blob/fbc80fe6931a464cb276268e3f869987f80cf78f/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java#L105-L110

### DIFF
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java
@@ -81,6 +81,7 @@ public class HybridPlotManager extends ClassicPlotManager {
         } else {
             minY = 1;
         }
+        BaseBlock airBlock = BlockTypes.AIR.getDefaultState().toBaseBlock();
         for (int x = pos1.getX(); x <= pos2.getX(); x++) {
             short absX = (short) ((x - hpw.ROAD_OFFSET_X) % size);
             if (absX < 0) {
@@ -98,7 +99,7 @@ public class HybridPlotManager extends ClassicPlotManager {
                             queue.setBlock(x, minY + y, z, blocks[y]);
                         } else {
                             // This is necessary, otherwise any blocks not specified in the schematic will remain after a clear
-                            queue.setBlock(x, minY + y, z, BlockTypes.AIR.getDefaultState().toBaseBlock());
+                            queue.setBlock(x, minY + y, z, airBlock);
                         }
                     }
                 }
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java
@@ -11,7 +11,7 @@ import com.github.intellectualsites.plotsquared.plot.util.block.GlobalBlockQueue
 import com.github.intellectualsites.plotsquared.plot.util.block.LocalBlockQueue;
 import com.google.common.collect.Sets;
 import com.sk89q.worldedit.world.block.BaseBlock;
-
+import com.sk89q.worldedit.world.block.BlockTypes;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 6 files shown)


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridPlotManager.java
```diff
@@ -103,6 +104,9 @@ private void createSchemAbs(HybridPlotWorld hpw, LocalBlockQueue queue, Location
                     for (int y = 0; y < blocks.length; y++) {
                         if (blocks[y] != null) {
                             queue.setBlock(x, minY + y, z, blocks[y]);
+                        } else {
+                            // This is necessary, otherwise any blocks not specified in the schematic will remain after a clear
+                            queue.setBlock(x, minY + y, z, BlockTypes.AIR.getDefaultState().toBaseBlock());
```
alexschrod: This air block should probably be extracted out into a local variable outside the loop for the sake of performance, right?

MattBDev: I would think the java compiler would do this for us but I'm not going to pull this change until late Friday. Go ahead and make it a variable in another commit if you want.

alexschrod: Alright, so I looked into the WE code, and it wouldn't be all that inefficient, since there's caching in there, but I still think Java has to make the method calls each time, since I'd be amazed if it could somehow deduce that there aren't side effects in them, so this should in theory save some method calls at least.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/block/BasicLocalBlockQueue.java
```diff
@@ -116,32 +118,18 @@ public final boolean execute(final LocalChunk<T> lc) {
...
-                    return true;
-                }
-                this.blockChunks.put(pair, previous);
-                lastWrappedChunk = previous;
-            }
+        // Trying to mix PlotBlock and BaseBlock leads to all kinds of issues.
+        // Since BaseBlock has more features than PlotBlock, simply convert
+        // all PlotBlocks to BaseBlocks
+        if (id instanceof StringPlotBlock) {
+            StringPlotBlock stringPlotBlock = (StringPlotBlock) id;
```
Sauilitired: It might be worth checking for (and storing) the cached base block in `StringPlotBlock` (https://github.com/IntellectualSites/PlotSquared/blob/41299e685156311d74661fb9aefa90325dcd47c8/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/StringPlotBlock.java#L18) 

This is stored initially when getting a StringPlotBlock from a base block.

MattBDev:
Don't worry about the conflicts in the pull request. Once the comment @Sauilitired made is addressed I'll resolve them and merge.

alexschrod:
@MattBDev Ah, sorry, I was already in the middle of a rebase when you wrote that comment, so... already resolved, I suppose? 😅 

I'll address @Sauilitired 's suggestion shortly.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/block/BasicLocalBlockQueue.java
alexschrod: I'm honestly a little embarrassed about how I did this using `instanceof` and down-casting rather than the more sensible abstract method approach. With the new approach in my latest commit, both StringPlotBlock and LegacyPlotBlock caches BaseBlock lookups.

PR #2382, opened 2019-05-31
PlotManager + PlotArea cleanup

alexschrod:
I have made changes that can be summed up as these three key points:
* `PlotManager`s now hold a reference to the `PlotArea` they are managing.
* Removed `PlotArea` from all `PlotManager` methods, as the `PlotManager` has a reference to its `PlotArea`, as per the previous point.
* Moved `PlotManager` creation from the `PlotGenerator` to the `PlotArea`, i.e. a `PlotArea` now creates its own `PlotManager`. This lets it pass on the most specific `PlotArea` subclass possible to its `PlotManager`, greatly reducing the need for down-casting or instanceof checks everywhere.

The constant casting (and occasional use of instanceof) indicated to me that there was a big code smell going on with these classes, and the changes I've made have all but eliminated it. In theory, what I've done shouldn't change anything about the functionality of anything, although it is of course possible I've missed something, in which case, let me know.

There are some remaining undesirable down-castings and instanceof uses in other places, which indicates that there are likely leaky abstractions and poor encapsulation still remaining in the code.

### DIFF
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugRoadRegen.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugRoadRegen.java
@@ -2,11 +2,9 @@ package com.github.intellectualsites.plotsquared.plot.commands;
 
 import com.github.intellectualsites.plotsquared.commands.CommandDeclaration;
 import com.github.intellectualsites.plotsquared.plot.config.Captions;
-import com.github.intellectualsites.plotsquared.plot.generator.HybridPlotManager;
-import com.github.intellectualsites.plotsquared.plot.generator.HybridPlotWorld;
 import com.github.intellectualsites.plotsquared.plot.object.Location;
 import com.github.intellectualsites.plotsquared.plot.object.Plot;
-import com.github.intellectualsites.plotsquared.plot.object.PlotArea;
+import com.github.intellectualsites.plotsquared.plot.object.PlotManager;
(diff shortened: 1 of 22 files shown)


MattBDev:
Wow. I've been meaning to do this stuff forever.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugRoadRegen.java
```diff
@@ -18,20 +16,16 @@
 
     @Override public boolean onCommand(PlotPlayer player, String[] args) {
         Location loc = player.getLocation();
-        PlotArea plotArea = loc.getPlotArea();
-        if (!(plotArea instanceof HybridPlotWorld)) {
-            return sendMessage(player, Captions.NOT_IN_PLOT_WORLD);
-        }
+        PlotManager manager = loc.getPlotArea().getPlotManager();
```
MattBDev: Isn't it likely that this may cause a NPE since not every location is going to contain a plot area?

alexschrod:
@MattBDev damn, talk about pulling the rug out from under me with those latest commits you've done there.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugRoadRegen.java
alexschrod: Ah, a detail I had not considered. Should be fixed with my latest commit.

PR #2452, opened 2019-08-04
fix .git directory in build.gradle

andrewgazelka:
for me it referenced ~/.gradle/daemon/5.4/.git which obviously is not
right


N0tMyFaultOG:
Closed in https://github.com/IntellectualSites/PlotSquared/commit/7542dbeab8a5e90f9461575f06d4fe3b6539b21c

Issue #2412, opened 2019-07-01
Time stuck when switching Worlds


zachrocks22:
__*NOTICE: Bukkit/Spigot versions 1.7.10 to 1.12.2 are considered legacy and will receive limited support. Please consider upgrading to 1.13 for future support. Plugins exist for 1.13+ that bring back old behaviors found in 1.8*__
# Bug report template 

**[REQUIRED] PlotSquared Version Number:** v4.298

**[REQUIRED] Spigot/Paper Version Number:**  Paper Build #119

**[REQUIRED] Minecraft Version Number:** 1.14.4

**[REQUIRED] Description of the problem:** When setting a time flag in a plot, then switching worlds while still in that plot, the time of day gets stuck.

**Any relevant console output or screenshots:** None! There are no stack-traces thrown. :(

**Plugins being used on the server:** Using Plotsquared (Latest Release) with Multiverse (v3.0.0 from jekins), PerWorldInventories (Latest), GriefPrevention (Latest), WorldEdit (Latest Beta from Bukkit), and WorldGuard (Latest Beta from Bukkit).

**How to replicate:** 
1) Claim a plot
2) Set a time flag (/plot flag set time 500)
3) Switch worlds or teleport to a different world while in the plot.
4) Time is now stuck until the player goes back to the plotworld and exits the plot or relogs.

**Checklist**:
- [X] I included all information required in the sections above
- [X] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
    ( I did find this issue: https://github.com/IntellectualSites/PlotSquared/issues/2357, but could not find a conclusive fix)
- [X]  I made sure I am using an up-to-date version of PlotSquared
- [X] I made sure the bug/error is not caused by any other plugin

RedstoneFuture:
Did you test this situation before the 1.14 update?

zachrocks22:
Yes. The same issue occurred on 1.13 paper spigot.

AeSix:
https://github.com/IntellectualSites/PlotSquared/issues/2357

zachrocks22:
Just wanted to reconfirm for latest 1.14.4 paper (Build 160) and latest PlotSquared (Build 317)

[...]

PR #2460, opened 2019-08-11
Add toggle for time

jewome62:
Add a preference to disable/enable time change from plot flag.

### DIFF
--- a/Nukkit/src/main/resources/plugin.yml
+++ b/Nukkit/src/main/resources/plugin.yml
@@ -119,6 +119,7 @@ permissions:
       plots.toggle: true
       plots.toggle.titles: true
       plots.toggle.chat: true
+      plots.toggle.time: true
       plots.set.biome: true
       plots.set.home: true
       plots.set.alias: true
--- a/Bukkit/src/main/resources/plugin.yml
+++ b/Bukkit/src/main/resources/plugin.yml
@@ -126,6 +126,7 @@ permissions:
       plots.toggle: true
       plots.toggle.titles: true
       plots.toggle.chat: true
+      plots.toggle.time: true
       plots.set.biome: true
       plots.set.home: true
       plots.set.alias: true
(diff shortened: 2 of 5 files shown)


Nukkit/src/main/resources/plugin.yml
```diff
@@ -119,6 +119,7 @@ permissions:
       plots.toggle: true
       plots.toggle.titles: true
       plots.toggle.chat: true
+      plots.toggle.time: true
```
N0tMyFaultOG: Is redundant, breaking doesn't build Nukkit artifacts :P

MattBDev:
This whole idea seems half baked. How is this different from the current time flag that is already present?
Is this supposed to constantly keep the plot set to a current time all the time? If so, it fails to do it properly.

AeSix:
@MattBDev  maybe something to do with https://github.com/IntellectualSites/PlotSquared/issues/2412
Is this correct @jewome62 ?

Nukkit/src/main/resources/plugin.yml
Sauilitired: In the case that this ever happens, this doesn't hurt. Not that it will, but let's pretend it might.

Sauilitired:
My opinions on this, copied and pasted from Discord:
> The disable time thing looks very similar to the disabling of titles, but for plot time. I think it’s a fair change to implement, as it doesn’t break anything
It simply allows players to disable their time getting updated when entering a plot. This is useful for many reasons, but I can see players using shaders appreciating it even more, as playing in complete darkness isn’t always a lot of fun

jewome62:
Yes, I confirm, that's for people who don't want theirs screen flashes with day/night when they fly over many plot.

jewome62:
@MattBDev The goal is let user (with permission to do it) to do not suffer change of aggressive change of time
- As admin and moderator who are often on server
- visually impaired
- for shaders

My work is currently tested on my server without problem

MattBDev:
If you can confirm it works, I won't stand in the way of this being merged.
I'll leave this to @N0tMyFaultOG to either close this or merge it.

jewome62:
Wiki updated
Thanks

PR #2470, opened 2019-08-18
Item Frame Problem

Hajatoby:
Everyone can steal from Item Frames items, although no one is drenched. Nukkit


MattBDev:
You submitted to the wrong branch and included a ton of unnecessary commits.

PR #2517, opened 2019-10-01
Improve world generator diagnosis

N0tMyFaultOG:
## Overview

**Fixes https://github.com/IntellectualSites/PlotSquaredSuggestions/issues/62**

## Description
Can be obtained from the suggestion

## Checklist

### DIFF
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugPaste.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugPaste.java
@@ -117,7 +117,7 @@ import java.util.stream.Collectors;
 
                 try {
                     final File MultiverseWorlds =
-                            new File(PlotSquared.get().IMP.getDirectory(), "../../plugins/Multiverse-Core/worlds.yml");
+                            new File(PlotSquared.get().IMP.getDirectory(), "../Multiverse-Core/worlds.yml");
                     incendoPaster
                             .addFile(new IncendoPaster.PasteFile("MultiverseCore/worlds.yml", readFile(MultiverseWorlds)));
                 } catch (final IOException ignored) {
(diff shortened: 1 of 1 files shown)


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugPaste.java
```diff
@@ -115,6 +115,15 @@ private static String readFile(@NonNull final File file) throws IOException {
                         "&cSkipping PlotSquared.use_THIS.yml because it's empty");
                 }
 
+                try {
```
Sauilitired: The command description should be altered to mention this file as well, to not cause any further issues with people claiming we steal their data. Last time wasn't very fun.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/commands/DebugPaste.java
```diff
@@ -115,6 +115,15 @@ private static String readFile(@NonNull final File file) throws IOException {
...
+                    final File MultiverseWorlds =
+                            new File(PlotSquared.get().IMP.getDirectory(), "../../plugins/Multiverse-Core/worlds.yml");
```
Sauilitired: Why is this going back two folders, to then re-enter plugins?

N0tMyFaultOG: Having `/plugins/Multiverse-Core/worlds.yml` does not lead to the expected result

MattBDev: Getting in certain directories with Bukkit has always been a pain.

N0tMyFaultOG: I had no luck doing it in a better way, so I used that. If you have a better AND working solution, hmu @Sauilitired :^)

Sauilitired: Lol that wasn’t my point, but if you go backwards two steps you’re already leaving the plugin folder, then you directly re-enter it. What would be different by just doing ../Multiverse-Core?

N0tMyFaultOG: That works! I was not entirely sure how the paths behaved so I just wanted to be sure it works ^^

PR #2533, opened 2019-10-09
Place flag bugfix (fixxes issue #2524)

### LINKED ISSUE #2524: Blocks in 'place' flag only placeable if clicked block is also in flag
**Debug paste link:** 
https://athion.net/ISPaster/paste/view/da7449b0efc14e5c9b20249b5171ae49
**[REQUIRED] Spigot/Paper Version Number:** 
CraftBukkit version git-Spigot-94af569-c2d1201 (MC: 1.14.4) (Implementing API version 1.14.4-R0.1-SNAPSHOT)  [latest version]

**[REQUIRED] Minecraft Version Number:** 1.14.4

**[REQUIRED] Description of the problem:** 
If you try to place a block on a plot where you are not a member, but where the block type is in the 'place' flag, you can only place it if the block that you clicked at is _also_ in the 'place' flag. 

If you click on a block which is not in 'place', the message "Have an admin set the flag: place" is shown.
If you click on a block which is in 'place', but the block in your hand is not, the message "You are lacking the permission node: plots.admin.build.other" is shown.
Only if both the block in your hand and the one you click at is in 'place', it can be placed. 

Note that while the above debugpaste shows the current release of PlotSquared (4.353), the bug also happens with the current github version - I normally use a custom version which branched at 4.386 (but the stuff I added - #2506, and some new flag which I might turn to a pull request later when it works as intended, see my fork if you are curious - does not concern block placement).

I tracked it to https://github.com/IntellectualSites/PlotSquared/blob/breaking/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java#L2139: here lb is the block that was interacted with, i.e. the one clicked at, and EventUtil.manager.checkPlayerBlockEvent uses the type of lb when checking the 'place' flag.   

Since 'place' is checked later anyway (in https://github.com/IntellectualSites/PlotSquared/blob/breaking/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java#L3005), maybe the check in onInteract can be removed?

**Plugins being used on the server:**
see debugpaste above, none should be relevant for this bug.

**[REQUIRED] How to replicate:** 
* add some block to the place flag of a given plot (e.g. `/plot flag add stone`)
* place some of that block on the plot (if not already present), make sure that there are other 
block types reachable (e.g. grass_block)
* switch to a user that is not member/trusted on the plot (you could also do the above steps as op on a plot that is not yours, and then deop yourself)
* try to 
** place the block in place on blocks of  different type, which is not in 'place' --> error message shown (but this should imho work)
** place it on a block with the same type --> can be placed
** place a different block (which is not in 'place') on the block in 'place' --> different error message shown (which is correct)

**[REQUIRED] Checklist**:

mindw0rm:
## Overview
Fixes the problem with the "place" flag as described in  #2524 (or below) .

**Fixes #2524**

## Description
Players who are not trusted in a block are supposed to be able to place blocks that are added to the "place" flag.
However, due to a Bug in [PlayerEvents](https://github.com/IntellectualSites/PlotSquared/blob/breaking/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java), the wrong material (namely that of the block that was clicked on) is passed to EventUtil.checkPlayerBlockEvent.

This pull request fixes the issue by changing the material of the block before it is passed to EventUtil.checkPlayerBlockEvent. This is done be simply moving a line some lines up.

It is only changed if the action passed to EventUtil is PLACE_BLOCK, [where](https://github.com/IntellectualSites/PlotSquared/blob/breaking/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/EventUtil.java#L229) the block is only used to check against the "place" flag, which is exactly what we want. There should be no side effects, as far as I can tell, since the block isn't used anywhere else (and if the condition will not trigger, the material change would have happened anyway). 

## Checklist

### DIFF
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -2042,14 +2042,14 @@ import java.util.regex.Pattern;
                 if (type == Material.AIR) {
                     type = offType;
                 }
+                // in the following, lb needs to have the material of the item in hand i.e. type
+                lb = new BukkitLazyBlock(PlotBlock.get(type.toString()));
                 if (type.isBlock()) {
                     location = BukkitUtil
                         .getLocation(block.getRelative(event.getBlockFace()).getLocation());
                     eventType = PlayerBlockEventType.PLACE_BLOCK;
-                    lb = new BukkitLazyBlock(new StringPlotBlock(type.toString())); 
                     break;
                 }
-                lb = new BukkitLazyBlock(PlotBlock.get(type.toString()));
                 if (type.toString().toLowerCase().endsWith("egg")) {
                     eventType = PlayerBlockEventType.SPAWN_MOB;
                 } else {
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -2046,6 +2046,7 @@ import java.util.regex.Pattern;
                     location = BukkitUtil
                         .getLocation(block.getRelative(event.getBlockFace()).getLocation());
                     eventType = PlayerBlockEventType.PLACE_BLOCK;
+                    lb = new BukkitLazyBlock(new StringPlotBlock(type.toString())); 
                     break;
                 }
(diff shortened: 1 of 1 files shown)


mindw0rm:
You don't need 599ef8a - if you have a way to select the commits, take only 1f4f78a. 
The both do the same, but the latter is much more elegant ;)

MattBDev:
I'll approve this as I can see no potential side effects. If any bugs occur before release, hopefully one of our breaking users will be able to catch it.

PR #2542, opened 2019-10-22
New plot flag untrusted-visit

mindw0rm:
## Overview

This adds a new flag 'untrusted-visit' which allows plot owners to make their plot(s) reachable to everyone via the visit command. 

**Fixes IntellectualSites/PlotSquaredSuggestions#64**

## Description
Besides adding the new flag itself, it is checked in two locations:
* in the Visit command: to allow visiting if the player is not trusted and doesn't have the permission `plots.visit.other` (i.e. main functionality of the flag)
* in the onTeleport event, where `deny-teleport` is checked.
  The reasoning here is that you'll want `untrusted-visit` to override `deny-teleport`, or else it won't make sense to set both flags. With this change you can prevent players to teleport within your plot (at least with my bugfix #2506 or equivalent changes), but still allow visits (denying teleports is mandatory for things like e.g. labyrinths or jump'n'run courses (else cheating would be possible with ender pearls or chorus fruits), but if they can't use visit, they'll have to walk to the plot). 
Note that this is only implemented for bukkit (because deny-teleport is only checked for bukkit as well, and also because I can't test the others ;) ).

The last change is adding a new method getLocationFull to BukkitUtil, which transforms a bukkit Location into a PlotSquared Location _including_ yaw and pitch, since this is required for the location compare, and adding the copying of yaw and pitch to one of the existing ``getLocation` methods gave me a bad feeling regarding side effects. 

So the flag is considered only under very specific conditions, so it shouldn't have any side effects to other functionality. 

## Checklist

### DIFF
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/listeners/PlayerEvents.java
@@ -697,7 +697,11 @@ import java.util.regex.Pattern;
                 Plot plot = area.getPlot(location);
                 if (plot != null) {
                     final boolean result = Flags.DENY_TELEPORT.allowsTeleport(plotPlayer, plot);
-                    if (!result) {
+                    // there is one possibility to still allow teleportation:
+                    // to is identical to the plot's home location, and untrusted-visit is true
+                    // i.e. untrusted-visit can override deny-teleport
+                    // this is acceptable, because otherwise it wouldn't make sense to have both flags set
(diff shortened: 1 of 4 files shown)


mindw0rm:
Obviously you don't want this. Sorry that I bothered you.

Sauilitired:
It makes everything slightly more complex, but considering the feature is opt-in and won't affect any existing setups, I don't have any objections.

@mindw0rm Thanks for contributing to PlotSquared, I sincerely mean that. I'm sorry that the conversation in the issue report took a sour turn.

MattBDev:
@mindw0rm can you please resolve the merge conflicts?

mindw0rm:
@MattBDev done

PR #2561, opened 2019-11-04
Add text-api to PlotSquared

MattBDev:
Since WE already uses text-api and FAWE will soon follow it would be more consistent to use Text Components in PlotSquared. Some current code handled by PlotMessage.java is easier to understand and write with TextComponents.

Discussion Needed:
Should TextComponents replace all usages of PlotMessage?
Is it worth giving up custom colors in the captions file? or Is there a way to prevent loosing the ability to use custom colors?

### DIFF
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/object/BukkitPlayer.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/object/BukkitPlayer.java
@@ -13,6 +13,8 @@ import com.github.intellectualsites.plotsquared.plot.util.PlotWeather;
 import com.github.intellectualsites.plotsquared.plot.util.StringMan;
 import com.github.intellectualsites.plotsquared.plot.util.UUIDHandler;
 import io.papermc.lib.PaperLib;
+import net.kyori.text.Component;
+import net.kyori.text.adapter.bukkit.TextAdapter;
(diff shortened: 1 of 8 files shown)


N0tMyFaultOG:
> Is it worth giving up custom colors in the captions file? or Is there a way to prevent loosing the ability to use custom colors?

One of the user friendly benefits PlotSquared offers are the lots of ways to customize it. Stripping colors would cut the way to customize the plugin tremendously. If we want to use TextComponents, we should find a way to not remove the ability to edit/color messages to your demand imo.

boy0001:
> "Is there a way to prevent loosing the ability to use custom colors?"

WorldEdit's text component technically supports color, so we'd have to write a converter. You could probably hack something together by looking at how FancyMessage converts from legacy to json.

MattBDev:
`LegacyComponentSerializer.legacy().deserialize("&a&lfoo&9bar", '&')` would be the code you are thinking of Jesse. Unfortunately Legacy Components can't use `HoverEvent `or `ClickEvent` in the text-api. 
I'm not trying to remove custom colors, I just didn't see an easy way forward.

boy0001:
Yes, that works. I don't see why hover events couldn't either..? the deserializer returns a TextComponent capable of it.

Sauilitired:
Clearing the milestone for this one until it's no longer marked as WIP.

PR #2564, opened 2019-11-10
Use patterns instead of BlockBucket

boy0001:
## Overview
Use WorldEdit Patterns instead of BlockBuckets. This allows for fun things like noise patterns, scripting and more. 

## Description
The BlockBucket class has been refactored to just handle the config reading/writing (since that's not implemented by the WorldEdit pattern class)

It uses the WorldEdit blacklist instead of PlotSquared's right now. (Since WorldEdit does the parsing).

Needs testing.

## Checklist

- [] I ensured my changes do not break other parts of the code


MattBDev:
:O

PR #2578, opened 2019-11-14
Add hotbar message

boy0001:
## Overview
Uses hotbar for plot entry/exit messages. No idea if this is what we want as the default behavior.

**Fixes IntellectualSites/PlotSquaredSuggestions#25 **

## Checklist


N0tMyFaultOG:
Hm, the titles might be big but I'd not get a rid of them fully. A hotbar message seems nice to have, but I'd not replace the titles with it, dunno, but I guess people use them. Maybe add a boolean config option so people can choose what to use?

PR #2652, opened 2020-01-01
Revert teleportation behavior to pre-4.400

### LINKED ISSUE #2651: Revert teleportation behavior to pre-4.400
# Bug Report Template:

### Server type:

- [X] Spigot / Paper *(CraftBukkit should not be used, re-test with Spigot first!)*

### Minecraft Version:
- [X] Minecraft 1.15

### Server build info: 

```
This server is running Paper version git-Paper-31 (MC: 1.15.1) (Implementing API version 1.15.1-R0.1-SNAPSHOT)
```

### PlotSquared Version:
PlotSquared-4.436

### Description of the problem:

Since [these](https://github.com/IntellectualSites/PlotSquared/commit/20dd98d3b9686fcd29586219f77397f588a3710e#diff-43d7174f7dd247049a836b624807d29dR194) [commits](https://github.com/IntellectualSites/PlotSquared/commit/68a5166d2f2ad342436f4683afac04dc18f70888#diff-43d7174f7dd247049a836b624807d29dR192), some teleports performed by PS have had their TeleportCause changed. This has created some unexpected behavior in plugins listening to the teleport event and expecting the old COMMAND cause.

# AFFIRMATION OF COMPLETION:

- [] I included all information required in the sections above

zombachu:
## Overview

**Fixes #2651**

## Description
See issue

## Checklist


MattBDev:
The issue here is the fact that the cause of a teleport here isn't always a command. It makes more sense for the cause to be a plugin. Can you share an example where the cause needs to be a command?

zombachu:
The specific case I'm running into is that the /back command my server uses no longer properly works with PS commands but as a matter of principle I thought it would be better to fix it in the source that changed rather than just my end.

See my most recent commit for a better resolution for this issue. I added optional methods that allow for the specification of the cause on a case-by-case basis. This should have backwards compatibility with any users of the API, though if anyone is reimplementing any of the interfaced I touched for whatever reason then they're going to need to update.

PR #2669, opened 2020-01-22
Integrate placeholders from PlotSquared-Expansion

N0tMyFaultOG:
This PR makes https://github.com/IntellectualSites/PlotSquared-Expansion redundant for v4 and upcoming versions and bundles placeholders within PlotSquared itself. This way, you don't need to download the expansion anymore, P2 will register placeholders itself.


N0tMyFaultOG:
Keep this up for v5

PR #2693, opened 2020-02-19
Flags remake

Sauilitired:
Flag Changes:
- Complete remade the flag API and framework
- Switched flag storage from plot_settings to plot_flags
- Remade the flag command
- loads of minor fixes and changes...


N0tMyFaultOG:
I think the table conversion should be a part of a bigger conversion from v4 to v5, so we'll figure out the conversion later on. Unless someone feels like committing it to the flag branch

PR #2694, opened 2020-02-20
Pull/2693

Sauilitired:
Flag changes.


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/SchematicHandler.java
```diff
@@ -152,6 +146,8 @@ public void paste(final Schematic schematic, final Plot plot, final int xOffset,
             }
             try {
                 // Set flags
+                /*
+                TODO: Revisit
```
MattBDev: Revisit when? Is this something that will be worked on after the v5 release?

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/SchematicHandler.java
```diff
@@ -451,6 +452,8 @@ public abstract void getCompoundTag(String world, Set<CuboidRegion> regions,
     public void getCompoundTag(final Plot plot, final RunnableVal<CompoundTag> whenDone) {
         getCompoundTag(plot.getWorldName(), plot.getRegions(), new RunnableVal<CompoundTag>() {
             @Override public void run(CompoundTag value) {
+                /*
+                TODO: Revisit
```
MattBDev: See other comment above.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/flags/FlagContainer.java
```diff
@@ -0,0 +1,251 @@
...
+    public enum PlotFlagUpdateType {
+        /**
+         * A flag was added to a plot container
+         */
+        FLAG_ADDED,
+        /**
+         * A flag was removed from a plot container
+         */
+        FLAG_REMOVED
+    }
```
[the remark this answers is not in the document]
Sauilitired: I've added a FLAG_UPDATED enum. It was a bit of an oversight.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/SchematicHandler.java
Sauilitired: Right. We need to determine whether this is something we want to support at all. What is the purpose of storing plot flags in the schematics to begin with? We should discuss this in the developer channel on Discord.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/flags/InternalFlag.java
```diff
@@ -0,0 +1,4 @@
+package com.github.intellectualsites.plotsquared.plot.flags;
+
+public interface InternalFlag {
```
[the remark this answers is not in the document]
Sauilitired: Plugins should be able to create their own internal flags. I'll add some documentation
that explains what the interface is supposed to represent (already present in the wiki).

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/PlotSettings.java
```diff
@@ -154,10 +104,6 @@ public String getLeaveMessage(PlotArea plotArea) {
             .collect(ImmutableList.toImmutableList());
     }
 
-    void setComments(List<PlotComment> comments) {
-        this.comments = comments;
-    }
-
```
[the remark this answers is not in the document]
Sauilitired: Lombok setter (as all other simple setters are lombokified)

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/util/SchematicHandler.java
MattBDev: I see no reason we should be saving flags inside schematics because it isn't part of the schematic standard and we already have issues with schematics as is.

Issue #2607, opened 2019-11-26
Plot info showing the wrong biome


SecretlyJealous:
# NOTICE: 
```diff
! PlotSquared for Minecraft Java Edition versions between 1.7 through to 1.12.2 are considered
! legacy, and will receive limited to no support. Please consider upgrading to 1.13+ for
! future support.  Plugins exist for 1.13+ which bring back behaviors found in 1.8.8
! All versions of PlotSquared for Sponge and Nukkit(X) will recieve limited to no support
! due to lack of developer interest and time. Additionally, NukkitX has not had feature
! updates since the Better Together, which prevents some PlotSquared features from ever
! functioning.  Contributions are always welcome however!
```

**Feature requests & Suggestions are to be submitted at the [PlotSquared Suggestions tracker](https://github.com/IntellectualSites/PlotSquaredSuggestions)**

**Code contributions are to be done through [PRs](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request), tagging the specific issue ticket(s) if applicable.**

**[DISCORD INVITE LINK](https://discord.gg/cSMxtGn)** and please, for the love of the little sanity we have left, use the correct channels!

# BUG REPORT TEMPLATE:

## Required Information section:
> ALL FIELDS  IN THIS SECTION ARE REQUIRED, and must contain appropriate information
### Server config info (Debug/file paste links):
worlds.yml: https://pastebin.com/4QUjDJHn
settings.yml https://pastebin.com/JQu1Vraj
PlotSquared version: PlotSquared-4.411

### Server type:
**select one**

- [X] Spigot / Paper *(CraftBukkit should not be used, re-test with Spigot first!)*
- [] Sponge *- NOTE: NOT ACTIVELY MAINTAINED*
- [] NukkitX *- NOTE: NOT ACTIVELY MAINTAINED*

### Minecraft Version:
**select one**

- [X] Minecraft 1.14.4 ***This is the ONLY actively maintained target for PlotSquared***
- [] Minecraft 1.13.2
- [] Minecraft Java Edition *other versions, please specify*:
- [] Minecraft Bedrock Edition *specify version*:

### Server build info: 

``` git-Paper-226 (MC: 1.14.4) ```

### WorldEdit/FAWE versions:

- [] FAWE *version*:
- [X] WorldEdit *version*: worldedit-bukkit-7.0.1

### Description of the problem:
[...]

PR #2698, opened 2020-02-23
Make getBiome return a completable future and use PaperLib to load th…

Sauilitired:
…e chunk asynchronously.

## Overview
This PR will ensure that the chunk is loaded before trying to read the biome, which
should fix issues where the world returned the wrong biome.

**Fixes #{issue number}**
#2607

## Description
PaperLib is used to read the chunk asynchronously. A block is then fetched at the query location and the biome is collected from the block instance.

## Checklist


Sauilitired:
Nevermind, this was a horrible idea :laughing:

Issue #2603, opened 2019-11-25
Tile entity content not moved over


DevJoey:
# Bug Report Template:

## Required Information section:
> ALL FIELDS  IN THIS SECTION ARE REQUIRED, and must contain appropriate information
### Server config info (Debug/file paste links):

https://athion.net/ISPaster/paste/view/1d4d53f293a34d09b189791b15da41e2

### Server type:
**select one**

- [x] Spigot / Paper *(CraftBukkit should not be used, re-test with Spigot first!)*
- [] Sponge *- NOTE: NOT ACTIVELY MAINTAINED*
- [] NukkitX *- NOTE: NOT ACTIVELY MAINTAINED*

### Minecraft Version:
**select one**

- [x] Minecraft 1.14.4 ***This is the ONLY actively maintained target for PlotSquared***
- [] Minecraft 1.13.2
- [] Minecraft Java Edition *other versions, please specify*:
- [] Minecraft Bedrock Edition *specify version*:

### Server build info: 

```
This server is running Paper version git-Paper-226 (MC: 1.14.4) (Implementing API version 1.14.4-R0.1-SNAPSHOT)
```

### WorldEdit/FAWE versions:

- [] FAWE *version*:
- [x] WorldEdit *version*:  WorldEdit version 7.0.1;61bc012

### Description of the problem:

If i move a plot then this error occurs and contents of barrels are not moved. the barrel on the moved plot are empty. the old moved plot get not be cleared. Iam use a older version on my server but the bug is still present in the newest dev build of p2 (Testet)

### How to replicate:

Move a plot with barrels that are filled with items. generally move plots. 

## Additional Information:
> The information here is optional for you to provide, however it may help us to more readily diagnose any compatibility and bug issues.

### Other plugins being used on the server:

Only P2, WE and MULTIVERSE

### Relevant console output, log lines, and/or screenshots:

https://hasteb.in/pezixifu.md

### Additional relevant comments/remarks:

# AFFIRMATION OF COMPLETION:

- [x] I included all information required in the sections above
- [x] I made sure there are no duplicates of this report [(Use Search)](https://github.com/IntellectualSites/PlotSquared/issues?utf8=%E2%9C%93&q=is%3Aissue)
- [x] I made sure I am using an up-to-date version of PlotSquared
- [x] I made sure the bug/error is not caused by any other plugin
- [ ] I didn't read but checked everything above.

MattBDev:
[...]

PR #2699, opened 2020-02-24
Features/v5/move swap fixes + some other important commits (comments, etc)

### LINKED ISSUE #2282: BUG /plot swap returns the original plot owners after restart
**Debug paste link:**
https://incendo.org/paste/view/0dd2befb90b54ed78b91c8b43209d5c9

I tested a few times to make sure. /plot swap works fine and swaps the plots and owners, but after a restart, it unswaps the owners of the plot. This is an old issue that was closed and now reopened.
I have not tested this on 1.13

**How to replicate:** 
Use /plot swap on different plot owners. Then restart the server and you'll see it unswaps the plot owners.

**Checklist**:
- [X] I included a `/plot debugpaste` link

Sauilitired:
## Overview

**Fixes #2282, #2603**

## Description
This PR makes sure that plot owners are swapped properly. It also makes sure to clear the 
plot on plot move (previously it just left the intact, but unclaimed). Furthermore, swapping
now updates the signs.

Some changes to the API were made, namely to make swapping and moving rely on
futures. 

The merge also includes other commits from v4, such as the `/plot comment` fix. 

## Checklist


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/Plot.java
```diff
@@ -3076,18 +3070,18 @@ public boolean move(final Plot destination, final Runnable whenDone, boolean all
         final int offsetZ = db.getZ() - ob.getZ();
         if (this.owner == null) {
             TaskManager.runTaskLater(whenDone, 1);
-            return false;
+            return CompletableFuture.completedFuture(false);
         }
-        boolean occupied = false;
+        AtomicBoolean occupied = new AtomicBoolean(false);
         Set<Plot> plots = this.getConnectedPlots();
```
dordsor21: Why are we swapping connected plots if we're not allowing moving/swapping of connected plots?

Sauilitired: I didn't really want to remove the code, although it can be removed without consequences so I guess it'd be a good idea to.

Sauilitired: But also, there's no real harm in allowing the code to swap merged plots. We just disallow it in the command as it's not something we really want to support, kind of.

Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/util/block/GenChunk.java
```diff
@@ -109,7 +110,7 @@ public boolean setBiome(int x, int z, Biome biome) {
     }
 
     @Override public boolean setBlock(int x, int y, int z, @NotNull Pattern pattern) {
-        return setBlock(x, y, z, PatternUtil.apply(pattern, x, y, z));
+        return setBlock(x, y, z, PatternUtil.apply(Preconditions.checkNotNull(pattern, "Pattern may not be null"), x, y, z));
```
MattBDev: It shouldn't be null anyways because of the annotation we have on this method.

Sauilitired: It was null when the patterns failed to parse and it was really hard to track down.

PR #2708, opened 2020-03-08
Use worldedit to regenerate chunks.

dordsor21:
Use worldedit to regenerate chunks. It's very slow, but it works.

this should be used as a quick temp fix

### DIFF
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/util/block/BukkitLocalQueue.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/util/block/BukkitLocalQueue.java
@@ -74,16 +74,12 @@ public class BukkitLocalQueue extends BasicLocalBlockQueue {
                 worldObj.regenerateChunk(x, z);
             } catch (UnsupportedOperationException e) {
                 com.sk89q.worldedit.world.World world = BukkitAdapter.adapt(worldObj);
-                EditSession editSession =
-                    WorldEdit.getInstance().getEditSessionFactory().getEditSession(world, -1);
-                CuboidRegion region =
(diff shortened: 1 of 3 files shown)


Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/util/block/BukkitLocalQueue.java
```diff
@@ -64,7 +70,22 @@ public BukkitLocalQueue(String world) {
     @Override public final void regenChunk(int x, int z) {
         World worldObj = Bukkit.getWorld(getWorld());
         if (worldObj != null) {
-            worldObj.regenerateChunk(x, z);
+            try {
+                worldObj.regenerateChunk(x, z);
+            } catch (UnsupportedOperationException e) {
+                com.sk89q.worldedit.world.World world = BukkitAdapter.adapt(worldObj);
+                EditSession editSession =
```
wizjany: you can just try-with-resources the whole editSession and let it close itself
```java
try (EditSession editSession = ...) {
   world.regen...
}
```

Sauilitired:
How *slow* is this? As in, is a good idea to have it enabled by default, or should it be opt-in?

dordsor21:
> How _slow_ is this? As in, is a good idea to have it enabled by default, or should it be opt-in?

Currently it's this or it doesn't work

Sauilitired:
> > How _slow_ is this? As in, is a good idea to have it enabled by default, or should it be opt-in?
> 
> Currently it's this or it doesn't work

I'm aware, but I think some server owners would choose not having plot clearing at all over having it freeze their entire server.

Sauilitired:
After more consideration, it's probably better for server owners to instead just limit clearing using permissions. It being slow is better than it being broken.

PR #2711, opened 2020-03-13
Encapsulation and Magic Values

SirYwell:
The changes in this PR are intended to improve the API, especially in places where attributes are exposed.
Also, some magic values are replaced with proper enums.

I didn't touch the Plot class for now, as it would be causing a lot of trouble with other current PRs besides the fact that change there should be discussed heavily first.

### DIFF
--- a/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/BukkitMain.java
+++ b/Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/BukkitMain.java
@@ -39,6 +39,7 @@ import com.github.intellectualsites.plotsquared.plot.generator.HybridUtils;
 import com.github.intellectualsites.plotsquared.plot.generator.IndependentPlotGenerator;
 import com.github.intellectualsites.plotsquared.plot.object.Plot;
 import com.github.intellectualsites.plotsquared.plot.object.PlotArea;
+import com.github.intellectualsites.plotsquared.plot.object.PlotAreaTerrainType;
 import com.github.intellectualsites.plotsquared.plot.object.PlotAreaType;
 import com.github.intellectualsites.plotsquared.plot.object.PlotId;
 import com.github.intellectualsites.plotsquared.plot.object.PlotPlayer;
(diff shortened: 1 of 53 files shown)


Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/BukkitMain.java
```diff
@@ -739,8 +741,22 @@ public final ChunkGenerator getDefaultWorldGenerator(@NotNull final String world
             SetupObject setup = new SetupObject();
             setup.plotManager = manager;
             setup.setupGenerator = worldConfig.getString("generator.init", manager);
-            setup.type = worldConfig.getInt("generator.type");
-            setup.terrain = worldConfig.getInt("generator.terrain");
+            String type = worldConfig.getString("generator.type");
+            if (type == null) {
```
Sauilitired: This exact snipper (parsing type, and terrain) exists in two different places, with identical parsing logic. It might be a good idea to move this into the enum types themselves?

SirYwell: Yeah, I'll take a look at that again

Sauilitired:
@SirYwell Want me to review before the conflicts are resolved?

SirYwell:
I'll resolve them in the next day. You can review anyways, as it's only one import conflict and one change of PlotArea.worldname to PlotArea.worldName()

Bukkit/src/main/java/com/github/intellectualsites/plotsquared/bukkit/generator/BukkitPlotGenerator.java
```diff
@@ -88,8 +88,8 @@ public BukkitPlotGenerator(final String world, final ChunkGenerator cg) {
                 Set<PlotArea> areas = PlotSquared.get().getPlotAreas(name);
                 if (!areas.isEmpty()) {
                     PlotArea area = areas.iterator().next();
-                    if (!area.MOB_SPAWNING) {
-                        if (!area.SPAWN_EGGS) {
+                    if (!area.isMobSpawning()) {
```
Sauilitired: The naming for these is a bit weird, although standard. I don't really mind it, but it could be a bit confusing.

It kind of applies to most boolean getters, so I don't know if it's worth changing. I would like to hear some opinions on this, however.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/worlds/SinglePlotArea.java
```diff
@@ -54,8 +55,8 @@ public void loadWorld(final PlotId id) {
         SetupObject setup = new SetupObject();
         setup.plotManager = "PlotSquared:single";
         setup.setupGenerator = "PlotSquared:single";
-        setup.type = TYPE;
-        setup.terrain = TERRAIN;
+        setup.type = getType();
+        setup.terrain = getTerrain();
```
dordsor21: Feels like we should update everything, so we don't mix directly using variables with encapsulated variables.

MattBDev:
I'll review this in a few hours.

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/worlds/SinglePlotArea.java
SirYwell: You mean the `SetupObject`? I agree, altough this isn't really part of the API, so we can change it at any time as we want.

PR #2715, opened 2020-03-20
Add option to have road regeneration be persistent across restarts.

dordsor21:
Let's make regenallroads persistent! It's the nicest way of doing it I could think of :)

### DIFF
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/PlotSquared.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/PlotSquared.java
@@ -22,37 +22,12 @@ import com.github.intellectualsites.plotsquared.plot.generator.HybridUtils;
 import com.github.intellectualsites.plotsquared.plot.generator.IndependentPlotGenerator;
 import com.github.intellectualsites.plotsquared.plot.listener.WESubscriber;
 import com.github.intellectualsites.plotsquared.plot.logger.ILogger;
-import com.github.intellectualsites.plotsquared.plot.object.BlockBucket;
-import com.github.intellectualsites.plotsquared.plot.object.ConsolePlayer;
-import com.github.intellectualsites.plotsquared.plot.object.Location;
-import com.github.intellectualsites.plotsquared.plot.object.Plot;
-import com.github.intellectualsites.plotsquared.plot.object.PlotArea;
-import com.github.intellectualsites.plotsquared.plot.object.PlotCluster;
-import com.github.intellectualsites.plotsquared.plot.object.PlotFilter;
-import com.github.intellectualsites.plotsquared.plot.object.PlotId;
-import com.github.intellectualsites.plotsquared.plot.object.PlotManager;
-import com.github.intellectualsites.plotsquared.plot.object.PlotPlayer;
-import com.github.intellectualsites.plotsquared.plot.object.StringWrapper;
+import com.github.intellectualsites.plotsquared.plot.object.*;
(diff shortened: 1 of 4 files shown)


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/PlotSquared.java
```diff
@@ -1525,6 +1558,44 @@ public void copyFile(String file, String folder) {
...
         try {
             eventDispatcher.unregisterAll();
+            if (HybridUtils.UPDATE && Settings.Enabled_Components.PERSISTENT_ROAD_REGEN && (
+                !HybridUtils.regions.isEmpty() || !HybridUtils.chunks.isEmpty())) {
+                PlotSquared.log(Captions.PREFIX
+                    + "Road regeneration incomplete. Saving incomplete regions to disk.");
+                PlotSquared.debug("   Regions: " + HybridUtils.regions.size());
+                PlotSquared.debug("   Chunks: " + HybridUtils.chunks.size());
+                try {
+                    ArrayList<Integer[]> regions = new ArrayList<>();
```
SirYwell: I don't know how much data is serialized here, but using `ArrayList<int[]>` instead of `ArrayList<Integer[]>` will likely decrease the file size significantly

dordsor21: ab9e84d4718b8c567b59b69d7d09b673c99e1689

Core/src/main/java/com/github/intellectualsites/plotsquared/plot/PlotSquared.java
```diff
@@ -1525,6 +1558,44 @@ public void copyFile(String file, String folder) {
...
+                            + "persistent_regen_data file already exists and could not be deleted.");
+                    }
+                    if (!file.createNewFile()) {
+                        PlotSquared
+                            .log(Captions.PREFIX + "Error create persistent_regen_data file.");
+                    }
+                    ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(file));
+                    oos.writeObject(list);
+                    oos.flush();
+                    oos.close();
```
SirYwell: Don't know why I didn't see that before, but `oos.close()` should be called in a `finally` block as it wouldn't close the resource on an exception otherwise. 

As an alternative way, you could use `Files.newOutputStream(Path.of(this.IMP.getDirectory(), "persistent_regen_data_" + HybridUtils.area.id + "_" + HybridUtils.area.worldname))` in a try-with-resource statement and therefore also minimize the scope of the try-catch block.

The same applies for reading obviously.

dordsor21: Oof, idk why I put the closing inside the try in the first place...

dordsor21:
> HybridUtils.UPDATE is never set to false.

https://github.com/IntellectualSites/PlotSquared/blob/v4/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridUtils.java#L380

MattBDev:
> > HybridUtils.UPDATE is never set to false.
> 
> https://github.com/IntellectualSites/PlotSquared/blob/v4/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/generator/HybridUtils.java#L380
> 
> >  GitHub**[IntellectualSites/PlotSquared](https://github.com/IntellectualSites/PlotSquared)**PlotSquared - Reinventing the plotworld. Contribute to IntellectualSites/PlotSquared development by creating an account on GitHub.
Ctrl+F has failed me.

PR #2754, opened 2020-04-10
Remove all direct access to Plot.owner

Sauilitired:
Remove all direct access to Plot.owner

New methods were added for access to the absolute owner of a plot, and the documentation of the owner getters to clarify the purpose of the methods.

### DIFF
--- a/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/Plot.java
+++ b/Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/Plot.java
@@ -411,16 +411,13 @@ public class Plot {
     }
 
     /**
-     * plot owner
+     * Get the plot owner of this particular sub-plot.
      * (Merged plots can have multiple owners)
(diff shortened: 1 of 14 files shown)


Core/src/main/java/com/github/intellectualsites/plotsquared/plot/object/Plot.java
```diff
@@ -113,55 +107,50 @@
      * @deprecated magical
      */
     @Deprecated public int temp;
-
+    /**
+     * plot owner
+     * (Merged plots can have multiple owners)
+     * Direct access is Deprecated: use getOwners()
+     *
+     * @deprecated
```
SirYwell, line 115: Not deprecated

Sauilitired:
> There's a lot of `Plot#getOwnerAbs() == null` checks, it's probably better to replace it with `Plot#hasOwner()`

Oh yeah, definitely.

PR #2757, opened 2020-04-10
Package Restructure

MattBDev:
This PR changes the package structure of the project dramatically. The review of these changes should be taken very seriously and heavily critiqued. Once these changes are made it is unlikely that packages will be changing again.


Sauilitired:
Also, good call removing the Nukkit and Sponge modules. If we decide to revive development of those modules, then we can just rip them out of v4.

PR #2776, opened 2020-04-18
If all players are denied from a plot, this will allow administrators…

Cheezybob99:
… with permission to see chatspy, talk in plots they have been denied from.

## Overview
This will allow staff members/administrators to speak in the plot chat of a plot they have been denied. This will be useful for private conversations with plot owners when all players are denied, or, allowing server staff to continue to speak in the plot, even if they have been denied by players trying to evade them.

**Fixes N/A**

## Description
When a staff member or administrator tries to speak in a plot where everyone is denied, they cannot, this will resolve that making life easier for server staff. This adds a further check to PlayerEvents, checking that the user does not have permission to see chatspy before canceling the event.

## Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEvents.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEvents.java
@@ -946,7 +946,7 @@ public class PlayerEvents extends PlotListener implements Listener {
             || area.isForcingPlotChat())) {
             return;
         }
-        if (plot.isDenied(plotPlayer.getUUID())) {
+        if (plot.isDenied(plotPlayer.getUUID()) && !plotPlayer.hasPermission("plots.admin.command.chat")) {
             return;
         }
         event.setCancelled(true);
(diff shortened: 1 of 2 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEvents.java
```diff
@@ -946,7 +946,7 @@ public void playerMove(PlayerMoveEvent event) {
             || area.isForcingPlotChat())) {
             return;
         }
-        if (plot.isDenied(plotPlayer.getUUID())) {
+        if (plot.isDenied(plotPlayer.getUUID()) && !plotPlayer.hasPermission("plots.admin.command.chat")) {
```
Sauilitired, line 949: The permission node should be a caption in `com.plotsquared.core.configuration.Captions`, then `com.plotsquared.core.util.Permissions.hasPermission(player, caption)` can be used.

Sauilitired:
I think there was a slight misunderstanding. That permission is the permission node for the chat command, which means everyone with permission to use the plot chat command now have the admin override.

Instead, a new override permission node must be created in the captions file (similar to the other ones). It's a bit tedious, I'm aware.

PR #2807, opened 2020-05-04
Lazily initialize block tags

SirYwell:
This PR fixes block categories being loaded later by WE. It disables the validation process when loading the flags from the database.

Things that need to be considered:
- Currently, it simply throws an NPE if the category does not exist when initialiting lazily
- The hashCode method generated by lombok for the BlockTypeWrapper most likely fails to work correctly. Probably we set the string even if it was created using a BlockCategory and exclude the BlockCategory from the hashCode

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeListFlag.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeListFlag.java
@@ -42,6 +42,8 @@ import java.util.stream.Collectors;
 public abstract class BlockTypeListFlag<F extends ListFlag<BlockTypeWrapper, F>>
     extends ListFlag<BlockTypeWrapper, F> {
 
+    public static boolean skipCategoryVerification = false;
+
     protected BlockTypeListFlag(List<BlockTypeWrapper> blockTypeList, Caption description) {
         super(blockTypeList, Captions.FLAG_CATEGORY_BLOCK_LIST, description);
     }
@@ -57,15 +59,7 @@ public abstract class BlockTypeListFlag<F extends ListFlag<BlockTypeWrapper, F>>
             final BlockState blockState = BlockUtil.get(blockString);
             if (blockState == null) {
                 // If it's not a block state, we assume it's a block category
-                final BlockCategory blockCategory;
-                if (!blockString.startsWith("#")
-                    || (blockCategory = BlockCategory.REGISTRY.get(blockString.substring(1)))
-                    == null) {
-                    throw new FlagParseException(this, blockString,
-                        Captions.FLAG_ERROR_INVALID_BLOCK);
-                } else {
-                    blockTypeWrapper = BlockTypeWrapper.get(blockCategory);
-                }
+                blockTypeWrapper = getCategory(blockString);
             } else {
                 blockTypeWrapper = BlockTypeWrapper.get(blockState.getBlockType());
             }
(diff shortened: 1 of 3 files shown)


Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeListFlag.java
```diff
@@ -91,4 +85,24 @@ protected BlockTypeListFlag(List<BlockTypeWrapper> blockTypeList, Caption descri
         return tabCompletions;
     }
 
+    private BlockTypeWrapper getCategory(final String blockString) throws FlagParseException {
+        if (!blockString.startsWith("#")) {
```
dordsor21, line 89: Why did we need to move this inside getCategory? Obviously still works but I think it makes more send in the parent method.

Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeWrapper.java
```diff
@@ -86,15 +95,26 @@ public boolean accepts(final BlockType blockType) {
         }
     }
 
+    public BlockCategory getBlockCategory() {
+        if (this.blockCategory == null && this.blockCategoryId != null) { // only if name is available
+            this.blockCategory = Preconditions.checkNotNull(BlockCategory.REGISTRY.get(this.blockCategoryId));
```
dordsor21, line 100: Will this cause a soft failure wherever getBlockCategory is called? It might be best to add an exception to the method so we're not destroying something we shouldn't in parent methods

Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeListFlag.java
SirYwell: I thought it would be more readable

MattBDev:
I'll review this by Friday.

Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeWrapper.java
```diff
@@ -86,15 +97,45 @@ public boolean accepts(final BlockType blockType) {
         }
     }
 
+    public BlockCategory getBlockCategory() {
+        if (this.blockCategory == null && this.blockCategoryId != null) { // only if name is available
+            this.blockCategory = BlockCategory.REGISTRY.get(this.blockCategoryId);
+            if (this.blockCategory == null) {
```
Sauilitired, line 103: if (category is null and registry is not empty), this means that in the case that this method is called before the registry has been loaded, it'll have to refresh next time

Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeWrapper.java
```diff
@@ -64,8 +75,8 @@ private BlockTypeWrapper(@NotNull final BlockCategory blockCategory) {
             } else {
                 return key;
             }
-        } else if (this.blockCategory != null) {
-            final String key = this.blockCategory.toString();
+        } else if (this.getBlockCategory() != null) { // calling the method will initialize it lazily
+            final String key = this.getBlockCategory().toString();
             if (key.startsWith("minecraft:")) {
                 return '#' + key.substring(10);
             } else {
```
Sauilitired, line 82: return the category ID if both the state and category are null

Core/src/main/java/com/plotsquared/core/plot/flag/types/BlockTypeWrapper.java
```diff
@@ -44,16 +46,25 @@
 public class BlockTypeWrapper {
```
Sauilitired, line 46: Remove equalsAndHashCode and implement it using equals and hashcode builders from guava/apache (intellij has a generation thingy for this :cow:)

PR #2824, opened 2020-05-13
Split ChunkManager to Chunk and Region Managers

dordsor21:
Will require a push to the mvn repo for FAWE to work correctly with it.


dordsor21:
I'm not sure that it's wrong though^^ chop-down-if-long is correct, so when below the max line character length it doesn't wrap

PR #2834, opened 2020-05-16
Features/v5/components

Sauilitired:
This has been tested on intellectualsites.com

(This is built on top of Features/v5/world-manager rather than v5, mostly because I fucked up, but also so I can bully you into merging it :cow:)

### DIFF
--- a/Bukkit/build.gradle
+++ b/Bukkit/build.gradle
@@ -30,7 +30,7 @@ dependencies {
         exclude(module: "bukkit")
     }
     implementation("me.clip:placeholderapi:2.10.4")
-    compile("se.hyperver.hyperverse:Core:0.6.0-SNAPSHOT")
+    compile("se.hyperver.hyperverse:Core:0.6.0-SNAPSHOT"){ transitive = false }
 }
 
 sourceCompatibility = 1.8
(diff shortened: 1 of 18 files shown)


dordsor21:
Since it includes the generation changes stuff, perhaps the other PR should be closed?

Sauilitired:
Fair, I've copied over the description from that one:

This should make world management a fair bit cleaner

This has been tested with Hyperverse and Bukkit, with both normal plot worlds and augmented worlds.

A side effect is that PlotSquared will now register itself in bukkit.yml for every world that loads with a normal square plot generator on setup, as the world creation methods are loaded that way.

The idea is to then add more plugin specific features to these world managers, where needed.

dordsor21:
I mean that should really be done regardless. It ensures that stray generators can't hijack the world, and if the main world is changed, there's no issue for certain.

Bukkit/build.gradle
```diff
@@ -29,6 +30,7 @@ dependencies {
         exclude(module: "bukkit")
     }
     implementation("me.clip:placeholderapi:2.10.4")
+    compile("se.hyperver.hyperverse:Core:0.6.0-SNAPSHOT")
```
N0tMyFaultOG, line 33: I'd add `{ transitive = false }` because we don't have stuff like taskchain, acf or hocon on our maven and it's not needed.

PR #2843, opened 2020-05-19
Add option for roads to respect a plot area's flags

dordsor21:
Since we've got a global flag container that works properly, it makes sense to have the option for these flags to be applicable to the entire plot area in most cases.

Fixes https://github.com/IntellectualSites/PlotSquaredSuggestions/issues/120


Sauilitired:
If we're going to be accessing flags from areas, can you copy the flag getter from the plot object so we don't have to use the value getter in the flag? Just a QoL thing but it'd make the code a little prettier

dordsor21:
That's fair. I basically just copy-pasted the same section of code and changed the flag lol

Sauilitired:
We could create a new flag container for the roads and let it inherit from the area container, if we want to be able to define road specific flags

RedstoneFuture:
https://github.com/IntellectualSites/PlotSquaredSuggestions/issues/120

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
```diff
@@ -370,6 +373,40 @@ public void loadDefaultConfiguration(ConfigurationSection config) {
...
+            String prefix = " ";
+            for (final PlotFlag<?, ?> flag : roadFlagCollection) {
+                Object value = flag.toString();
+                roadFlagBuilder.append(prefix).append(CaptionUtility
+                    .format(null, Captions.PLOT_FLAG_LIST.getTranslated(), flag.getName(),
+                        CaptionUtility.formatRaw(null, value.toString(), "")));
+                prefix = ", ";
+            }
+        }
+        PlotSquared.log(Captions.PREFIX + "&3 - road flags: &7" + roadFlagBuilder.toString());
```
N0tMyFaultOG, line 408: This could be changed to debug.

dordsor21: It's log for the plot flags as well, so figured I'd keep the same.

N0tMyFaultOG: Ah ok, makes sense. I shifted a few things like that to debug in the past so it doesn't clog up things unnecessary when you have lots of worlds.
I've seen people having like 50+ worlds and that can get rather spammy then.

PR #2846, opened 2020-05-21
Features/v5/worlds

Sauilitired:
Rework the way we store plot areas by using R-trees in clustered plot worlds.

This has not yet been tested. 

Note that this does not break the API at all. The aim was to just replace the implementation.


Sauilitired:
I’ve tested this and it appears to work like expected.

PR #2847, opened 2020-05-22
Features/v5/single area

Sauilitired:
Dependent on #2846

This alters the world generator behavior to allow for (optional) per-area schematics as opposed to just having per-world schematics, it also adds `/plot area single <id>` which uses a WorldEdit selection to a) export a schematic, b) setup a region.

There are some minor problems where it currently has a dumb offset, probably because I forgot to set the road width.

Another import issue that should be resolved (in a separate branch) is that PlotSquared delegates all partial plot clearing to WorldEdit, even when it's using PlotSquared terrain. This is bad.


Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -2014,6 +2017,23 @@ public int getPlotCount() {
...
     }
 
+    /**
+     * Check if the chunk uses vanilla/non-PlotSquared generation
+     *
+     * @param world            World name
+     * @param chunkCoordinates Chunk coordinates
+     * @return True if the chunk uses non-standard generation, false if not
+     */
+    public boolean isNonStandardGeneration(@NotNull final String world,
```
SirYwell, line 2027: does non-standard mean non-plotsquared?

Sauilitired: No, it means anything that isn't a normal (standard) plot world

PR #2856, opened 2020-06-05
Overhauled setup command

SirYwell:
This PR is a rewrite of the setup command. The main target is to improve readability of the single setup steps but keeping and extending functionality and extendability of the previous implementation.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/configuration/Captions.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/Captions.java
@@ -319,6 +319,7 @@ public enum Captions implements Caption {
     REMOVED_GRANTED_PLOT("$2You used %s0 plot grant(s), you've got $1%s1 $2left", "Economy"),
     //</editor-fold>
     //<editor-fold desc="Setup">
+    SETUP_NOT_STARTED("$7No setup started. Use $2/plot setup $7to start a setup process.", "Setup"),
     SETUP_INIT("$1Usage: $2/plot setup <value>", "Setup"),
     SETUP_STEP("$3[$1Step %s0$3] $1%s1 $2- $1Expecting: $2%s2 $1Default: $2%s3", "Setup"),
     SETUP_INVALID_ARG("$2%s0 is not a valid argument for step %s1. To cancel setup use: $1/plot setup cancel", "Setup"),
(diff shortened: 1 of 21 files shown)


Core/src/main/java/com/plotsquared/core/configuration/Captions.java
```diff
@@ -319,6 +319,7 @@
     REMOVED_GRANTED_PLOT("$2You used %s0 plot grant(s), you've got $1%s1 $2left", "Economy"),
     //</editor-fold>
     //<editor-fold desc="Setup">
+    SETUP_NOT_STARTED("$7No setup started. Use $2/plot setup $7to start a setup process.", "Setup"),
```
N0tMyFaultOG, line 322: Let's use $1 here, $7 is a bit too aggressive.

Core/src/main/java/com/plotsquared/core/setup/CommonSetupSteps.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.stream.Collectors;
+
+import static com.plotsquared.core.util.MainUtil.sendMessage;
+
+public enum CommonSetupSteps implements SetupStep {
+    CHOOSE_GENERATOR(Captions.SETUP_INIT) {
+
+        @Override public SetupStep handleInput(PlotPlayer<?> plotPlayer, PlotAreaBuilder builder, String arg) {
+            if (!SetupUtils.generators.containsKey(arg)) {
+                String prefix = "\n&8 - &7";
```
N0tMyFaultOG, line 59: Is there any specific reason why we don't include that in the Captions? See https://github.com/IntellectualSites/PlotSquared/blob/v5/Core/src/main/java/com/plotsquared/core/configuration/Captions.java#L336 for reference

Core/src/main/java/com/plotsquared/core/setup/CommonSetupSteps.java
```diff
@@ -0,0 +1,288 @@
...
+        return Arrays.stream(type.getEnumConstants()).map(e -> e.toString().toLowerCase()).collect(Collectors.toList());
+    }
+
+    private static SettingsNodesWrapper wrap(String plotManager) {
+        return new SettingsNodesWrapper(SetupUtils.generators.get(plotManager).getPlotGenerator()
+                .getNewPlotArea("CheckingPlotSquaredGenerator", null, null, null)
+                .getSettingNodes(), CHOOSE_WORLD_NAME);
+    }
+
+    private static boolean isValidWorldName(String s) {
```
N0tMyFaultOG, line 283: Is that still necessary? World names can be uppercase as well. The commit forcing world names to be lowercase refers to 1.13 and has been reverted years ago.

Core/src/main/java/com/plotsquared/core/setup/CommonSetupSteps.java
SirYwell: That's a weird special case, as the generators are sent  different from the other messages here. https://github.com/IntellectualSites/PlotSquared/blob/21b5b10358d45b837085a7678bc0e5424f21bae0/Core/src/main/java/com/plotsquared/core/command/Setup.java#L56 is the same weird part, but I had no good idea for now how to keep it dynamic. I'm open for any suggestions for that

Core/src/main/java/com/plotsquared/core/setup/CommonSetupSteps.java
SirYwell: We can change that to match `[a-zA-Z0-9]` instead, yeah (or are there any other restrictions/possibilities?)

N0tMyFaultOG: I wanted to get a rid of it, since it's no longer necessary cb/b wise. Or do we rely on lowercased world names?

Sauilitired: There's no reason to allow uppercase world names imo

MattBDev: We could add a configuration to allow world names that are not lowercase. I have been strongly opposed to allowing non lowercase due to legacy setups possibly being messed up. I also want to reduce the number of calls to `toLowerCase()` and `equalsIgnoreCase()`

PR #2859, opened 2020-06-22
Add placeholder to get the value of a plot flag

### LINKED ISSUE #106: plot denied players can still enter plots
as the title says, players who are on the denied list are still able to enter and annoy players who have denied them

EpiCanard:
Hello :)

## Overview
For my server, I need to get the flags value of a plot (to build a gui with DeluxeMenus).
For this purpose, I have developed the feature I need. I first add this feature in Placeholder-Expansion but you told me it is legacy, so I come here to add it in PlotSquared V5. 

**Fixes #106** ([issue](https://github.com/IntellectualSites/PlotSquaredSuggestions/issues/106))

## Description
For this purpose, this PullRequest add the following feature :
- Add a new placeholder `%plotsquared_currentplot_flag_<flagName>%` to get the value of a flag when the player is on a plot

Behavior :
- If the flag is set on the current plot, it returns the flag value serialized in string
- If the flag is not set on the current plot, it returns the default value
- If the flag doesn't exist, it returns an empty string

## Checklist

Thank you in advance for your feedback
Best regards,
EpiCanard

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
@@ -29,6 +29,8 @@ import com.plotsquared.core.PlotSquared;
 import com.plotsquared.core.configuration.Settings;
 import com.plotsquared.core.player.PlotPlayer;
 import com.plotsquared.core.plot.Plot;
+import com.plotsquared.core.plot.flag.GlobalFlagContainer;
+import com.plotsquared.core.plot.flag.PlotFlag;
 import me.clip.placeholderapi.PlaceholderAPIPlugin;
 import me.clip.placeholderapi.expansion.PlaceholderExpansion;
 import org.bukkit.Bukkit;
@@ -36,8 +38,12 @@ import org.bukkit.entity.Player;
 
 import java.util.Set;
 import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class Placeholders extends PlaceholderExpansion {
+
+    private Pattern flagPattern = Pattern.compile("currentplot_flag_(.+)");
 
     public Placeholders() {
     }
@@ -59,7 +65,7 @@ public class Placeholders extends PlaceholderExpansion {
     }
 
     @Override public String getVersion() {
-        return "2.4";
+        return "2.5";
     }
 
     @Override public String onPlaceholderRequest(Player p, String identifier) {
@@ -124,7 +130,7 @@ public class Placeholders extends PlaceholderExpansion {
                     return "";
                 }
 
-                String name = PlotSquared.get().getImpromptuUUIDPipeline() .getSingle(uid,
+                String name = PlotSquared.get().getImpromptuUUIDPipeline().getSingle(uid,
                     Settings.UUID.BLOCKING_TIMEOUT);
 
                 if (name != null) {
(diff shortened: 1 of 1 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
```diff
@@ -181,6 +187,31 @@ public Placeholders() {
...
+     * If the flag exists but it is not set on current plot it return the default value.
+     *
+     * @param plot     Current plot where the player is
+     * @param flagName Name of flag to get from current plot
+     * @return The value of flag serialized in string
+     */
+    private String getFlag(final Plot plot, final String flagName) {
+        final PlotFlag<?, ?> flag = GlobalFlagContainer.getInstance().getFlagFromString(flagName);
+        if (flag != null) {
+            return plot.getFlags().stream()
```
SirYwell, line 209: That should give the same result as `plot.getFlag(flag).toString()`

Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
```diff
@@ -181,6 +187,31 @@ public Placeholders() {
             default:
                 break;
         }
+        final Matcher matcher = this.flagPattern.matcher(identifier);
+        if (matcher.find()) {
```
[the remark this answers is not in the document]
EpiCanard: I can add a `^` at the start ofthe line to be sure it matches from the start.
Anyway if you prefer `startsWith` and `split`, it's no a problem I can change that.

EpiCanard:
> I like the idea but it's might worth to differentiate between actual flags applied to a plot and inherited plot flags. While the former might return null values, the latter are guaranteed to return non-null values for existing flags.

I can create two placeholders, one that return effective flags and the other only the plot flags set.
I found it interesting to know which value is actually used even if it is not set in the current plot.

Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
EpiCanard: Change applied in last commit

Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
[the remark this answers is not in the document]
EpiCanard: Change applied in last commit

EpiCanard:
As wrote above, I created two placeholders :
- `%plotsquared_currentplot_localflag_<flagName>%` that return only the value of the flag set on current plot
- `%plotsquared_currentplot_flag_<flagName>%` that return the value of the flag set on current plot or inherited flag value if it is not set on current plot

Bukkit/src/main/java/com/plotsquared/bukkit/placeholder/Placeholders.java
```diff
@@ -181,6 +183,40 @@ public Placeholders() {
             default:
                 break;
         }
+        if (identifier.startsWith("currentplot_localflag_")) {
+            final String[] splitValues = identifier.split("currentplot_localflag_");
```
SirYwell, line 187: Sorry, I wanted to mention that yesterday already but things came but so I forgot about it. In that case, a simple `identifier.substring("currentplot_localflag_".length())` would be fine too (it would return an empty String `""` if no flag name is given)

EpiCanard: I made some benchmark (regex vs split vs substring) on my side and I saw much better performance with substring, so I took the liberty to also change other flags which use split.

PR #2862, opened 2020-06-27
Update to 1.16.1

MattBDev:
The 1.16.1 update was very insignificant for PlotSquared. Only one change was made and I noticed some missing blocks in an if statement.
I have not tested PlotSquared on 1.16.1 yet even with these changes but I feel it is safe to say that it is working.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitMain.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitMain.java
@@ -627,46 +627,46 @@ public final class BukkitMain extends JavaPlugin implements Listener, IPlotMain<
                 Iterator<Entity> iterator = entities.iterator();
                 while (iterator.hasNext()) {
                     Entity entity = iterator.next();
-                    switch (entity.getType()) {
-                        case EGG:
-                        case FISHING_HOOK:
-                        case ENDER_SIGNAL:
-                        case AREA_EFFECT_CLOUD:
-                        case EXPERIENCE_ORB:
-                        case LEASH_HITCH:
-                        case FIREWORK:
-                        case LIGHTNING:
-                        case WITHER_SKULL:
-                        case UNKNOWN:
-                        case PLAYER:
+                    switch (entity.getType().toString()) {
+                        case "EGG":
+                        case "FISHING_HOOK":
+                        case "ENDER_SIGNAL":
+                        case "AREA_EFFECT_CLOUD":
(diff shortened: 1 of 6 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/BukkitMain.java
```diff
@@ -790,7 +791,6 @@ private void startUuidCatching(@NotNull final SQLiteUUIDService sqLiteUUIDServic
                         case MUSHROOM_COW:
                         case OCELOT:
                         case PIG:
```
Sauilitired, line 793: This needs to be solved differently, possibly by using the names. We need this version to be backward compatible to 1.13. This should also include the new mobs, such as zombified piglins, piglins and hoglins.

MattBDev: Resolved in the new commit.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEvents.java
```diff
@@ -1389,6 +1389,11 @@ public void onFade(BlockFadeEvent event) {
             case BUBBLE_CORAL_FAN:
             case FIRE_CORAL_FAN:
             case HORN_CORAL_FAN:
+            case BRAIN_CORAL_WALL_FAN:
+            case BUBBLE_CORAL_WALL_FAN:
+            case FIRE_CORAL_WALL_FAN:
+            case HORN_CORAL_WALL_FAN:
+            case TUBE_CORAL_WALL_FAN:
```
dordsor21, line 1396: If we're worried about entity types causing issues, might this also be something we're worried about?

N0tMyFaultOG: What do you mean? This is not related to 1.16.

Sauilitired: New ones shouldn't cause issues. The issue is when removing enums, as older server versions won't get the same level of protection

dordsor21: That makes sense

MattBDev: These were added to Minecraft awhile ago so it shouldn't cause any compatibility problems.

PR #2863, opened 2020-06-27
Get biome data from correct schematic for intersections (Fixes PS-50)

MegaNarwhal:
## Overview

**Fixes https://issues.intellectualsites.com/issue/PS-50**

## Description
Copy-paste from issue tracker:
> When attempting to use custom road schematics, I ran into an issue where the roads generated with /p createroadschematic were all ocean biome. I was able to fix this for the road by creating the schematic using WorldEdit and including the proper biome in the schematic using //copy -b, however when I followed the same steps for the intersection, the intersections still generate with the biome as ocean. As you can see in the attached screenshot, the roads are properly showing as plains compared to the dark forest biome set for the plots, however the intersections are generating as ocean.

## Checklist


dordsor21:
Nice catch. Thank you

PR #2864, opened 2020-06-29
Let's go rainbow

N0tMyFaultOG:
TODO:

- [x] Generate default messages file
- [ ] Move to MiniMessage for Captions
- [ ] Yeet Captions and get permissions their own class
- [ ] Yeet style.yml
- [ ] Yeet default translations setup
- [ ] Add a sane config option to use a translation file rather renaming the file
- [ ] Add support for multiple locales and override it with the player's default ?
- [ ] Possibly make parts of p2 translate that weren't translatable before (or have been simply forgotten)
- [ ] Make menus like `/plot help` clickable 
- [ ] Setup a crowding project
- [x] Don't test this because it works


RedstoneFuture:
Other idea: you use [Minedown library](https://github.com/Phoenix616/MineDown#advanced-syntax) of Phoenix616.

Sauilitired:
> Other idea: you use [Minedown library](https://github.com/Phoenix616/MineDown#advanced-syntax) of Phoenix616.

MiniMessage has become the de facto standard of the community now, and it’s targeting text/adventure, which we already rely on. I think it’s the best fit for what we need.

PR #2869, opened 2020-07-06
Split visit command into home and visit

SirYwell:
This PR introduces a new command `home` to make the visit command less ambiguous.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/HomeCommand.java
+++ b/Core/src/main/java/com/plotsquared/core/command/HomeCommand.java
@@ -1,0 +1,169 @@
+/*
+ *       _____  _       _    _____                                _
+ *      |  __ \| |     | |  / ____|                              | |
+ *      | |__) | | ___ | |_| (___   __ _ _   _  __ _ _ __ ___  __| |
+ *      |  ___/| |/ _ \| __|\___ \ / _` | | | |/ _` | '__/ _ \/ _` |
+ *      | |    | | (_) | |_ ____) | (_| | |_| | (_| | | |  __/ (_| |
+ *      |_|    |_|\___/ \__|_____/ \__, |\__,_|\__,_|_|  \___|\__,_|
+ *                                    | |
+ *                                    |_|
+ *            PlotSquared plot management system for Minecraft
+ *                  Copyright (C) 2020 IntellectualSites
+ *
+ *     This program is free software: you can redistribute it and/or modify
+ *     it under the terms of the GNU General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ *
+ *     This program is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.plotsquared.core.command;
+
+import com.plotsquared.core.PlotSquared;
+import com.plotsquared.core.configuration.Captions;
+import com.plotsquared.core.events.TeleportCause;
+import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.Plot;
+import com.plotsquared.core.plot.PlotArea;
+import com.plotsquared.core.plot.PlotId;
+import com.plotsquared.core.util.MainUtil;
+import com.plotsquared.core.util.MathMan;
+import com.plotsquared.core.util.query.PlotQuery;
+import com.plotsquared.core.util.query.SortingStrategy;
+import com.plotsquared.core.util.task.RunnableVal2;
+import com.plotsquared.core.util.task.RunnableVal3;
+import org.jetbrains.annotations.NotNull;
+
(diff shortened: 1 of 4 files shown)


RedstoneFuture:
```
usage = "/plot visit [<player>|<alias>|<world>|<id>] [#]",
usage = "/plot home [<page>|<alias>|<area;x;y>|<area> <x;y>]",
```
Good, with this split we can support number-playernames like [2234](https://de.namemc.com/profile/2234.1), I think.


Can two different people have the same alias for 2 different plos? Or is "/plot home <alias>" the same as "/plot visit alias"?

How com with "/plot visit world 1"? To the first plot of "world"?

What do you think about:

```
usage = "/plot visit [<player> <#>|<alias>|<world or area> <x;y>]",
usage = "/plot home [<#>]",
```

SirYwell:
> Good, with this split we can support number-playernames
That was the main reason for this change.

Aliases are unique per plot area, so `/plot gome <alias>` and `/plot visit <alias>` will target the same plot, although home only allows you to teleport to plots you're the owner of, whereas you can teleport to other's plot using visit (if the required permissions are set).

Visit is meant to visit plots of other players, using either their name, their name and a number, a plot id or an alias.

Core/src/main/java/com/plotsquared/core/command/Visit.java
```diff
@@ -54,8 +54,8 @@
 @CommandDeclaration(command = "visit",
     permission = "plots.visit",
     description = "Visit someones plot",
-    usage = "/plot visit [<player>|<alias>|<world>|<id>] [#]",
-    aliases = {"v", "tp", "teleport", "goto", "home", "h", "warp"},
+    usage = "/plot visit <<player>|<alias>|<world>> [#]",
```
Sauilitired, line 57: This is no longer correct, is it? Should probably be `<player>|<alias>|<plot>`

Core/src/main/java/com/plotsquared/core/command/Visit.java
```diff
@@ -155,7 +155,7 @@ private void visit(@NotNull final PlotPlayer player, @NotNull final PlotQuery qu
         switch (args.length) {
             // /p v [...] [...] <page>
             case 3:
-                if (!MathMan.isInteger(args[1])) {
+                if (!MathMan.isInteger(args[2])) {
```
Sauilitired, line 158: Is this then `/plot visit <user> <area> <number>`?

Core/src/main/java/com/plotsquared/core/command/Visit.java
```diff
@@ -214,61 +203,31 @@ private void visit(@NotNull final PlotPlayer player, @NotNull final PlotQuery qu
...
+                Captions.COMMAND_SYNTAX.send(player, getUsage());
+                return CompletableFuture.completedFuture(false);
             default:
         }
 
         return CompletableFuture.completedFuture(true);
     }
 
-    public Collection<Command> tab(PlotPlayer player, String[] args, boolean space) {
+    @Override public Collection<Command> tab(PlotPlayer player, String[] args, boolean space) {
```
Sauilitired, line 230: in `case 0` it still completes numbers, this should only be done for `/plot home`. `case 1` is also wrong as it assumes a number for the first argument, which is only the case for home, so the entire tab completion here is wonky

Core/src/main/java/com/plotsquared/core/command/HomeCommand.java
```diff
@@ -0,0 +1,169 @@
...
+                query.withPlot(plot);
+                break;
+            case 0:
+                query.withSortingStrategy(SortingStrategy.SORT_BY_CREATION);
+                break;
+        }
+        home(player, query, page, confirm, whenDone);
+        return CompletableFuture.completedFuture(true);
+    }
+
```
Sauilitired, line 168: missing completions?

Core/src/main/java/com/plotsquared/core/command/Visit.java
SirYwell: Yes, seems to be the only remaining part with 3 arguments. So the usage should be `/plot visit <player>|<alias>|<plot> [area] [#]` I guess?

Sauilitired: it's a bit messier than that, `/plot visit <player>|<alias>|<plot> [area]|[#] [#]` lol

RedstoneFuture:
`usage = "/plot home [<page>|<alias>|<area;x;y>|<area> <x;y>|<area> <page>]",` Ok.

`usage = "/plot visit <player>|<alias>|<plot> [area]|[#] [#]",`

Do you mean:

`usage = "/plot visit <<player>|<alias>|<area;x;y>|<area> <x;y>|<area>> [page]",`

/p v testplayer
/p v testplayer 2
/p v testalias
/p v world;1;1
/p v world 1;2
/p v world 2

RedstoneFuture:
**Tests | worked commands:**

/p h
/p h <#>
/p h <alias> **(Currently not with merged plots!)**
/p h <X>;<Y>
/p h <area/world> <X>;<Y>
/p h <area/world>;<X>;<Y>
/p h <area/world> <#>


/p v <player>
/p v <player> <area/world>
/p v <player> <area/world> <#>
/p v <alias>
/p v <X>;<Y>
/p v <area/world>;<X>;<Y>


**`plots.visit.other` still needed. Maybe we remove this permission check.**
/p v <area/world> <X>;<Y> **not working (consistence ? Or unimpossible? )**

PR #2870, opened 2020-07-07
Feature/v6/platform

Sauilitired:
The purpose of this branch is to clean up stuff, and move around some platform specific code while trying to make PlotSquared.java less of a mess.

It also makes locations immutable (as they should have been)!

Now with 100% more Guice!


Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitUtil.java
```diff
@@ -271,23 +271,23 @@ public static BukkitPlayer getPlayer(@NonNull final Player player) {
...
         }
-        final PlayerManager<?, ?> playerManager = PlotSquared.imp().getPlayerManager();
+        final PlayerManager<?, ?> playerManager = PlotSquared.platform().getPlayerManager();
         return ((BukkitPlayerManager) playerManager).getPlayer(player);
     }
 
     public static Location getLocation(@NonNull final org.bukkit.Location location) {
-        return new Location(location.getWorld().getName(), MathMan.roundInt(location.getX()),
-            MathMan.roundInt(location.getY()), MathMan.roundInt(location.getZ()));
+        return Location.at(com.plotsquared.bukkit.util.BukkitWorld.of(location.getWorld()),
```
dordsor21, line 279: I'd be tempted to import the P2 BukkitWorld and directly reference the WorldEdit one, as the P2 one is used more.

Core/src/main/java/com/plotsquared/core/command/DebugExec.java
```diff
@@ -89,12 +89,12 @@ public DebugExec() {
...
         try {
             if (PlotSquared.get() != null) {
-                File file = new File(PlotSquared.get().IMP.getDirectory(),
+                File file = new File(PlotSquared.imp().getDirectory(),
                     Settings.Paths.SCRIPTS + File.separator + "start.js");
                 if (file.exists()) {
                     init();
                     String script = StringMan.join(Files.readLines(new File(new File(
-                            PlotSquared.get().IMP.getDirectory() + File.separator
+                            PlotSquared.imp().getDirectory() + File.separator
```
dordsor21, line 97: If we're changing code that's been commented out, should it be fixed properly?

Sauilitired: Search and replace ^-^ But yes, probably. Although we should probably look over scripting in general. The scripting engine has been deprecated and most scripts are just broken.

Core/src/main/java/com/plotsquared/core/generator/SingleWorldGenerator.java
```diff
@@ -36,12 +36,13 @@
...
-    private Location dirt2 = new Location(null, 15, 2, 15);
-    private Location grass1 = new Location(null, 0, 3, 0);
-    private Location grass2 = new Location(null, 15, 3, 15);
+
+    private static final Location bedrock1 = Location.at("", 0, 0, 0);
+    private static final Location bedrock2 = Location.at("", 15, 0, 15);
+    private static final  Location dirt1 = Location.at("", 0, 1, 0);
+    private static final  Location dirt2 = Location.at("", 15, 2, 15);
+    private static final  Location grass1 = Location.at("", 0, 3, 0);
+    private static final  Location grass2 = Location.at("", 15, 3, 15);
```
dordsor21, line 45: double spaces

Core/src/main/java/com/plotsquared/core/command/DebugExec.java
AeSix: You should always do the best job you can with the tools, resources, skill and knowledge you have.  
Just because it's commented out don't mean it can't be used later, and if it's done properly, at the time it's touched, then it's more likely it can be used without pain.  That's not to say that other code won't change nullifying the changes, or that section all together, but if you're going to touch it, touch it like you're about to make love to it!

MattBDev:
I'm going to miss `IMP`

Sauilitired:
The thing that's left to do here, is to continue moving everything over to injection. We then merge this branch, and continue working on improvements in separate branches.

PR #2879, opened 2020-07-14
Features/v5/placeholders

Sauilitired:
Make placeholders less annoying and error prone. This will also allow us to add support for other placeholder APIs.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/placeholders/PlaceholderRegistry.java
+++ b/Core/src/main/java/com/plotsquared/core/util/placeholders/PlaceholderRegistry.java
@@ -1,0 +1,230 @@
+/*
+ *       _____  _       _    _____                                _
+ *      |  __ \| |     | |  / ____|                              | |
+ *      | |__) | | ___ | |_| (___   __ _ _   _  __ _ _ __ ___  __| |
+ *      |  ___/| |/ _ \| __|\___ \ / _` | | | |/ _` | '__/ _ \/ _` |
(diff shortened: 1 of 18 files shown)


Sauilitired:
This needs testing.

Core/src/main/java/com/plotsquared/core/util/placeholders/PlaceholderRegistry.java
```diff
@@ -0,0 +1,230 @@
...
+                return "0";
+            }
+            return String.valueOf(plot.getDenied().size());
+        });
+        this.createPlaceholder("has_build_rights", (player, plot) ->
+            plot.isAdded(player.getUUID()) ? "true" : "false");
+        this.createPlaceholder("currentplot_x", (player, plot) -> Integer.toString(plot.getId().getX()));
+        this.createPlaceholder("currentplot_y", (player, plot) -> Integer.toString(plot.getId().getY()));
+        this.createPlaceholder("currentplot_xy", (player, plot) -> plot.getId().toString());
+        this.createPlaceholder("currentplot_rating", (player, plot) -> Double.toString(plot.getAverageRating()));
```
N0tMyFaultOG, line 119: Does the rating placeholder work when you are using the like system?

Sauilitired: yes, uses the same system internally

Sauilitired:
just to be clear, this is currently broken and we need to delay hooking into mvdw until the server has loaded (ServerLoadEvent or w/e)

PR #2882, opened 2020-07-17
Revamp queue system to combine both Chunk Tasks and general Queueing.

dordsor21:
Requires update to WorldEdit (my current open PRs) for the regeneration stuff


Core/src/main/java/com/plotsquared/core/queue/QueueProvider.java
```diff
@@ -25,24 +25,17 @@
...
-                        e.printStackTrace();
-                        failed = true;
-                    }
-                }
+            @Override public QueueCoordinator getNewQueue(@Nonnull World world) {
                 try {
-                    return (LocalBlockQueue) fallback.getConstructors()[0].newInstance(world);
+                    return (QueueCoordinator) primary.getConstructors()[0].newInstance(world);
                 } catch (Throwable e) {
                     e.printStackTrace();
```
Sauilitired, line 40: We need better error handling here, and at the very least a more descriptive error (log something)

Core/src/main/java/com/plotsquared/core/queue/ChunkQueueCoordinator.java
```diff
@@ -114,15 +130,15 @@ private void storeCache(final int x, final int y, final int z, final BlockState
...
-    @Override public Location getMax() {
-        return Location.at(getWorld(), top.getX(), top.getY(), top.getZ());
+    @Override @Nonnull public Location getMax() {
+        return Location.at(getWorld().getName(), top.getX(), top.getY(), top.getZ());
     }
 
-    @Override public Location getMin() {
-        return Location.at(getWorld(), bot.getX(), bot.getY(), bot.getZ());
+    @Override @Nonnull public Location getMin() {
+        return Location.at(getWorld().getName(), bot.getX(), bot.getY(), bot.getZ());
```
Sauilitired, line 142: We should probably add an overloaded location getter for worldedit worlds

Core/src/main/java/com/plotsquared/core/queue/ChunkCoordinator.java
```diff
@@ -25,26 +25,23 @@
...
-
-    public OffsetLocalBlockQueue(LocalBlockQueue parent, int ox, int oy, int oz) {
-        super(parent);
-        this.ox = ox;
-        this.oy = oy;
-        this.oz = oz;
-    }
+import com.plotsquared.core.util.task.PlotSquaredTask;
+
+public abstract class ChunkCoordinator implements PlotSquaredTask {
```
Sauilitired, line 30: This class needs documentation, as it's the uppermost one. Same with all the methods, as this is part of the API

Core/src/main/java/com/plotsquared/core/components/ComponentPresetManager.java
```diff
@@ -191,11 +188,13 @@
 
                 BackupManager.backup(player, plot, () -> {
                     plot.addRunning();
+                    QueueCoordinator queue = plot.getArea().getQueue();
```
Sauilitired, line 191: I kinda feel like we should have some utility method to simplify this, as this is also done in the set command, and it'd be nice to have it exposed via API. Dunno how tho

Bukkit/src/main/java/com/plotsquared/bukkit/generator/DelegatePlotGenerator.java
```diff
@@ -64,16 +64,15 @@ public DelegatePlotGenerator(ChunkGenerator chunkGenerator, String world) {
...
         Location min = result.getMin();
         int chunkX = min.getX() >> 4;
         int chunkZ = min.getZ() >> 4;
         Random random = new Random(MathMan.pair((short) chunkX, (short) chunkZ));
         try {
             ChunkGenerator.BiomeGrid grid = new ChunkGenerator.BiomeGrid() {
-                @Override
-                public void setBiome(@Range(from = 0, to = 15) int x,
+                @Override public void setBiome(@Range(from = 0, to = 15) int x,
                     @Range(from = 0, to = 15) int z, @Nonnull Biome biome) {
```
Sauilitired, line 76: Get rid of the jb annotations

Core/src/main/java/com/plotsquared/core/components/ComponentPresetManager.java
dordsor21: Technically QueueProvider#of(com.sk89q.worldedit.world.World) but it's easier to use the PlotArea getQueue than giving every class the WorldUtil and then converting a String to a WorldEdit World

Sauilitired: I meant the entire component setting, not just the queue getter. Should have been clearer about that 🌵

dordsor21: Ahh right, it's exposed in Plot#setComponent(String component, Pattern blocks, @Nullable QueueCoordinator queue) I guess? If the queue isn't null it writes the blocks to the queue and doesn't enqueue. If the queue is null it makes its own and then enqueues. I'll document that everywhere...

PR #2888, opened 2020-07-25
Fix problem with config creation

Sauilitired:
This fixes an issue where PlotSquared would try to convert databases created in v5 from v4 to v5, which obviously fails, because they never used v4.

This is a quite major issue and I hope this fix resolves it. Although, it should be tested with migration from v4 to v5 and then a fresh v5 server.


N0tMyFaultOG:
I've given this a try earlier and it works well.

PR #2890, opened 2020-08-06
Fixes an OfflinePlayerUtil breakaged caused by NMS Changes

MeFisto94:
## Overview
A user reported https://athion.net/ISPaster/paste/view/c41c30820c3a4ab999b681b4f7172a60 on the discord and the cause is
```java
java.lang.RuntimeException: No such method
	at com.plotsquared.core.util.ReflectionUtils.callMethod(ReflectionUtils.java:160) ~[?:?]
	at com.plotsquared.bukkit.util.OfflinePlayerUtil.getWorldServerNew(OfflinePlayerUtil.java:107) ~[?:?]
```

I don't know how to trigger this issue specifically, but something needs to request an OfflinePlayer.
Also note that this code will be gone in v6 anyway, so it's more of a "hack" anyway.

## Description
According to my local paper workspace, the signature has changed to `public WorldServer getWorldServer(ResourceKey<World> resourcekey)`.

The problem that specifically needs testing is that in my workspace, `DimensionManager` is now called `World`, which according to the stacktrace is not the case, because then it would've failed at getField.
My IntelliJ shows that for Spigot it might be called `DimensionManager` while it's called `World` for Paper.
I am patching my workspace to the latest version, but since this change needs testing anyway, could someone maybe have a look at both spigot and paper and see if this discrepancy really exists (maybe it's between 1.16 and 1.16.1?)

## Checklist


N0tMyFaultOG:
I gave it a test earlier, but since I couldn't reproduce it with the origin version either, I cannot guarantee it fixes the issue reported. But from the diff, it seems fine to me.

MeFisto94:
@SirYwell said it didn't work for that user:
https://discordapp.com/channels/268444645527126017/400537494300655616/741559515786510356
https://media.discordapp.net/attachments/741417132784615545/741553100409667664/unknown.png

That being said with a small plugin you could reproduce that issue, namely when using PlotPlayer#wrap (orAPI#wrapPlayer) on either an UUID or an OfflinePlayer. I don't remember which, if only a UUID or an OfflinePlayer works, but that will get you there.

MeFisto94:
FIXES PS-94

PR #2897, opened 2020-08-14
Add service system and rewrite Auto to use the service system

Sauilitired:
Completely untested


Core/src/main/java/com/plotsquared/core/services/plots/AutoService.java
```diff
@@ -0,0 +1,139 @@
...
+            return autoQuery.sizeX == 1 && autoQuery.sizeZ == 1;
+        }
+
+    }
+
+
+    final class MultiPlotService implements AutoService, Predicate<AutoQuery> {
+
+        @Override public List<Plot> handle(@Nonnull final AutoQuery autoQuery) {
+            /* TODO: Add timeout? */
```
N0tMyFaultOG, line 121: Probably.

PR #2916, opened 2020-09-20
Feature/v6/arkitektonika

SirYwell:
## Description

This PR implements Arkitektonika support in PlotSquared.
Todos/Discussion:
- [ ] Change urls in config to a properly hosted instance at some point
- [ ] Maybe change temp directory path
- [x] ~~Make sure merged plots are supported (doesn't work right now due to an issue in Plot.java, that should be addressed separately)~~ 
- [ ] Make sure that the legacy schematic upload is at least supported with FAWE (doesn't work with WE and P2 5.12.x)

I've also refactored some methods in SchematicHandler to use CompletableFutures instead of RunnableVals and cleaned up/deprecated some code there.

## Checklist

- [ ] I ensured my changes do not break other parts of the code

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
@@ -415,6 +415,24 @@ public class Settings extends Config {
             " - Downloads can be deleted by the user",
             " - Supports plot uploads, downloads and saves",}) public static String URL =
             "https://ark.jacobandersen.dev/";
+        @Comment({"Whether or not the legacy web interface will be used for /plot download and /plot save",
+                 "Note that this will be removed in future versions. Updating to Arkitektonika is highly suggested"})
+        public static boolean LEGACY_WEBINTERFACE = false;
+    }
+
+    @Comment("Schematic web interface related settings")
+    public static class Arkitektonika {
+
+        @Comment("The url of the backend server (Arkitektonika)")
+        public static String BACKEND_URL = "https://ark.jacobandersen.dev/";
+
+        @Comment({"The url used to generate a download link from.",
+                        "{key} will be replaced with the generated key"})
+        public static String DOWNLOAD_URL = "https://sw.jacobandersen.dev/download/?key={key}";
+
+        @Comment({"The url used to generate a deletion link from.",
+                         "{key} will be replaced with the generated key"})
+        public static String DELETE_URL = "https://sw.jacobandersen.dev/delete/?key={key}";
     }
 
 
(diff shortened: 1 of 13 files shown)


Bukkit/build.gradle
```diff
@@ -103,8 +103,10 @@ shadowJar {
     dependencies {
         include(dependency(":PlotSquared-Core"))
         include(dependency("io.papermc:paperlib:1.0.4"))
+        include(dependency("net.kyori:examination-api:1.0.0"))
+        include(dependency("net.kyori:examination-string:1.0.0"))
```
N0tMyFaultOG, line 107: We can safely remove the old `-SNAPSHOT` dependency located further down the file.
```gradle
include(dependency("net.kyori:examination-api:1.0.0-SNAPSHOT"))
include(dependency("net.kyori:examination-string:1.0.0-SNAPSHOT"))
````

File: Core/src/main/java/com/plotsquared/core/configuration/Settings.java [hunk 2]
```diff
@@ -411,12 +411,31 @@ public static boolean convertLegacy(File file) {
...
+        @Comment({"Whether or not the legacy web interface will be used for /plot download and /plot save",
+                 "Note that this will be removed in future versions. Updating to Arkitektonika is highly suggested"})
+        public static boolean LEGACY_WEBINTERFACE = false;
+    }
+
+    @Comment("Schematic web interface related settings")
+    public static class Arkitektonika {
+
+        @Comment("The url of the backend server (Arkitektonika)")
+        public static String BACKEND_URL = "https://ark.jacobandersen.dev/";
```
N0tMyFaultOG, line 430: It would be cool we could host Arkitektonika on Athion (@dordsor21 👀) so we have it is centralized alongside our other stuff and create an ALIAS Record so we can have it under `*.intellectualsites.com`

N0tMyFaultOG (changes requested):

N0tMyFaultOG:
Putting it here as well, using `/plot save` without using the legacy web interface results in an exception rather than telling people what's wrong https://haste.athion.net/yidomepuvo.md

SirYwell:
The exception occurs on every invalid commadn at the moment. I'd address that separately

Core/src/main/java/com/plotsquared/core/configuration/Settings.java
Sauilitired: I can host on the IS box, we just need to change the default before pushing v6.

Sauilitired (approved):

Core/src/main/java/com/plotsquared/core/configuration/Settings.java
N0tMyFaultOG: That would work as well. What do you mean by change the default? The existing web interface?

SirYwell: I added a note for this on the project

N0tMyFaultOG (approved):
We're gonna exchange upload urls for prod.

PR #2934, opened 2020-11-22
Cancel boat placement on interactable blocks, fixes PS-175

traksag:
## Overview
**Fixes [PS-175](https://issues.intellectualsites.com/issue/PS-175)**

## Description
Use `EntityPlaceEvent` instead of `PlayerInteractEvent` for boats. There's no robust way to detect whether a boat is about to be placed using only the `PlayerInteractEvent`.

## Checklist


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1009,17 +1011,11 @@ public void onInteract(PlayerInteractEvent event) {
...
-                    case BIRCH_BOAT:
                     case CHEST_MINECART:
                     case COMMAND_BLOCK_MINECART:
-                    case DARK_OAK_BOAT:
                     case FURNACE_MINECART:
                     case HOPPER_MINECART:
-                    case JUNGLE_BOAT:
                     case MINECART:
-                    case OAK_BOAT:
-                    case SPRUCE_BOAT:
```
SirYwell, line 1022: According to [its previus documentation](https://docs.improbable-game.fr/spigot-api/1.14.2/org/bukkit/event/entity/EntityPlaceEvent.html) this event wasn't triggered for boats pre 1.15. I'd prefer to keep this part, as while it doesn't resolve the bug on 1.13 and 1.14, it will keep functionality the same on those version.

In general, I'd prefer to handle all entity placements that can be handled that way using that event, as it's safer and cleaner.

SirYwell (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1068,6 +1070,34 @@ public void onInteract(PlayerInteractEvent event) {
...
+        if (area == null) {
+            return;
+        }
+        PlayerBlockEventType eventType = PlayerBlockEventType.PLACE_VEHICLE;
+        Block block = event.getBlock();
+        BlockType blockType = BukkitAdapter.asBlockType(block.getType());
+        Location location = BukkitUtil.getLocation(block.getLocation());
+        if (!PlotSquared.get().getEventDispatcher()
+            .checkPlayerBlockEvent(pp, eventType, location, blockType, true)) {
+            event.setCancelled(true);
```
dordsor21, line 1097: you ought to `event.setUseInteractedBlock(Event.Result.DENY);` here as well

traksag: There's no such method for `EntityPlaceEvent`

dordsor21 (approved):

SirYwell (approved):

N0tMyFaultOG (approved):

PR #2935, opened 2020-11-28
NullPointerException fix - Chat command

luvily:
# Issue
If a player without a plot tries to enable the plot chat there is thrown a NullPointerException.
https://github.com/IntellectualSites/PlotSquared/blob/v5/Core/src/main/java/com/plotsquared/core/command/Chat.java


N0tMyFaultOG:
That doesn't appear to be an issue.

PR #2937, opened 2020-12-05
Fix PlotRangeIterator

SirYwell:
This aimes to fix an issue causing the server to freeze/OOM on actions like `/plot info`, `/plot home`, and `/plot download` on merged plots.

It works from my testings but I'd like someone else to approve that.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/Plot.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/Plot.java
@@ -2299,12 +2299,13 @@ public class Plot {
                 continue;
             }
             boolean merge = true;
-            PlotId bot = PlotId.of(current.getId().getX(), current.getId().getY());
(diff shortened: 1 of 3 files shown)


dordsor21:
What's changed here that means OOMs/freezes no longer occur? Are there any logs to be able to compare the changes?

SirYwell:
The main issue was that the iterator didn't include the last plot in some cases, means the test `singlePlotIterator` failed on `range.hasNext()`. That caused the `getRegions` method to run endless for the south and norht directions. 
It also returned wrong results in some cases as x and y are set to start at the beginning but y is reset to 0 afterwards and the returned id always used start.x + x and start.y + y.

The tests should cover these issues.

dordsor21 (approved):
That makes sense. Thanks

N0tMyFaultOG (approved):

PR #2941, opened 2021-01-02
Use WE Expressions instead of js evaluation

SirYwell:
This replaces the old P2 `Expression` with the WE `Expression`. The WE expression syntax is powerful enough to replace the evaluation of JavaScript expressions.

**Note**: Previously, `{args}` was replaced with the amount of plots a user has, now this is the `plots` variable in the expression. So this is a breaking change that needs to be noted for server owners.

The scripting implementation in P2 is already outdated and will cause further issues in Java >=15 due to the removal of the nashorn engine, so it's better to not depend on that anymore for basic and often-used features like cost evaluation.


dordsor21:
Are we able to remove any libraries now? Presumablythe debugexec used the same though?

N0tMyFaultOG (changes requested):
```java
player.sendMessage(
                        TranslatableCaption.of("economy.removed_balance"),
                        Template.of("money", String.valueOf(cost))
                );
```
I'd call here `this.econHandler.format(cost)` instead of parsing the raw amount. Same on further occurrences.

SirYwell:
> 
> 
> Are we able to remove any libraries now? Presumablythe debugexec used the same though?

We're currently using nashorn, which is bundled with the JRE if below Java 15. To keep the DebugExec features, we'd either need to shade our own JS engine (rhino, graaljs) or depend on rhino bundled with with WE. GraalJS is a) very large, as long as we cannot get rid of some transitive dependencies and b) hacky to implement when loading stuff with a different ClassLoader (see [here](https://java.programmingpedia.net/en/knowledge-base/56712178/graalvm---no-language-and-polyglot-implementation-was-found-on-the-classpath)). The dodumentation I found for rhino is archived, links are broken and so I wasn't interested spending more time looking into it.
Anyways, the DebugExec stuff should be addressed separately.

N0tMyFaultOG (approved):

dordsor21:
> > Are we able to remove any libraries now? Presumablythe debugexec used the same though?
> 
> We're currently using nashorn, which is bundled with the JRE if below Java 15. To keep the DebugExec features, we'd either need to shade our own JS engine (rhino, graaljs) or depend on rhino bundled with with WE. GraalJS is a) very large, as long as we cannot get rid of some transitive dependencies and b) hacky to implement when loading stuff with a different ClassLoader (see [here](https://java.programmingpedia.net/en/knowledge-base/56712178/graalvm---no-language-and-polyglot-implementation-was-found-on-the-classpath)). The dodumentation I found for rhino is archived, links are broken and so I wasn't interested spending more time looking into it.
> Anyways, the DebugExec stuff should be addressed separately.

Fair enough

dordsor21 (approved):

PR #2942, opened 2021-01-02
Implement CaptionLoader API to be used by third party plugins/addons

SirYwell:
This allows to load resources from other classloaders than the P2 one. 
Therefore, we can use this in addons too to manage messages the same way.


N0tMyFaultOG (approved):

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
```diff
@@ -93,7 +116,7 @@ private CaptionLoader() {
      * @see Files#list(Path)
      * @see #loadSingle(Path)
      */
-    @Nonnull public static CaptionMap loadAll(@Nonnull final Path directory) throws IOException {
+    public @NonNull CaptionMap loadAll(final @NonNull Path directory) throws IOException {
```
dordsor21, line 119: why.. D:

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
```diff
@@ -117,18 +140,11 @@ private CaptionLoader() {
...
-    @Nonnull public static CaptionMap loadSingle(@Nonnull final Path file) throws IOException {
-        final String fileName = file.getFileName().toString();
-        final Matcher matcher = FILE_NAME_PATTERN.matcher(fileName);
-        final Locale locale;
-        if (matcher.matches()) {
-            locale = Locale.forLanguageTag(matcher.group(1));
-        } else {
-            throw new IllegalArgumentException(fileName + " is an invalid message file (cannot extract locale)");
-        }
+    public @NonNull CaptionMap loadSingle(final @NonNull Path file) throws IOException {
```
dordsor21, line 146: \>.>

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
```diff
@@ -202,4 +215,27 @@ private static boolean patch(final Map<String, String> map, final Locale locale)
...
+     * The pattern is required to have (at least) one capturing group, as this is used to access the locale
+     * tag.The function will throw an {@link IllegalArgumentException} if the matcher doesn't match the file name
+     * of the input path. The language tag is loaded using {@link Locale#forLanguageTag(String)}.
+     *
+     * @param pattern the pattern to match and extract the language tag with.
+     * @return a function to extract a locale from a path using a pattern.
+     * @see Matcher#group(int)
+     * @see Path#getFileName()
+     */
+    public static @NonNull Function<Path, Locale> patternExtractor(final @NonNull Pattern pattern) {
```
dordsor21, line 230: ...

dordsor21 (dismissed):

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
Citymonstret: Don't question it :eyes: We're going to switch to checker-qual anyway as those annotations are actually following the new Java standard, so they'll all get moved around :P

dordsor21: I mean it does make sense tbf, it's the "CaptionMap" that's not gonna be null, not "public". Just looks fugly when the rest of P2 doesn't do it that way

Core/src/main/java/com/plotsquared/core/configuration/caption/load/DefaultCaptionProvider.java
```diff
@@ -0,0 +1,71 @@
...
+     *                    {@code String.format(toFormat, Locale#toString)}
+     * @return a caption provider using string formatting to determine resource urls.
+     */
+    static DefaultCaptionProvider forClassLoaderFormatString(ClassLoader classLoader, String toFormat) {
+        return forClassLoader(classLoader, locale -> String.format(toFormat, locale.toString()));
+    }
+
+    /**
+     * Loads default translation values for a specific language and returns it as a map.
+     * If no default translation exists, {@link null} is returned. A returned map might be empty.
```
Citymonstret, line 65:
```suggestion
     * If no default translation exists, {@code null} is returned. A returned map might be empty.
```

Core/src/main/java/com/plotsquared/core/configuration/caption/load/DefaultCaptionProvider.java
```diff
@@ -0,0 +1,71 @@
...
+     * Returns a DefaultCaptionProvider that loads captions from a {@link ClassLoader}'s resources.
+     * The resource urls are determined by replacing the first occurrence of {@code %s} in the string with
+     * {@link Locale#toString()}.
+     *
+     * @param classLoader the class loader to load caption resources from.
+     * @param toFormat    a string that can be formatted to result in a valid resource url when calling
+     *                    {@code String.format(toFormat, Locale#toString)}
+     * @return a caption provider using string formatting to determine resource urls.
+     */
+    static DefaultCaptionProvider forClassLoaderFormatString(ClassLoader classLoader, String toFormat) {
```
Citymonstret, line 59:
```suggestion
    static @NonNull DefaultCaptionProvider forClassLoaderFormatString(@NonNull ClassLoader classLoader, @NonNull String toFormat) {
```

Core/src/main/java/com/plotsquared/core/configuration/caption/load/DefaultCaptionProvider.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Returns a DefaultCaptionProvider that loads captions from a {@link ClassLoader}'s resources.
+     * The resource urls are determined by applying the given function to a locale.
+     *
+     * @param classLoader the class loader to load caption resources from.
+     * @param urlProvider the function to get an url from a locale.
+     * @return a caption provider using a function to determine resource urls.
+     */
+    static DefaultCaptionProvider forClassLoader(ClassLoader classLoader, Function<Locale, String> urlProvider) {
```
Citymonstret, line 45:
```suggestion
    static @NonNull DefaultCaptionProvider forClassLoader(@NonNull ClassLoader classLoader, @NonNull Function<@NonNull Locale, @NonNull String> urlProvider) {
```

Core/src/main/java/com/plotsquared/core/configuration/caption/load/DefaultCaptionProvider.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Loads default translation values for a specific language and returns it as a map.
+     * If no default translation exists, {@link null} is returned. A returned map might be empty.
+     *
+     * @param locale the locale to load the values for.
+     * @return a map of default values for the given locale.
+     */
+    @Nullable Map<String, String> loadDefaults(final @NonNull Locale locale);
```
Citymonstret, line 70:
```suggestion
    @Nullable Map<@NonNull String, @NonNull String> loadDefaults(final @NonNull Locale locale);
```

Core/src/main/java/com/plotsquared/core/configuration/caption/load/ClassLoaderCaptionProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Function;
+
+import static com.plotsquared.core.configuration.caption.load.CaptionLoader.loadFromReader;
+
+class ClassLoaderCaptionProvider implements DefaultCaptionProvider {
```
Citymonstret, line 43: Might as well make this class final as nothing in the package will extend it.

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
Citymonstret, line 230: If this is how the pattern is used, wouldn't it make more sense to make it a `Predicate<Path>`? You can then use `pattern.asPredicate()` to pass a pattern.

Citymonstret (changes requested):

Core/src/main/java/com/plotsquared/core/configuration/caption/load/CaptionLoader.java
SirYwell: We're extracting the language tag from the path with that pattern, whereas the predicate would only tell us *if* it matches

Citymonstret: Ah ye my bad, didn't se it being used underneath. Then it's all good!

Citymonstret (approved):

Issue #3010, opened 2021-05-11
Player receives 


ThisIsIPat:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

The entering or leaving player receives the feedback instead of the actual owner of the plot.

This affects both the ``notify-enter`` and the ``notify-leave`` flag.

### To Reproduce

Prerequisite: Own a plot, another player

1. Stand on the plot
2. Set one of the related plot flags, for example by using ``/p flag set notify-enter``.
3. Have a second player enter the plot.

### Expected behaviour

The owner receives a notification of the player entering / leaving the plot.

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/36f47999c91e4bac88016fbd8c2a047a

### PlotSquared Version

PlotSquared-6.0.0-SNAPSHOT (Private Fork, bug reproducible in Vanilla)

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

``/plot debugpaste`` was not successfully executed the first time:
request must contain a file list
(Worked second time)

The bug is intuitively comprehensible within the code: Only two lines.

PR #3011, opened 2021-05-11
Fixed feedback recipient for notify flags

ThisIsIPat:
The entering or leaving player receives the feedback instead of the actual owner of the plot.

This affects both the ``notify-enter`` and the ``notify-leave`` flag.

[Related issue](https://github.com/IntellectualSites/PlotSquared/issues/3010)


NotMyFault (approved):
Thanks for the pr!

Issue #3013, opened 2021-05-12
Wrong affordance check for economy prices


ThisIsIPat:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

If one is able to afford a plot and the economy is supported, the message "Vault is not installed. No Vault-like economy service can be provided." appears.

### To Reproduce

Prerequisites: Vault enabled. Economy hooked into Vault. economy.use is set to true in the worlds.yml within the section of the respective world for reproducing the bug.

1. Visit a non-claimed plot
2. Attempt to execute ``/p claim``

### Expected behaviour

The purchase is made.

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/37408d38839e4b9485264a3a703574d9

### PlotSquared Version

PlotSquared-6.0.0-SNAPSHOT (Private Fork, bug reproducible in Vanilla)

### Checklist

- [X] I have included a Plot debugpaste.
- [ ] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

ThisIsIPat:
PR currently worked on

NotMyFault:
I'm pretty sure this is down to me swapping the returns when I added the vault check.

ThisIsIPat:
Edit: Not reproducible with cost "0".

PR #3014, opened 2021-05-12
Pre-eliminating if-guard for economy check within cost condition

ThisIsIPat:
The economy check within the ``auto`` and ``claim`` subcommands is generally causing an error message, stating Vault is not installed, if the player can actually afford claiming the plot.

**Fixes https://github.com/IntellectualSites/PlotSquared/issues/3013**


NotMyFault (approved):
Thanks for the pr!

SirYwell (approved):

Issue #3034, opened 2021-05-17
Burning player can decrease water level of cauldron by extinguishing themselves with it without having interaction permissions


ThisIsIPat:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

If a player is on fire, it may jump into a cauldron on any plot, specifically the ones it doesn't own or is trusted on, and modify its water level by extinguishing themselves

### To Reproduce

Prerequisites: Two players

1. Player A claims a plot
2. Player A places a cauldron
3. Player A fills the cauldron with water
(3.5) Player A does NOT trust Player B
4. Player B sets itself on fire
5. Player B jumps into the cauldron filled with water on Player As plot

### Expected behaviour

The cauldrons water level does not change as long as the player doesn't have permissions to interact with the plot.
It may or may not extinguish the fire on the player.

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/0c59391058b9403a94b320c4db3f95f1

### PlotSquared Version

PlotSquared-6.0.0-SNAPSHOT (Private Fork, bug reproducible in Vanilla)

### Checklist

- [X] I have included a Plot debugpaste.
- [ ] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

NotMyFault:
Fixed in https://github.com/IntellectualSites/PlotSquared/commit/7c4a85325c6781cce503c6f7a67b40dbd6b3fd94

PR #3035, opened 2021-05-17
Added event handler for cauldron water level modification to prevent untrusted players interacting with a plot

ThisIsIPat:
Changes in the water level of a cauldron are now cancelled unless a player modifies it and that player has permissions to interact with the plot

Valid permissions:
``plots.admin.interact.road``
``plots.admin.interact.unowned``
``plots.admin.interact.other``
``Player is trusted or owner of the plot``

**Fixes https://github.com/IntellectualSites/PlotSquared/issues/3034**


ThisIsIPat:
Note: It still extinguishes the player. It merely doesn't change the water level.

Another note: The water level is now also not modified by rain or, since 1.17, dripstone. This could possibly be implemented via a flag. If requested, the PR will be modified so that rain and dripstone will still modify a cauldrons water level.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -479,6 +480,48 @@ public void onBlockSpread(BlockSpreadEvent event) {
...
+        Location location = BukkitUtil.adapt(event.getBlock().getLocation());
+        PlotArea area = location.getPlotArea();
+        if (area == null) {
+            return;
+        }
+        Plot plot = area.getPlot(location);
+        // TODO Add flag to allow / deny 1.17 dripstone-based machines
+        // TODO Add flag to allow / deny modification by rain
+        if (entity instanceof Player) {
+            Player player = (Player) entity;
```
NotMyFault, line 495:
```suggestion
        if (entity instanceof Player player) {
```

NotMyFault (changes requested):

NotMyFault:
> Note: It still extinguishes the player. It merely doesn't change the water level.

Afaik the enum CauldronLevelChangeEvent.ChangeReason takes an extinguish cause determining whether the entity should be extinguished or not.

ThisIsIPat:
Should cauldron modification by rain be disabled by default?

NotMyFault:
> Should cauldron modification by rain be disabled by default?

I'd enable it. We don't cancel too many natural events on default if not really necessary, see liquid flow flags e.g.

NotMyFault (approved):
Thanks for the pr (again)!

ThisIsIPat:
@NotMyFault Modified event handling so the event represents being cancelled by PlotSquared properly (Instead of having the new cauldron water level be the old one).

Requesting re-review.

NotMyFault (approved):

SirYwell (approved):

PR #3060, opened 2021-05-20
Implemented the ability to change titles to actionbar on plot entry

### LINKED ISSUE #2993: hotbar msg instead title or a separate title flag
If I activate the title flag the user get infos about the flag with a plot-join. This title I find are to big and would like use the hotbar for the same infos.

darbyjack:
## Overview

**Fixes #2993**

## Description
This PR implements a config option to send the "titles" of plot entry to a player as an ActionBar instead of a Title.

## Checklist


Core/src/main/java/com/plotsquared/core/configuration/Settings.java
```diff
@@ -58,6 +58,8 @@
     public static int TITLES_STAY = 50;
     @Comment("Plot titles fading out (duration in ticks)")
     public static int TITLES_FADE_OUT = 20;
+    @Comment("Show the plot titles as an actionbar?")
+    public static boolean TITLES_AS_ACTIOBAR = false;
```
NotMyFault, line 62: Missing 'N' in actioNbar

Core/src/main/java/com/plotsquared/core/listener/PlotListener.java
```diff
@@ -303,8 +303,17 @@ public boolean plotEntry(final PlotPlayer<?> player, final Plot plot) {
                             Template plotTemplate = Template.of("plot", lastPlot.getId().toString());
                             Template worldTemplate = Template.of("world", player.getLocation().getWorldName());
                             Template ownerTemplate = Template.of("owner", owner);
-                            final Consumer<String> userConsumer = user -> player
-                                    .sendTitle(header, subHeader, plotTemplate, worldTemplate, ownerTemplate);
+
+                            Consumer<String> userConsumer;
+
+                            if (Settings.TITLES_AS_ACTIOBAR) {
```
NotMyFault, line 309: Missing N

NotMyFault (changes requested):

Core/src/main/java/com/plotsquared/core/player/PlotPlayer.java
```diff
@@ -841,6 +842,27 @@ public void sendTitle(
                 .title(titleComponent, subtitleComponent, times));
     }
 
+    /**
+     * Sends an actionbar to the player on plot entry.
+     * This takes in two captions as it will internally merge
+     * the captions together to send as a single component.
```
NotMyFault, line 848: We might need to outline that this method is explicitly used for titles as actionbars. Because you very likely can't/wont use this method for forming regular actionbars.

darbyjack: So do you just want an updated explanation of it in the javadocs?

NotMyFault: Yeah. The current describing sounds like it can be used everywhere to send actionbars.

darbyjack: Done!

NotMyFault: Thanks

NotMyFault (approved):

Core/src/main/java/com/plotsquared/core/player/PlotPlayer.java
```diff
@@ -841,6 +842,32 @@ public void sendTitle(
...
+     * of the project.
+     *
+     * This takes in two captions as it will internally merge
+     * the captions together to send as a single component.
+     *
+     * @param header Header
+     * @param subHeader Subheader
+     * @param replacements Variable replacements
+     */
+    public void sendPlotEntryBar(
```
SirYwell, line 859: I'd prefer to not have such a specific method in PlotPlayer. It's probably better to just do that stuff in the plot listener too and use getAudience there *or* just say we only use the header for the action bar and have a more general sendActionBar(Caption, Template...). That way, the content is also fully modifiable without the hardcoded `|`.

Not sure what the others think about it, but I'm in favor of second approach. That behavior just needs to be outlined in the settings comment then.

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/player/PlotPlayer.java
NotMyFault: This method is intended to be for titles only as of drafting it no implementation of actionbars but for titles has been considred.
If we want to add a more generalized method I'm not opposed to it.

SirYwell: > This method is intended to be for titles only as of drafting it no implementation of actionbars but for titles has been considred.

Yes, that's too specific to end up in PlotPlayer in my opinion.The current implementation is also limited in the way it can be configured, that's why I brought up the second approach.

darbyjack: I believe I attempted that original approach but after talking with NMF I could not (probably a misunderstanding on my end) find a reliable way to combine those two captions for the title and subtitle into a single spaced proper component to pass through cleanly.

SirYwell: Yeah combining them always has some hardcoded magic and probably will never fulfill everyone's needs. That's why I'm suggesting to only use the title (badly worded before from my side) and ignore the subtitle for the action bar. If that's documented in the settings comment, I think it should be clear enough, we don't need a specialized method but only a `sendActionBar` one, and people can configure it the way they want.

darbyjack: So you would only like to show the plot name that they are entering and drop the owner information when they are using actionbar?

SirYwell: You can change the title entry in the messages to put it back in - it wouldn't be there by default tho. But I think that's still better than hardcoding a separator.

darbyjack: Implemented!

NotMyFault:
This does not resolve inner template references, no? E.g. A `<prefix>` in a caption.

Core/src/main/java/com/plotsquared/core/player/PlotPlayer.java
```diff
@@ -841,6 +841,20 @@ public void sendTitle(
...
+     * Method designed to send an ActionBar to a player.
+     *
+     * @param caption Caption
+     * @param replacements Variable replacements
+     */
+    public void sendActionBar(
+            final @NonNull Caption caption,
+            final @NonNull Template... replacements
+    ) {
+        final Component component = MiniMessage.get().parse(caption.getComponent(this), replacements);
```
SirYwell, line 854: To fix the issue with <prefix> not getting replaced, this should be solved similar to how it is written in `sendMessage` below (means call `replace` on the raw String returned by `getComponent(this)`).

SirYwell (changes requested):
I think the issue mentioned by @NotMyFault should be easy to fix, everything else is awesome!

SirYwell (approved):
Thank you very much

Issue #3078, opened 2021-06-01
Players can attack animals on plots they don't have to permission to PvE on


ThisIsIPat:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

No matter whether the PvE flag or animal-attack flag is set or a player is added to a plot, an attack on an animal on that plot is never cancelled.

### To Reproduce

Prerequisites: A plot ``Player A`` doesn't own or is trusted on with an animal on it.

1. Player A attacks the animal while both it and the animal are on the plot.

### Expected behaviour

The animal does not get damaged, as the player has no permission to interact with the animal, especially damaging it.

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/b66141c809d04a89b6db5b5c6ad73c2d

### PlotSquared Version

PlotSquared-6.0.0-SNAPSHOT (Private Fork, bug reproducible in Vanilla)

### Checklist

- [X] I have included a Plot debugpaste.
- [ ] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

PR is worked on, but I can tell that the code is a bit iffy. I hope I'll complete with an implementation you'd agree with.

ThisIsIPat:
The bug is limited to animals only (Not other entities)

NotMyFault:
Fixed in https://github.com/IntellectualSites/PlotSquared/commit/8a244d12fc5fa7056cd6aae8cd1c89875a817789

PR #3079, opened 2021-06-01
Fix if-condition limited to roads for animal attack cancelling on plots

ThisIsIPat:
The if-condition for checking whether the user (doesn't!) have the administrative PvE permission was only performed if the attacked animal was on a road - not on a plot!

Similar logic to my solution found in the else-Block (Entity category check) directly below the vehicle check.

**Fixes https://github.com/IntellectualSites/PlotSquared/issues/3078**


NotMyFault (approved):
Thanks for the contribution, good catch! Something at least I didn't spot yet 👀

SirYwell (approved):

PR #3178, opened 2021-07-24
Fix/v6/click event actions

SirYwell:
## Overview

This is a different approach to the click event handling introduced in https://github.com/IntellectualSites/PlotSquared/commit/51ff043af98a511293017462dbba4c14babffed5. 

## Description

Instead of replacing the elements in the MiniMessage string each time, the strings are instead parsed, transformed and serialized back to strings. That way, the handling does not depend on the MiniMessage grammar but only on adventure. This makes it very robust and more flexible. By default, all click events are removed, but you can enable e.g. clickable urls if you want.

I also added JUnit 5 as dependency for tests. I left JUnit 4 there because I didn't want to update old tests with this PR.

## Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/configuration/caption/CaptionUtility.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/caption/CaptionUtility.java
@@ -25,11 +25,41 @@
  */
 package com.plotsquared.core.configuration.caption;
 
+import com.plotsquared.core.configuration.Settings;
 import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.flag.PlotFlag;
+import com.plotsquared.core.plot.flag.implementations.DescriptionFlag;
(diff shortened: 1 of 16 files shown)


Core/src/main/java/com/plotsquared/core/configuration/caption/CaptionUtility.java
```diff
@@ -66,4 +96,21 @@ public static String format(
...
     }
 
+    public static String stripClickEvents(final @NonNull String miniMessageString) {
+        // parse, transform and serialize again
+        Component component = MiniMessage.get().parse(miniMessageString);
+        component = CLICK_STRIP_TRANSFORM.transform(component);
+        return MiniMessage.get().serialize(component);
+    }
+
+    public static String stripClickEvents(
```
NotMyFault, line 106: Javadocs?

NotMyFault (approved):

dordsor21 (approved):

Issue #3185, opened 2021-07-27
Formatting mistake in fail feedback for /p kick


ThisIsIPat:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

Using the kick command on a plot the command executor does not own triggers the caption ``permission.no_permission``, but doesn't provide a template argument for ``<node>``. See screenshot for result.

_A debug paste is not required to understand the issue._

### To Reproduce

Prerequisites: Player A, who is not an admin, Plot by Player B

1. Player A visits Plot B
2. Player A executes ``/p kick (any player)``

### Expected behaviour

Usage of ``permission.no_plot_perms`` instead, as it requires no templates and more accurately describes why / that a player cannot perform this command on the plot they are on.

### Screenshots / Videos

![You are lacking the permission node: <node>](https://cdn.discordapp.com/attachments/838863310756904990/868174528541519912/unknown.png)

Note: Although the screenshot uses german text, the problem lies in the code which is missing the <node> template for ``permission.no_permission`` and should be reproducible in the English version

### Error log (if applicable)

-

### Plot Debugpaste

-

### PlotSquared Version

PlotSquared-6.0.6 (Private Fork, bug reproducible in Vanilla)

### Checklist

- [ ] I have included a Plot debugpaste.
- [ ] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

PR is underway

PR #3186, opened 2021-07-27
Fix /p kick msg

ThisIsIPat:
``/p kick`` doesn't format the missing permissions as intended - <node> cannot be formatted as no argument is given, and no argument really makes the most sense when using ``permission.no_permission``. My suggestion is using ``permission.no_plot_perms``, instead, as it better describes that you are not allowed to execute this command in general, as you are not the owner of the plot. ("You must be the plot owner to perform this action.")

**Fixes https://github.com/IntellectualSites/PlotSquared/issues/3185**


NotMyFault (approved):
LGTM!

NotMyFault:
Tho, that addresses linked issue, but the message isn't accurate still. The condition contains a permission check that responds with the message that you need to be the plot owner seems a bit odd to me.

SirYwell (approved):

PR #3191, opened 2021-07-31
Fix sponge absorbing water across plot borders

traksag:
## Overview

Currently sponges placed on plots also absorb water on the road. E.g. waterlog a plot wall slab, place some water next to it on the road, then place a sponge on your plot next to the waterlogged slab. The water in the waterlogged slab and on road will get removed.

## Description

Use the `SpongeAbsorbEvent` to deal with absorbing water across plot borders (and area borders). This event was added to Bukkit right after it updated to 1.13.

I copied the `StructureGrowEvent` listener and changed it around a bit:
- Removed this seemingly useless part https://github.com/IntellectualSites/PlotSquared/blob/6e7bd0a5365040f9579ff369f11761fd5834c5b3/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java#L1007-L1024
- Don't cancel water absorption on unowned plots and roads, so staff can use sponges to clear up water there if they so desire.
- Cancel event if no water is going to be absorbed, so the sponge doesn't turn into a wet sponge.

I only tested in a square plot world, I didn't stuff across area borders.

## Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -1191,25 +1191,17 @@ public class BlockEventListener implements Listener {
         PlotArea area = location.getPlotArea();
         List<org.bukkit.block.BlockState> blocks = event.getBlocks();
         if (area == null) {
-            for (int i = blocks.size() - 1; i >= 0; i--) {
-                location = BukkitUtil.adapt(blocks.get(i).getLocation());
-                if (location.isPlotArea()) {
-                    blocks.remove(i);
-                }
-            }
+            blocks.removeIf(block -> BukkitUtil.adapt(block.getLocation()).isPlotArea());
         } else {
             Plot origin = area.getOwnedPlot(location);
-            for (int i = blocks.size() - 1; i >= 0; i--) {
-                location = BukkitUtil.adapt(blocks.get(i).getLocation());
-                if (!area.contains(location.getX(), location.getZ())) {
-                    blocks.remove(i);
-                    continue;
+            blocks.removeIf(block -> {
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):
LGTM

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -1183,4 +1184,37 @@ public void onLeavesDecay(LeavesDecayEvent event) {
...
     }
 
+    @EventHandler(ignoreCancelled = true)
+    public void onSpongeAbsorb(SpongeAbsorbEvent event) {
+        Block sponge = event.getBlock();
+        Location location = BukkitUtil.adapt(sponge.getLocation());
+        PlotArea area = location.getPlotArea();
+        List<org.bukkit.block.BlockState> blocks = event.getBlocks();
+        if (area == null) {
+            for (int i = blocks.size() - 1; i >= 0; i--) {
```
SirYwell, line 1194: Is there any reason why the list is iterated backwards? If not, a simple `removeIf` would be better probably (and if it needs to be reversed, we could probably use the `ListIterator` instead).

traksag: No particular reason. I copied this loop from https://github.com/IntellectualSites/PlotSquared/blob/6e7bd0a5365040f9579ff369f11761fd5834c5b3/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java#L982-L987
Similarly for the backwards loop below. Changing it to `removeIf` sounds good to me.

SirYwell: Yeah I guess you can change it then.

SirYwell (approved):

PR #3192, opened 2021-07-31
Fix emptying buckets in plot border

traksag:
## Overview

You can currently place water in the plot border as follows: place a waterloggable block on the edge of your plot, waterlog it, click on the side of the block connecting to the plot border with a water bucket. The water will be placed inside the plot border.

Similarly, you can also place lava in plot borders. You don't even need to waterlog the block (although the block does need to be waterloggable).

## Description

In the `PlayerBucketEmptyEvent` listener, fix the code that determines where the fluid is going to be placed:
- If the waterloggable block is waterlogged, the water is going to be placed on the side.
- If the bucket contains lava, the lava is going to be placed on the side.

A month after Bukkit updated to 1.14.4, the `PlayerBucketEmptyEvent` got the API method `getBlock()` that tells you where the fluid is going to be placed. Using that API method would make PlotSquared more robust to future Minecraft changes. However, I believe v6 is supposed to support 1.13, so I opted against using that API method.

## Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -1170,7 +1170,8 @@ public class PlayerEventListener extends PlotListener implements Listener {
         final Block block;
         // if the block can be waterlogged, the event might waterlog the block
         // sometimes
-        if (event.getBlockClicked().getBlockData() instanceof Waterlogged) {
+        if (event.getBlockClicked().getBlockData() instanceof Waterlogged waterlogged
+                && !waterlogged.isWaterlogged() && event.getBucket() != Material.LAVA_BUCKET) {
             block = event.getBlockClicked();
         } else {
             block = event.getBlockClicked().getLocation()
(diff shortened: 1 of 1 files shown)


NotMyFault:
Unless I [tested it](https://streamable.com/w0f6ko) improperly, this doesn't do anything but creates more issues?

traksag:
The v6 branch has the same behaviour as in your video. Here's a video of the two issues this pull requests aims to resolve (placing water and lava inside the plot border):

https://user-images.githubusercontent.com/58818927/127784314-31ad87ea-c26c-406b-8cac-af082afb48a8.mov

NotMyFault:
Oic, thanks for the clarification. Looks like I reproduced the description improperly or discovered another weird edge case 👀

NotMyFault (approved):
LGTM, the comment is valid, but we retain 1.13 support for now. Welp, somewhat of. Nobody really uses that version anyway anymore.

SirYwell (commented):
Could you add a comment in the code that `getBlock()` could be used on newer versions? That could be helpful in future.

SirYwell (approved):
Thanks!

PR #3200, opened 2021-08-08
Add option to "update" blocks that are altered in P2 queues.

### LINKED ISSUE #2996: "/plot clear" does not cause block updates
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

When a plot is being cleared, blocks get removed but block updates do not happen - similar to how in WorldEdit you can choose, using "//perf", whether or not to do block updates after operations.

I discovered that this was the case because the walls of my plot are iron bars, and clearing a plot when there are blocks adjacent to them results in them retaining their "attached" block state. These screenshots should describe what I mean:

Before "/plot clear":
image
After "/plot clear":
image

### To Reproduce

- Set up a plot world with walls as iron bars, fences or something like that.
- Place blocks on the inside of the plot but attached to the walls.
- Clear the plot using "/plot clear"

### Expected behaviour

Blocks should update upon clearing plots, so in the case of what the screenshots were showing, the iron bars should have "realised" that there were no longer stone blocks next to them and changed block states accordingly.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/a99742050b2a44948015d61d7de9426f

### PlotSquared Version

5.13.11-Premium

### Checklist

### Anything else?

_No response_

dordsor21:
Fixes #2996


NotMyFault (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
```diff
@@ -271,4 +292,102 @@ private void setWorldBlock(int x, int y, int z, @NonNull BaseBlock block, @NonNu
         return bukkitWorld;
     }
 
+    private boolean isEdge(int layer, int x, int y, int z, BlockVector2 blockVector2, LocalChunk localChunk) {
+        if (layer == 0 || layer == localChunk.getBaseblocks().length - 1) {
+            return false;
+        }
+        if (x == 0) {
+            LocalChunk localChunkX = getBlockChunks().get(blockVector2.withX(blockVector2.getX() - 1));
+            if (localChunkX == null || localChunkX.getBaseblocks()[layer] == null || localChunkX.getBaseblocks()[layer][ChunkUtil.getJ(
```
SirYwell, line 301: This formatting is really odd. Can you wrap on the `||` instead?

SirYwell (dismissed):

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
dordsor21: It's the editorconfig 🙃

SirYwell: Manually doing the line break should work and is a lot more readable

dordsor21: It'll get undone when someone updates code reformats lol

SirYwell (approved):

PR #3201, opened 2021-08-08
Slightly change how flag permissions are handled

dordsor21:
- Some flags should not check for a permission "value", mainly string flags
- Add a default true method for asking a flag if it requires a "valued" permission
- Add "keyed" permissions, where we would only expect certain values from a permission, allowing for the "*" modifier to make sense. E.g. with the music flag.
- Flag permissions without the ".*" make sense to be given to players in the permpack
- Fixes #3151

PR #3202, opened 2021-08-08
Add TitleFlag

dordsor21:
- Allows players to set custom plot titles
- Figured this would be the best implementation..?
- `/plot flag set title "a title" "the subtitle"`
- Added it to list of flags to be sanitised for clickables just in case...
- Should be merged after #3201


Core/src/main/java/com/plotsquared/core/listener/PlotListener.java
```diff
@@ -291,11 +292,22 @@ public boolean plotEntry(final PlotPlayer<?> player, final Plot plot) {
...
-                        .isEmpty()) {
+                String title;
+                String subtitle;
+                String[] titleFlag = plot.getFlag(TitleFlag.class);
+                boolean fromFlag;
+                if (titleFlag.length == 2) {
+                    title = titleFlag[0];
+                    subtitle = titleFlag[1];
+                    System.out.println(title);
+                    System.out.println(subtitle);
```
NotMyFault, line 303: Always helpful to use 'em.

NotMyFault (changes requested):
For sake of consistency, I'd choose a different class name and a parameter like `custom-titles`. Else this would be the first flag you set without any flag name.

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.plotsquared.core.configuration.caption.TranslatableCaption;
+import com.plotsquared.core.plot.flag.FlagParseException;
+import com.plotsquared.core.plot.flag.PlotFlag;
+import com.plotsquared.core.util.ItemUtil;
+import com.sk89q.worldedit.world.item.ItemType;
+import com.sk89q.worldedit.world.item.ItemTypes;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public class TitleFlag extends PlotFlag<String[], TitleFlag> {
```
SirYwell, line 36: This shouldn't be `String[]`. The flag is mutable this way and the type is pretty meaningless in general, in difference to the invariants expected from it.

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
```diff
@@ -0,0 +1,101 @@
...
+     * Construct a new flag instance.
+     *
+     * @param value Flag value
+     */
+    protected TitleFlag(String[] value) {
+        super(value, TranslatableCaption.of("flags.flag_category_string"), TranslatableCaption.of("flags.flag_description_title"));
+    }
+
+    @Override
+    public TitleFlag parse(@NonNull String input) throws FlagParseException {
```
SirYwell, line 50: This should probably accept something like `/plot flag set Hello World`, similar to the greeting flag too.

Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
```diff
@@ -145,9 +146,16 @@ private static boolean checkPermValue(
             } catch (final Exception e) {
                 return false;
             }
-            return true;
+            return true;//
+        }
+        boolean result;
+        String basePerm = Permission.PERMISSION_SET_FLAG_KEY.format(key.toLowerCase());
+        if (flag.isValuedPermission()) {
```
SirYwell, line 153: `valued` or `keyed`? Sounds confusing

Core/src/main/java/com/plotsquared/core/plot/flag/PlotFlag.java
```diff
@@ -158,6 +158,15 @@ public Caption getFlagCategory() {
         return this.flagCategory;
     }
 
+    /**
+     * Get if the flag's permission should check for values. E.g. plot.flag.set.music.VALUE
+     *
+     * @return if valued permission
+     */
+    public boolean isValuedPermission() {
```
SirYwell, line 166: Maybe this should be a more extensible system that e.g. also allows to check for ranges for Long/Double flags?

SirYwell (changes requested):
This flag is currently a title version of the greeting flag. There should be a version for farewell too imo (and therefore the name should be probably something like `greeting-title`)

dordsor21:
> This flag is currently a title version of the greeting flag. There should be a version for farewell too imo (and therefore the name should be probably something like `greeting-title`)

There's no such thing. It's there to allow the pop-up titles when players enter plots to be configurable by the owner. That would be out of scope of this PR.

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
dordsor21: Would that therefore default to only the title being "hello world"? Or should we assume the first "word" is the title, and the second, the subtitle?

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
dordsor21: Immutable list for two strings seems excessive. It could be a string, but every time you want to read it, you'd have to split the String...

Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
dordsor21: See #3201

Core/src/main/java/com/plotsquared/core/plot/flag/PlotFlag.java
dordsor21: See #3201

SirYwell:
I'm okay with another flag being out of scope, but renaming the flag afterwards is a breaking change, therefore I'd like to avoid that.

dordsor21:
> I'm okay with another flag being out of scope, but renaming the flag afterwards is a breaking change, therefore I'd like to avoid that.

It would need the entire system and default values etc also be added for leaving-plot-titles. This would also be very jarring as roads are small, so crossing the road would display two titles very quickly

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
SirYwell: You can create an own type for it

dordsor21: lol, duh, fair enough

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
SirYwell: I'd expect it to be the title only then. Not sure what the others think about it

Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
SirYwell: It checks hasKeyedPermission inside of that if. The other PR doesn't clear that up for me

dordsor21: The comment is better suited to that PR really I mean^

Core/src/main/java/com/plotsquared/core/plot/flag/PlotFlag.java
SirYwell: Our own flags aren't the only ones that exist. 
What I meant is, that it would probably be worth to let the flag check if a value is accepted based on a user context, not necessarily dependent on permissions. While this would make the flags less value-based, it would be far more flexible and future-proof.

dordsor21: The check is there so P2 knows if the permission `plots.flag.set.time.20000` should be looked up (it shouldn't). Similarly `plots.flag.set.greeting.Welcome to my plot!`. Whereas something like `plots.flag.set.music.cat` should be searched, removing the need for the plots.flag.set.greeting.* permission where it really shouldn't be required.

SirYwell:
> It would need the entire system and default values etc also be added for leaving-plot-titles. This would also be very jarring as roads are typically small, so crossing the road would display two titles very quickly

Fair, I don't think we need to implement that at all, but in my opinion, `title` isn't a very meaningful name that additionally *could* be inconsistent in future

dordsor21:
> > It would need the entire system and default values etc also be added for leaving-plot-titles. This would also be very jarring as roads are typically small, so crossing the road would display two titles very quickly
> 
> Fair, I don't think we need to implement that at all, but in my opinion, `title` isn't a very meaningful name that additionally _could_ be inconsistent in future

Most users are aware of what the titles are I think, but I'd also not want to name it something that suggests to users there should be a "leaving" title when I really don't think it's a good idea because of the spam it would definitely vause

Core/src/main/java/com/plotsquared/core/permissions/PermissionHolder.java
```diff
@@ -92,4 +107,18 @@ default int hasPermissionRange(
...
+     * Check if the owner of the profile has a given keyed permission. Checks both {@code permission.key}
+     * and {@code permission.*}
+     *
+     * @param world      World name
+     * @param permission Permission
+     * @param key        Permission "key"
+     * @return {@code true} if the owner has the given permission, else {@code false}
+     */
+    boolean hasKeyedPermission(
+            @Nullable String world, final @NonNull String permission,
```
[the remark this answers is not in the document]
dordsor21: Also part of #3201

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/TitleFlag.java
SirYwell: (we even have records, so that's a one-liner)

SirYwell:
Still not super happy about it, but I'll let the others decide :p

NotMyFault:
> Fair, I don't think we need to implement that at all, but in my opinion, title isn't a very meaningful name that additionally could be inconsistent in future

Titles and Title sound too similar regardless of the actor knows what these mean or not. Unfortunately I can't think of a better name rn, but maybe someone else can come up with something less similar.

NotMyFault (approved):

SirYwell (approved):

PR #3206, opened 2021-08-11
Fix typos

fredster33:
Hopefully this doesn't need the whole issue template, as it's trivial :)


NotMyFault:
It can stay the it is, but thanks for your contribution.

PR #3238, opened 2021-09-08
Some improvements to events

dordsor21:
Will require moving to v6.1.x

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/events/PlayerAutoPlotEvent.java
+++ b/Core/src/main/java/com/plotsquared/core/events/PlayerAutoPlotEvent.java
@@ -27,6 +27,7 @@ package com.plotsquared.core.events;
 
 import com.plotsquared.core.command.Claim;
 import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.Plot;
 import com.plotsquared.core.plot.PlotArea;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
@@ -39,8 +40,8 @@ public class PlayerAutoPlotEvent extends PlotEvent implements CancellablePlotEve
     private final PlotArea plotArea;
     private Result eventResult;
     private String schematic;
-    private int size_x;
-    private int size_z;
+    private int sizeX;
+    private int sizeZ;
 
     /**
      * PlayerAutoPlotEvent: called when a player attempts to auto claim a plot.
@@ -48,19 +49,31 @@ public class PlayerAutoPlotEvent extends PlotEvent implements CancellablePlotEve
      * @param player    The player attempting to auto claim
      * @param plotArea  The applicable plot area
      * @param schematic The schematic defined or null
-     * @param size_x    The size of the auto area
-     * @param size_z    The size of the auto area
+     * @param sizeX     The size of the auto area
+     * @param sizeZ     The size of the auto area
      */
     public PlayerAutoPlotEvent(
             PlotPlayer<?> player, PlotArea plotArea, @Nullable String schematic,
-            int size_x, int size_z
+            int sizeX, int sizeZ
     ) {
         super(null);
         this.player = player;
         this.plotArea = plotArea;
         this.schematic = schematic;
-        this.size_x = size_x;
-        this.size_z = size_z;
+        this.sizeX = sizeX;
+        this.sizeZ = sizeZ;
+    }
+
+    /**
+     * Returns null as the plots to be claimed haven't been chosen yet. This will depend on the size of the auto
+     * ({@link PlayerAutoPlotEvent#setSizeX(int)} and {@link PlayerAutoPlotEvent#setSizeZ(int)}). To see which plots have been
+     * chosen, see {@link PlayerAutoPlotsChosenEvent}.
+     *
+     * @return null
+     */
+    @Override
+    public @Nullable Plot getPlot() {
+        return null;
     }
 
     /**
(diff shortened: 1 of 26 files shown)


Core/src/main/java/com/plotsquared/core/events/PlayerAutoPlotEvent.java
```diff
@@ -99,20 +112,20 @@ public PlotArea getPlotArea() {
...
-    public int getSize_z() {
-        return this.size_z;
+    public int getSizeZ() {
+        return this.sizeZ;
     }
 
-    public void setSize_z(int size_z) {
-        this.size_z = size_z;
+    public void setSizeZ(int sizeZ) {
+        this.sizeZ = sizeZ;
```
NotMyFault, line 128: Javadocs

Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -1666,15 +1666,11 @@ public boolean hasRatings() {
...
-    public boolean claim(final @NonNull PlotPlayer<?> player, boolean teleport, String schematic) {
-        if (!canClaim(player)) {
-            return false;
-        }
-        return claim(player, teleport, schematic, true);
-    }
-
-    public boolean claim(final @NonNull PlotPlayer<?> player, boolean teleport, String schematic, boolean updateDB) {
-
+    public boolean claim(
```
NotMyFault, line 1669: Javadocs

Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
```diff
@@ -130,8 +133,23 @@ public PlayerAutoPlotEvent callAuto(
...
+    }
+
+    public PlotClaimedNotifyEvent callPlotClaimedNotify(Plot plot, boolean auto) {
+        PlotClaimedNotifyEvent event = new PlotClaimedNotifyEvent(plot, auto);
+        callEvent(event);
+        return event;
+    }
+
+    public PlayerTeleportToPlotEvent callTeleport(PlotPlayer<?> player, Location from, Plot plot, TeleportCause cause) {
+        PlayerTeleportToPlotEvent event = new PlayerTeleportToPlotEvent(player, from, plot, cause);
```
NotMyFault, line 152: Javadocs

NotMyFault (changes requested):
Otherwise, LGTM.

Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
dordsor21: I would rather not add any javadocs to this class... This is about as far away from API as we'd want to get and it's pretty obvious what each does to maintainers.

NotMyFault: Indeed, I solely like having javadocs on public methods, but we toss an `@ApiStatus.Internal` on the class to make it more obvious.

SirYwell (changes requested):
While we're still publishing it as snapshot versions, I'd prefer to avoid any unreasonable breaking changes. v6 is alread running on >1000 servers that might depend on the API (a lot of them do so because we fixed a lot of v5 bugs in v6 without backporting them, including limited functionality on Java 15 and above).

Core/src/main/java/com/plotsquared/core/events/PlayerTeleportToPlotEvent.java
```diff
@@ -43,10 +44,21 @@
      * @param player That was teleported
      * @param from   Start location
      * @param plot   Plot to which the player was teleported
+     * @param cause  Why the teleport is being completed
      */
-    public PlayerTeleportToPlotEvent(PlotPlayer<?> player, Location from, Plot plot) {
+    public PlayerTeleportToPlotEvent(PlotPlayer<?> player, Location from, Plot plot, TeleportCause cause) {
```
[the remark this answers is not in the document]
dordsor21: There is no reason anyone should be making this event themselves... But I'll add back the constructor

Core/src/main/java/com/plotsquared/core/plot/Plot.java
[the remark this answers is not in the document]
SirYwell: I guess it would be fine for now to add an overloading method, let the old method delegate to the new one with a default value and deprecate it.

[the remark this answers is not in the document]
SirYwell: I just saw that you did exactly that, lol

Core/src/main/java/com/plotsquared/core/events/TeleportCause.java
```diff
@@ -25,8 +25,39 @@
...
+    COMMAND_VISIT,
+    DEATH,
+    DENIED,
+    KICK,
+    LOGIN,
     PLUGIN,
-    UNKNOWN
+    UNKNOWN;
+
+    public static final class SETS {
```
SirYwell, line 57: Should be `Sets`

Core/src/main/java/com/plotsquared/core/events/TeleportCause.java
```diff
@@ -25,8 +25,39 @@
...
+
+        public static final EnumSet<TeleportCause> COMMAND = EnumSet.range(TeleportCause.COMMAND, TeleportCause.COMMAND_VISIT);
```
SirYwell, line 59: While I doubt anyone will mess with that, the sets should probably be wrapped into unmodifiable sets (and therefore also change the left hand side type)

SirYwell (changes requested):

SirYwell (approved):

NotMyFault (approved):

PR #3308, opened 2021-10-29
Split configuration for teleport on claim/auto

### LINKED ISSUE #2953: Break down "Teleport to your plot on claim"
It is already possible to disable teleporting to the newly created plot on claim. However, disabling teleportation on claim is not as granular as would be useful. The `teleport.on-claim` config option should be broken down into multiple config options for better control of teleporting on claim.

**Why?**

In some cases, the server operator may want to:
* Teleport players when they use `/p auto` to claim (since `/p auto` claims the next available plot, it may be far away from the player, requiring them to be teleported), but:
* Not teleport players when they use `/p claim` (since `/p claim` claims the plot you are standing in, the player doesn't need to be teleported because they are already there, and any teleportation is just disorienting).

Two options:
1. Break the `teleport.on-claim` config option out into `teleport.on-claim.claim` and `teleport.on-claim.auto` so that server administrators can choose what types of claiming to teleport on (e.g. "true" for `teleport.on-claim.auto` and `false` for `teleport.on-claim.claim`, thus resolving the issue outlined above).
2. Make existing `teleport.on-claim` config option not affect plots claimed using `/p auto` (may break functionality on some servers who already have this config option set to `false` for other reasons)

**Additional context**
IntellectualSites/PlotSquaredSuggestions#41 

**Checklist**:
- [x] I made sure there are no duplicates of my suggestion [(Use search)](https://github.com/IntellectualSites/PlotSquaredSuggestions/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+)
- [x] I made sure my suggestion does not exist yet [(Check the wiki)](https://wiki.intellectualsites.com/plotsquared/home)
- [x] I made sure my suggestion is useful for the majority of PlotSquared users
  - Yes, I have seen this requested in other issues (#41) and on Discord, but the current implementation doesn't work for the most common use case for this (a creative plot server)

dordsor21:
- Fixes #2953

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
@@ -556,8 +556,10 @@ public class Settings extends Config {
         public static boolean ON_DEATH = false;
         @Comment("Teleport to your plot on login")
         public static boolean ON_LOGIN = false;
-        @Comment("Teleport to your plot on claim")
+        @Comment("Teleport to your plot on claim (/plot claim)")
         public static boolean ON_CLAIM = true;
+        @Comment("Teleport to your plot on auto (/plot auto)")
+        public static boolean ON_AUTO = true;
         @Comment({"Add a delay to all teleport commands (in seconds)",
                 "Assign `plots.teleport.delay.bypass` to bypass the cooldown"})
         public static int DELAY = 0;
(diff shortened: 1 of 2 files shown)


NotMyFault:
@SirYwell Would you kindly drop in a review?

SirYwell (approved):

PR #3319, opened 2021-11-01
fix for issue 3317 (flag untrusted-visit not working)

### LINKED ISSUE #3317: Flag untrusted-visit not working and superfluous permission
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

There are actually two issues, which are closely related and have the same reason:
1. untrusted-visit does not work anymore (if set to true, untrusted players can't teleport to the plot)
2. the permission plot.admin.visit.untrusted is superfluous, since plot.visit.other has been introduced

The problem is located here: https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/command/Visit.java#L143
plot.visit.other is checked first, and if not given (which is usual for normal players, at least on our servers) the visit is denied, the checks for untrusted-visit and plot.admin.visit.untrusted will not be made. 

Note that for both plot.admin.visit.untrusted and plot.visit.other this is the only location where they are checked, so they obviously have the same usage, so there is no need to have them both. 

### To Reproduce

1. set untrusted-visit to true on an owned plot.
2. try to visit that plot with the `/plot visit` command with a second player that is not trusted on the plot -> visit is denied
3. give the second player the permission plot.admin.visit.untrusted
4. try to visit the plot -> still denied

### Expected behaviour

When untrusted-visit is set, untrusted players should be able to visit the plot. 

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/53f2e12f466148d3b215f33a23fedae4

### PlotSquared Version

tested with 5.13.12-premium, 6.1.3-premium and 6.1.4-snapshot

### Checklist

### Anything else?

The two if clauses in https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/command/Visit.java#L143 and https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/command/Visit.java#L150 can be merged into a single clause, then the flag will work again (I tested this on my local test server, which is also the one who created the debugpaste above. On our online servers, we use the premium version, currently still 5.x for unrelated reasons).
I still have both permissions in my modification, but obviously you don't need both. However, I think it's up to you to decide which one you want to keep (I would keep plot.visit.other).

mindw0rm:
## Overview

Fixes #3317 

## Description

This is my solution as described in #3317. I kept both permissions, but the error message (if the player has neither and untrusted-visit is not set) will show only plots.visit.other. 

## Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Visit.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Visit.java
@@ -140,18 +140,13 @@ public class Visit extends Command {
                 return;
             }
         } else {
-            if (!Permissions.hasPermission(player, Permission.PERMISSION_VISIT_OTHER)) {
+            // allow visit, if UntrustedVisit flag is set, or if the player has either the plot.visit.other or
+            // plot.admin.visit.untrusted permission
+            if (!plot.getFlag(UntrustedVisitFlag.class) && !Permissions.hasPermission(player, Permission.PERMISSION_VISIT_OTHER)
+                && !Permissions.hasPermission(player, Permission.PERMISSION_ADMIN_VISIT_UNTRUSTED)) {
                 player.sendMessage(
                         TranslatableCaption.of("permission.no_permission"),
                         Templates.of("node", "plots.visit.other")
-                );
-                return;
-            }
-            if (!plot.getFlag(UntrustedVisitFlag.class) && !Permissions
-                    .hasPermission(player, Permission.PERMISSION_ADMIN_VISIT_UNTRUSTED)) {
-                player.sendMessage(
-                        TranslatableCaption.of("permission.no_permission"),
-                        Templates.of("node", "plots.admin.visit.untrusted")
                 );
                 return;
             }


NotMyFault:
Considering v6 added a permission that does exactly what the untrusted visit flag does, I'd vouch for pulling this PR yet deprecating named flag in favor of the added permission. Opinions @IntellectualSites/plotsquared-team ?

mindw0rm:
The permission does **not** (and can not) do the same as the flag, because a permission is player specific, and the flag is plot specific. 
If you give someone the permission, they can visit all other (owned) plots, no matter if the plot owner wants this or not.
The flag allows the plot owners to control if they want to allow other to visit or not. 

Please keep the flag.

NotMyFault:
Right, the flag can control individual behavior. Yet that may need to be specified then.

NotMyFault (approved):

dordsor21 (approved):

Issue #3370, opened 2021-12-14
Update to MiniMessage/Adventure Text 4.10.0


NotMyFault:
~~This is a regression of https://github.com/IntellectualSites/PlotSquared/pull/3354 (PlotSquared 6.2.0), which affects following components:~~
- ~~`/plot help`: https://i.imgur.com/db2MCG6.png~~
- ~~`/plot list <all/mine/user>`: https://i.imgur.com/8WL5UTz.png~~
~~Plot listing does now take world names into consideration for tab completion, I never noticed that before, if that is not affected by named regression, ignore it and shuffle it into a different issue.~~
- ~~`/plot setup`: https://i.imgur.com/6LAW7Lc.png~~
- ~~`/plot download`: https://i.imgur.com/8pXzr4L.png~~
- ~~`/plot debugpaste`~~
- ~~`/plot inbox <inbox> clear`: https://i.imgur.com/WtyMaYX.png~~

~~Namely parts with clickable components are broken, others may follow.~~

Edit: The PR that introduced the breakage has been reverted, yet we should update MiniMessage to 4.10.0(-SNAPSHOT) in a timely manner, to not lack fixes, features again without a deprecation phase..

PR #3354, opened 2021-12-05
refactor: Address MiniMessage deprecations

NotMyFault:
## Description
According to OSS Sonatype, 4.1.0-SNAPSHOT was last updated back in July 2021, either renovate missed it or there was no release of 4.2.0. However, there are over 600 deprecation & "scheduled for removal" warnings we should address in a timely manner.

## Checklist

- [X] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [X] Ensure that the pull request title represents the desired changelog entry.
- [X] I tested my changes and approved their functionality.
- [X] I ensured my changes do not break other parts of the code
- [X] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md)

PR #3372, opened 2021-12-15
Update MiniMessage properly to 4.10

NotMyFault:
## Overview
Fixes https://github.com/IntellectualSites/PlotSquared/issues/3370

## Description
This PR addresses the reverted PR https://github.com/IntellectualSites/PlotSquared/pull/3354 properly, after it introduced some breaking regressions.

Additionally I dropped a few getter and setters on public methods in the caption package, in favor of record accessors, which we may use in the future.

Developer notes need to be proposed in the changelog, guiding API dependents into the right direction how to update properly.

## Checklist

- [ ] I tested my changes and approved their functionality.
- [ ] I ensured my changes do not break other parts of the code - skipped, Developer notes are going to be proposed
- [X] Developer notes for the changelog have been proposed


PierreSchwang (approved):
LGTM - Haven't found any broken messages during my tests.

SirYwell:
Definitely not a fan of this as long as it's targetting v6. We can't pretend to be using semver while we actually don't care about it.

As 4.10 isn't even released yet, this has to wait and we then need a way to get this done without breaking API.

NotMyFault:
> Definitely not a fan of this as long as it's targetting v6. We can't pretend to be using semver while we actually don't care about it.
> 
> As 4.10 isn't even released yet, this has to wait and we then need a way to get this done without breaking API.

MiniMessage/Adventure didn't bump their major version in the first place when moving into adventure and not retaining backwards compatibility. We don't know when PlotSquared v7 is released to GA, considering it's not something that is even in the plannings.
Adventure is a great and well maintained library but our current version is already EOL for more than 6 months. 
We're basically stuck on a version that is abandoned, if you insist to not update it.

I see no point in providing backwards compatible methods, especially if adventure didn't do so either.
I mean, it's nice to have but for whom would that be actually relevant, if you can update PlotSquared and your addons with ease.
The changes proposed are not difficult to adapt, they are solely a naming convention change. Change one method and tell your IDE to do a refactor and you are done. In fact, I kept deprecations of our methods. You can still call a `TranslatableCaption#of`, `StaticCaption#of`, `Templates#of` etc. just with the note to use the new ones that adapt the naming convention. The only things that are actually breaking are `TranslatableCaptions` in combination with Placeholders and message builders in combination with PlaceholderResolver, but if we propose a proper documentation how to adapt the changes, people can do so with ease. 

Though to note, you are right, this is a major breakage that violates semver, but in my opinion something worth to do, to switch to a maintained version of one our core components.

NotMyFault:
Would like to hear some other opinions about it, maybe also from the people using the API for captions.

dordsor21:
> Would like to hear some other opinions about it, maybe also from the people using the API for captions.

Using P2 for captions is quite painful as you have to relocate kyori yourself. Accessing caption API externally is not under the P2 classpath

NotMyFault:
> > Would like to hear some other opinions about it, maybe also from the people using the API for captions.
> 
> Using P2 for captions is quite painful as you have to relocate kyori yourself. Accessing caption API externally is not under the P2 classpath

Ya, that'd be a different concern then. But I've seen some people doing so.

dordsor21:
We ought not be using snapshot versions of libraries tbh, so updating yet isn't a good idea. But I'm not really a fan of having to bump the major P2 version when a library we shade does major changes either...

NotMyFault:
> We ought not be using snapshot versions of libraries tbh

Our current used version is a snapshot too, and so were (almost but Hyperverse, rorledning and aopalliance) all of our third party libraries snapshot builds until I started deploying them to the central repository or exchanged them with stable releases where applicable.

SirYwell:
You're required to use P2 captions for custom flags for example. It is *meant* to be used. You don't necessarily need to use any of adventure for *that* part. But the fact that adventure isn't relocated is a problem in itself, as we're exposing it.

I didn't come up with a good solution for now how to deal with mini message, but we either keep to semver or we don't - and that also affects the third part API we expose. The only real solutions I came up with:
We can maintain our own mini message fork, restoring removed methods/classes (fair amount of work), not update mini message (not great either - in terms of features. bug fixes etc), do weird binary hacking stuff (no, we don't want to do that) or bump our major version (mh, well).

What we can learn from this is that we shouldn't depend on anything that only has snapshot releases in our API. And especially as mini message is still in that stage, I'd at least wait until they get their mess done before breaking and breaking again.

dordsor21:
What if we added a layer of abstraction, allowing for the simple stuff to be done via P2-classpath-located Template and Caption Classes (named differently), but also allowed the minimessage objects to be obtained/used as required, then (not great but...) simply ignore the breaking changes from kyori libraries.

NotMyFault:
> I didn't come up with a good solution for now how to deal with mini message, but we either keep to semver or we don't - and that also affects the third part API we expose.

At this point I would rather say no and do a semi breaking release under a minor.

>The only real solutions I came up with:
We can maintain our own mini message fork, restoring removed methods/classes (fair amount of work), not update mini message (not great either - in terms of features. bug fixes etc), do weird binary hacking stuff (no, we don't want to do that) or bump our major version (mh, well).

We can spin up a `Libs` module for core components that can be consumed by third parties, but are provided at runtime so third parties don't need to do an empty relocation (see HoloPlots e.g.) or squash all their work under our classpath.

> What if we added a layer of abstraction, allowing for the simple stuff to be done via P2-classpath-located Template and Caption Classes (named differently),

What do you suggest?

dordsor21:
Abstracted<name> maybe? e.g. AbstractedTemplate. Then just add the methods (old and/or new) of #template and #of, both creating the MM Template, storing it inside the AbstractedTemplate and returning a new AbstractedTemplate (also keep the constructor private//package-private)

NotMyFault:
> Abstracted maybe? e.g. AbstractedTemplate. Then just add the methods (old and/or new) of #template and #of, both creating the MM Template, storing it inside the AbstractedTemplate and returning a new AbstractedTemplate (also keep the constructor private//package-private)

I'm unsure how that differs from my current approach but makes use of a generic and agnostic class name, if that is the goal?
I still kept the legacy names in the deprecated legacy classes.

dordsor21:
The current method returns the MM classes, not our own. Transitioning to returning a wrapped MM object, allowing for the base functionality required in 90% of cases would mean interaction with MM-classpath stuff is limited, but can still be down where required.

NotMyFault:
> The current method returns the MM classes, not our own. Transitioning to returning a wrapped MM object, allowing for the base functionality required in 90% of cases would mean interaction with MM-classpath stuff is limited, but can still be down where required.

Oic, indeed that can reduce it. Do you mind raising a PR or branch against this branch so we can move topic discussion for that specific approach on that PR? Can also just commit here, idm.

dordsor21:
Probably best to raise a new PR, or force push to this one. It would be quite a change as we'd likely want to use the same internally as well for consistency

NotMyFault:
Closing this for now, we'll revisit this in v7 when I can finally break things again 💥

PR #3373, opened 2021-12-15
Fix more MM builder issues

PierreSchwang:
## Overview
Follows #3372

## Description
- Fixes issue with unpairing of signed plot coordinates
- Fixes issues with MM and Component builders
- Fixes mistake in translation file

Everything I could think of - Maybe some other commands are still broken, haven't found any other yet

## Checklist


Core/src/main/resources/lang/messages_en.json
```diff
@@ -504,7 +504,7 @@
   "flag.flag_info_name": "<gray>Name: <gold><flag></gold></gray>",
   "flag.flag_info_category": "<gray>Category: </gray><gold><value></gold>",
   "flag.flag_info_description": "<gray>Description: </gray>",
-  "flag.flag_info_example": "<gray>Example: <click:suggest_command:<command> <flag> <value>><gold><command> <flag> <value></click></gold></gray>",
+  "flag.flag_info_example": "<gray>Example: <click:suggest_command:<command> <flag> <value>><gold><command> <flag> <value></gold></click></gray>",
```
NotMyFault, line 507: This change leads to 

Unless this is a different caption I'm looking at.

Core/src/main/java/com/plotsquared/core/plot/PlotId.java
```diff
@@ -110,7 +110,7 @@ private PlotId(final int x, final int y) {
      * @return Plot ID
      */
     public static @NonNull PlotId unpair(final int hash) {
-        return PlotId.of(hash >> 16, hash & 0xFFFF);
+        return PlotId.of(hash >> 16, (short) (hash & 0xFFFF));
```
NotMyFault, line 113: Preferably deprecate the existing method and create a new one named `shortUnpair` or something that actually casts to a short.

NotMyFault (changes requested):
nit: From a UI perspective, the line breaks at the info and inbox kinda disrupt the reading flow



Additionally, not sure whether if it was introduced in this PR or part of mine, enabling `titles-as-actionbar` and setting the flag `plot-title` only sends the first part of the flag to the player, e.g if you parse "heyo" "user", only the heyo pops up.

Running `/plot f set` returns

Core/src/main/resources/lang/messages_en.json
PierreSchwang: The change does fix exactly that behavior your screenshot is showing. you need to update that in the PlotSquared/lang/messages_en.json as well

Core/src/main/java/com/plotsquared/core/plot/PlotId.java
SirYwell: Fixing it like that is fine I guess, because it's a bug if it didn't work correctly, but I'd like to have more insights on under which circumstances this was discovered (and especially why it wasn't an issue before)

PierreSchwang: Found in the inbox list
grafik

Debugged the bit shifting etc and found out, that signed / negative coordinates are not correctly unpaired, because they exceed the bound for signed shorts. Therefor it is a bug. The method is only used in inboxes, maybe they are not used that often

Core/src/main/java/com/plotsquared/core/util/helpmenu/HelpPage.java
```diff
@@ -68,7 +68,7 @@ public void render(PlotPlayer<?> player) {
                             )
                     )
             );
-            Placeholder<?> help_objects = Placeholder.miniMessage("help_objects", StringMan.join(this.helpObjects, "\n"));
```
NotMyFault, line 71: The help menu looks even more cluttered now

We can safely get a rid of all line breaks.

PierreSchwang: Removed the other line breaks as well

Core/src/main/java/com/plotsquared/core/plot/PlotId.java
PierreSchwang: > Preferably deprecate the existing method and create a new one named shortUnpair or something that actually casts to a short.

Based on @SirYwell's comment and the intention of the method I would not create a new method for that. The current implementation is simply broken and shouldn't be extended with another one for the sake of fixing a fundamentally bug. 
The method is only used internally for the inbox. If other plugins are using this method, they will not be broken, as the hash generation is correct, only the unpairing for displaying the hash returns wrong data.

NotMyFault (approved):
Thanks for the fix, looks the help menu looks better

PR #3399, opened 2021-12-25
fix: npe thrown in entity spawn listener

yannicklamprecht:
## Overview
A NPE is thrown when spawning entities.

Error: https://gist.github.com/yannicklamprecht/062d467138b52ea8c8a4a3ab2401f63f

1. Build a PlotSquared version with this commit or later https://github.com/IntellectualSites/PlotSquared/commit/af2613202d05d1069d4416b53c50018dce2a6752
2. Start up a Paper server and spawn an entity
3. It will throw a NPE on line https://github.com/IntellectualSites/PlotSquared/commit/af2613202d05d1069d4416b53c50018dce2a6752#diff-88a5d02f8bdf730223f88142fcc0147612cf7b80f597d3e15e600c9862cd90bcR78

Background info: areaName is not set at that point in time, it will be set after that condition.

PS: Yeah I'm a little to lazy to create an issue. <3

## Description

It will fix a NPE introduced some commit before.

## Submitter Checklist

## Maintainer checklist

- [x] Changelog entries in the PR title are correct
- [x] Proper changelog labels are set so that the changelog can be generated automatically

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntitySpawnListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntitySpawnListener.java
@@ -75,8 +75,7 @@ public class EntitySpawnListener implements Listener {
 
     public static void testCreate(final Entity entity) {
         @NonNull World world = entity.getWorld();
-        if (areaName.equals(world.getName())) {
-        } else {
+        if (!world.getName().equals(areaName)) {
             areaName = world.getName();
             hasPlotArea = PlotSquared.get().getPlotAreaManager().hasPlotArea(areaName);
         }


NotMyFault (approved):
Good catch, thanks for your contribution!

PierreSchwang (approved):
Thank you 🤗

Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntitySpawnListener.java
```diff
@@ -75,8 +75,7 @@ public static void testNether(final Entity entity) {
 
     public static void testCreate(final Entity entity) {
         @NonNull World world = entity.getWorld();
-        if (areaName.equals(world.getName())) {
```
Citymonstret, line 78: Should probably use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object- here

yannicklamprecht: I think I've a state where  Objects.equals doesn't work.

Think of areaName != null  but World#getName() is null. -> that results in a not equals.
-> it enters the if branch.

`
areaName = world.getName(); // will be null then.
hasPlotArea = PlotSquared.get().getPlotAreaManager().hasPlotArea(areaName); // will throw an exception because the parameter is marked as nonNull.

I assume that we don't want that case to be happen? Did I miss something?


`

SirYwell: According to [the documentation](https://papermc.io/javadocs/paper/1.18/org/bukkit/generator/WorldInfo.html#getName()), `World#getName()` will never return `null`. So just using `world.getName().equals(areaName)` should be sufficient too, with no null-check needed

yannicklamprecht: See 0869565

SirYwell (approved):

NotMyFault:
We may merge it in 24 hours if there is no negative feedback, as per the [merge process](https://github.com/IntellectualSites/IntellectualSites/blob/main/docs/MAINTAINERS_CONTRIBUTORS.adoc#merge-process) documentation.

PR #3419, opened 2021-12-30
Fix: use ownersInclude instead of ownedBy where required

### LINKED ISSUE #3143: Merged plots visit different than v5
### Server Implementation

Paper

### Server Version

1.17.1

### Describe the bug

You can only teleport to one of two owner of the plots and not both.

### To Reproduce

1. Merge a plot with different owner, that a plot has two owner
2. /p i 
3. Try to /p v to the second owner
4. Found no plots in your search query
I thought in v5 that I can teleport to the merged plot with both of the owner in /p v. 

### Expected behaviour

That I can teleport with both names to the merged plots like in v5 because the player has a plot, but the message says he doesnt have one.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/2a586f634a814e319d96771bfb492ec8

### PlotSquared Version

6.0.5 Premium

### Checklist

### Anything else?

_No response_

dordsor21:
- Also account for multiple plot owners in Cluster player removal
- Add whereBasePlot to avoid merged plots listing multiple times if required
- Only use ownersInclude in visit if enabled in config
- Fixes #3143

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Visit.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Visit.java
@@ -56,6 +56,8 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeoutException;
 
+import static com.plotsquared.core.configuration.Settings.Teleport.VISIT_MERGED_OWNERS;
+
 @CommandDeclaration(command = "visit",
         permission = "plots.visit",
         usage = "/plot visit <player> | <alias> | <plot> [area]|[#] [#]",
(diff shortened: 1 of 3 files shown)


Core/src/main/java/com/plotsquared/core/command/Visit.java
```diff
@@ -235,8 +237,8 @@ private void visit(
                         } else {
                             final UUID uuid = uuids.toArray(new UUID[0])[0];
                             PlotQuery query = PlotQuery.newQuery();
-                            if (Settings.Teleport.VISIT_MERGED_OWNERS) {
```
NotMyFault, line 238: We use that format everywhere, though it's redundant but makes reading easier.

dordsor21: Was just to make the conditional operators shorter in the if else statements tbh

NotMyFault (approved):

SirYwell (approved):

NotMyFault:
Quick fix, we may merge it in ~24 hours if there is no negative feedback, as per the [merge process](https://github.com/IntellectualSites/IntellectualSites/blob/main/docs/MAINTAINERS_CONTRIBUTORS.adoc#merge-process) documentation.

Issue #3006, opened 2021-05-06
Problem with auto-delete plots


Espantonius:
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

Without touching anything or deleting anything from files, it doesn't delete the plots that have passed X days.

This is my auto-delete plots configuration: 

```
auto-clear:
  task1:
    threshold: 1
    required-plots: -1
    confirmation: true
    days: 90
    worlds: 
    - "*"
    calibration:
      variety: 0
      variety-sd: 0
      changes: 0
      changes-sd: 1
      faces: 0
      faces-sd: 0
      data-sd: 0
      air: 0
      air-sd: 0
      data: 0
```

I have enabled the option of enabled-components.plot-expiry: true 

### To Reproduce

1. Activate and configure the days to expire the plot.
2. Claim plot and wait for the configured days.
3. View the plot information. 
4. After the limit of days it doesn't delete it.

### Expected behaviour

That the plots are deleted after the configured days.

### Screenshots / Videos

This is a plot that has expired but has not been deleted: https://prnt.sc/12kmbr7
Data player: https://prnt.sc/12kmg6f

### Error log (if applicable)

none

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/fcda7c08d5674374860ebe16328ca469

### PlotSquared Version

PlotSquared version 5.13.11

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

I have previously reported this problem when deleting user information.
I haven't deleted anything for a few days and it's still the same.
The plots continue to count days of disconnection but don't delete them. 

I've tried many settings and none work, even the one on the wiki: https://github.com/IntellectualSites/PlotSquared/wiki/World-reduction#plot-condensation-added-in-275-intensive

I have also tried to change the configuration seen in this area of the wiki https://github.com/IntellectualSites/PlotSquared/wiki/Plot-analysis but when start the server again it restores by default and removes some options.

Espantonius:
I have been testing and modifying things but i have not yet managed to delete the plots.
I have seen that it doesn't list the plots that have expired either: https://prnt.sc/12ufdw8

NotMyFault:
Based on the information provided, I experienced similar. `days` were set to 7 and `/plot info`'s `Seen` revealed 1w 2d and 9h.
[...]

PR #3449, opened 2022-01-11
Fix: Missing Component returns + plot info on unknown plot owner

### LINKED ISSUE #3443: Unknown plots not responding to any commands
### Server Implementation

Paper

### Server Version

1.17.1

### Describe the bug

I have a strange problem with an old plotworld (many years old), where hundreds of plots on the world are empty, in claimed state and have 'Owner: unknown'.

If I type commands such as `/plot info` while standing on the plot, nothing happens, no error in the console. If I type a command like `/plot claim` while standing on a broken plot, it will say "This plot is already claimed".

I have tried running the plugin with `uuid-cache-size: 0` and `uuid-cache-size: 100000`

I also tried disabling all UUID services except luckperms (in case one of them was giving bad data?):
``` 
  # Whether the PaperMC service is enabled
  service-paper: false
  # Whether the LuckPerms service is enabled
  service-luckperms: true
  # Whether the Bukkit service is enabled
  service-bukkit: false
  # Whether the EssentialsX service is enabled
  service-essentialsx: false
```

### To Reproduce

Not exactly sure how to suggest reproducing this. I'm happy to assist in trying different fixes on the live server.

Here are the database files:
**storage.db**
https://www.dropbox.com/s/w6yzi07hcbz3plc/storage.db?dl=0
**user_cache.db**
https://www.dropbox.com/s/rrku5ymlv4tgr4b/user_cache.db?dl=0

### Expected behaviour

The plot should recognize it's owner and be loaded in properly.

Most of these plots are empty and by inactive players, so this bug is stopping them from being purged for inactivity.

### Screenshots / Videos

image

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/9d6f87b791a34424bea841ccca6812d3

### PlotSquared Version

PlotSquared version 6.2.2-Premium

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview

Fixes #3443 

## Description
d3fe1d3b2bab02a7e1ef66d3c69d0cfa49d65f26 introduced an issue now known as https://github.com/IntellectualSites/PlotSquared/issues/3443
That was due to a incorrect locale, which should've thrown an exception, but that happened in another thread and just halted the whole thread.
That happened because of missing player data in bukkits storage, documentation from OfflinePlayer#getLastSeen:
> Gets the last date and time that this player was seen on the server.
> If the player has never played before, this will return 0.

That would also explain https://github.com/IntellectualSites/PlotSquared/issues/3006, as the mentioned method is used to determine the last online time of all plot owners. if that is equal to 0, nothing will be deleted: https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/plot/expiration/ExpireManager.java#L229. That should be changed.

## Submitter Checklist

## Maintainer checklist

- [x] Changelog entries in the PR title are correct
- [x] Proper changelog labels are set so that the changelog can be generated automatically

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
+++ b/Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
@@ -164,25 +164,51 @@ public abstract class PlayerManager<P extends PlotPlayer<? extends T>, T> {
      * @return The player's name, None, Everyone or Unknown
      */
     public static @NonNull String getName(final @Nullable UUID owner) {
-        return getName(owner, true);
+        return getName(owner, LocaleHolder.console());
+    }
+
+    /**
+     * Get the name from a UUID.
+     *
+     * @param owner        Owner UUID
+     * @param localeHolder The locale holder to use as the translation provider
+     * @return The player's name, None, Everyone or Unknown
+     */
+    public static @NonNull String getName(final @Nullable UUID owner, final @NonNull LocaleHolder localeHolder) {
+        return getName(owner, localeHolder, true);
     }
 
     /**
      * Get the name from a UUID.
      *
      * @param owner    Owner UUID
-     * @param blocking Whether or not the operation can be blocking
+     * @param blocking Whether the operation can be blocking
      * @return The player's name, None, Everyone or Unknown
      */
     public static @NonNull String getName(final @Nullable UUID owner, final boolean blocking) {
+        return getName(owner, LocaleHolder.console(), blocking);
+    }
+
+    /**
+     * Get the name from a UUID.
+     *
+     * @param owner        Owner UUID
+     * @param localeHolder The locale holder to use as the translation provider
+     * @param blocking     Whether the operation can be blocking
+     * @return The player's name, None, Everyone or Unknown
+     */
+    public static @NonNull String getName(
+            final @Nullable UUID owner, final @NonNull LocaleHolder localeHolder,
+            final boolean blocking
+    ) {
         if (owner == null) {
-            TranslatableCaption.of("info.none");
+            return TranslatableCaption.of("info.none").getComponent(localeHolder);
         }
         if (owner.equals(DBFunc.EVERYONE)) {
-            TranslatableCaption.of("info.everyone");
+            return TranslatableCaption.of("info.everyone").getComponent(localeHolder);
         }
         if (owner.equals(DBFunc.SERVER)) {
-            TranslatableCaption.of("info.server");
+            return TranslatableCaption.of("info.server").getComponent(localeHolder);
         }
         final String name;
         if (blocking) {
@@ -198,7 +224,7 @@ public abstract class PlayerManager<P extends PlotPlayer<? extends T>, T> {
             }
         }
         if (name == null) {
-            TranslatableCaption.of("info.unknown");
+            return TranslatableCaption.of("info.unknown").getComponent(localeHolder);
         }
         return name;
     }
--- a/Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
+++ b/Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
@@ -27,7 +27,9 @@ package com.plotsquared.core.util;
 
 import com.plotsquared.core.PlotSquared;
 import com.plotsquared.core.configuration.Settings;
+import com.plotsquared.core.configuration.caption.Caption;
 import com.plotsquared.core.configuration.caption.LocaleHolder;
+import com.plotsquared.core.configuration.caption.StaticCaption;
 import com.plotsquared.core.configuration.caption.TranslatableCaption;
 import com.plotsquared.core.database.DBFunc;
 import com.plotsquared.core.player.ConsolePlayer;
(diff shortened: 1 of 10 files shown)


Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
```diff
@@ -164,25 +164,51 @@ public static void getUUIDsFromString(
      * @return The player's name, None, Everyone or Unknown
      */
     public static @NonNull String getName(final @Nullable UUID owner) {
-        return getName(owner, true);
+        return getName(owner, LocaleHolder.console());
```
dordsor21, line 167: Might be best to default to null here and deprecate the method, or state that no messages will be sent in the JD comments above. If someone is accessing the method from another plugin and suddenly start getting console output from it, it'll be a little confusing

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
```diff
@@ -164,25 +164,51 @@ public static void getUUIDsFromString(
...
     /**
      * Get the name from a UUID.
      *
      * @param owner    Owner UUID
-     * @param blocking Whether or not the operation can be blocking
+     * @param blocking Whether the operation can be blocking
      * @return The player's name, None, Everyone or Unknown
      */
     public static @NonNull String getName(final @Nullable UUID owner, final boolean blocking) {
+        return getName(owner, LocaleHolder.console(), blocking);
```
dordsor21, line 189: Same as above

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
```diff
@@ -164,25 +164,51 @@ public static void getUUIDsFromString(
...
+     * @param blocking     Whether the operation can be blocking
+     * @return The player's name, None, Everyone or Unknown
+     */
+    public static @NonNull String getName(
+            final @Nullable UUID owner, final @NonNull LocaleHolder localeHolder,
+            final boolean blocking
+    ) {
         if (owner == null) {
-            TranslatableCaption.of("info.none");
+            return TranslatableCaption.of("info.none").getComponent(localeHolder);
```
dordsor21, line 205: Rather than returning the error string, might be better just to send the error message directly to the player, and adjust this method and its callees by returning null or empty? The previous behaviour was returning null after all

SirYwell: The previous behaviour basically ignored the `@NonNull` contract and was wrong therefore. Changing a method from `@NonNull` to `@Nullable` changes its contract in an unsafe manner and should be avoided.

Directly sending the message from this method should be avoided, because that method can be used to show the name in separate places: embedded in different messages, sent as action bar, etc.

That said, the method should probably be deprecated, with a new method returning a `Caption`. This way, the `LocaleHolder` is not required here. WDYT?

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
NotMyFault: Yeah, deprecating the current methods and introduce new ones are the better way to go.

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
PierreSchwang: I mean, the contract is already not fullfilled. Until now, `return name` is called in either case, even if name is null.

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
```diff
@@ -203,6 +209,57 @@ public static void getUUIDsFromString(
...
 
+    /**
+     * Attempts to resolve the username by an uuid
+     * <p>
+     * <b>Note:</b> blocks the thread until the name was resolved or failed
+     *
+     * @param owner The UUID of the owner
+     * @return A caption containing either the name, {@code None}, {@code Everyone} or {@code Unknown}
+     * @see #resolveName(UUID, boolean)
+     * @since 6.2.3
```
NotMyFault, line 220: That'd be 6.3.0 then, considering this is hardly a patch for 6.2. But you can also use `@since TODO`, I have a script that injects the parsed version from `build.gradle.kts` if I publish releases.

NotMyFault (approved):

SirYwell (approved):

NotMyFault:

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
```diff
@@ -173,7 +177,9 @@ public static void getUUIDsFromString(
      * @param owner    Owner UUID
      * @param blocking Whether or not the operation can be blocking
      * @return The player's name, None, Everyone or Unknown
+     * @deprecated Use {@link #resolveName(UUID, boolean)}
      */
+    @Deprecated(forRemoval = true, since = "TODO")
```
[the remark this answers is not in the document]
NotMyFault: Is there something wrong with the TODO here?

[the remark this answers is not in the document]
PierreSchwang: It's a post-merge thing https://github.com/IntellectualSites/PlotSquared/pull/3449#discussion_r783457457

dordsor21 (approved):

Core/src/main/java/com/plotsquared/core/util/PlayerManager.java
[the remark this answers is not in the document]
NotMyFault: > Can/should it not be replaced with the version? Or is that a post-merge thing

Yeah, I replace `@since TODO` with the version released. Using the current version can lead to conflicts or doesn't display the appropriate version, e.g. if this TODO is 1.2.1 but another PR merged would bump to 1.3.0, we would have to address the 1.2.1 ones post merge, do it on the other PR or let the PR author do it; generally introduces more work and more to take core of, instead of simply replacing TODOs when doing releases.

PR #3463, opened 2022-01-20
PlayerInteractEvent does not necessarily interact with a block

dordsor21:


NotMyFault:
Does that address an open issue or was that more a "spot on sight"? Can't think of a scenario to replicate that atm out of my mind.

dordsor21:
Was being spammed into someone's console, and the event specifically marks the block as being nullable

NotMyFault (approved):

SirYwell (approved):

NotMyFault:

Issue #3175, opened 2021-07-22
Replace Y 0, 255/256 with the appropriate API methods


NotMyFault:
Actually, this is some kind of backburner, but nobody really bothered with using the proper methods back the years, because there was no need/demand to. However, using PlotSquared with datapacks or "simple" extended heights leads to it not working properly anymore.

PR #3473, opened 2022-01-25
Implement extended world heights from Y-64 to Y319

dordsor21:
Generation works fine. Everything seems to work fine except where FAWE requires updating.

Fixes https://github.com/IntellectualSites/PlotSquared/issues/3175


Core/src/main/java/com/plotsquared/core/PlotPlatform.java
```diff
@@ -98,6 +98,20 @@
...
+     *
+     * @return minimum world height
+     */
+    int versionMinHeight();
+
+    /**
+     * Gets the default maximum world height for the version of Minecraft that the server is running.
+     *
+     * @return maximum world height
+     */
```
NotMyFault, line 112: `@since TODO`

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -1100,6 +1100,18 @@ public boolean setupPlotWorld(
...
+                                    base + "world.min_gen_height",
+                                    ConfigurationUtil.INTEGER.parseString(value).shortValue()
+                            );
+                        }
+                        case "maxh", "maxheight" -> {
+                            this.worldConfiguration.set(
+                                    base + "world.max_gen_height",
+                                    ConfigurationUtil.INTEGER.parseString(value).shortValue()
+                            );
+                        }
```
NotMyFault, line 1114: Enhanced switches' rules don't necessary need to be labeled anymore.

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -1020,8 +1020,8 @@ public boolean setupPlotWorld(
             // save configuration
 
             final List<String> validArguments = Arrays
-                    .asList("s=", "size=", "g=", "gap=", "h=", "height=", "f=", "floor=", "m=", "main=",
-                            "w=", "wall=", "b=", "border="
+                    .asList("s=", "size=", "g=", "gap=", "h=", "height=", "minh=", "minheight=", "maxh=", "maxheight=",
+                            "f=", "floor=", "m=", "main=", "w=", "wall=", "b=", "border="
```
NotMyFault, line 1024: Is that still in use? Every time I see people using these builders its down to errors. Our plot setup builder is way more robust 🤔

Core/src/main/java/com/plotsquared/core/queue/LocalChunk.java
```diff
@@ -68,6 +71,13 @@ public int getZ() {
         return this.z;
     }
 
+    /**
+     * Get the minimum layer position stored (usually -4 or 0).
+     */
```
NotMyFault, line 76: since

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
```diff
@@ -75,12 +75,25 @@ public QueueCoordinator(@Nullable World world) {
...
     public ScopedQueueCoordinator getForChunk(int x, int z) {
+        return getForChunk(x, z, 0, 255);
+    }
+
+    /**
+     * Get a {@link ScopedQueueCoordinator} limited to the chunk at the specific chunk Coordinates
+     *
+     * @param x chunk x coordinate
+     * @param z chunk z coordinate
+     * @return a new {@link ScopedQueueCoordinator}
```
NotMyFault, line 90: since

Core/src/main/java/com/plotsquared/core/util/RegionManager.java
```diff
@@ -373,28 +379,39 @@ private void setCopyFromToConsumer(
...
             final CuboidRegion region,
             final int extendBiome,
             final BiomeType biome,
             final String world,
             final Runnable whenDone
+    ) {
+        setBiome(region, extendBiome, biome, PlotSquared.get().getPlotAreaManager().getPlotAreas(world, region)[0], whenDone);
+    }
+
+    public void setBiome(
```
NotMyFault, line 393: javadocs and since

Core/src/main/java/com/plotsquared/core/util/WorldUtil.java
```diff
@@ -81,11 +82,28 @@
...
 
+    /**
+     * Set the biome in a region
+     *
+     * @param area  Plot area
+     * @param p1x   Min X
+     * @param p1z   Min Z
+     * @param p2x   Max X
+     * @param p2z   Max Z
+     * @param biome Biome
```
NotMyFault, line 98: since

Core/src/main/resources/lang/messages_en.json
```diff
@@ -161,8 +161,9 @@
   "setup.road_width": "<gold>Road width</gold>",
   "setup.road_height": "<gold>Road height</gold>",
   "setup.road_block": "<gold>Road block</gold>",
-  "setup.wall_filling_block": "<gold>Wall filling block</gold><gray> (Block(s) been put under the plot border down to Y1)</gray>",
+  "setup.wall_filling_block": "<gold>Wall filling block</gold><gray> (Block(s) been put under the plot border down to min gen height + 1)</gray>",
   "setup.wall_height": "<gold>Wall height</gold>",
+  "setup.min_gen_height": "<gold>Minimum height from which to generate (for 1.18+ can be negative).</gold>",
```
NotMyFault, line 166: ...can be negative down to -63 or something? Would be good to include it I guess.

NotMyFault (changes requested):
Looking forward to give it a functional test later

Core/src/main/java/com/plotsquared/core/PlotSquared.java
dordsor21: Doesn't mean it should go in there to maybe be fixed

Core/src/main/resources/lang/messages_en.json
dordsor21: Well technically the worlds can be larger than that in 1.18, and P2 should support it fine

NotMyFault:
Currently testing with Fawe 75 on a newly generated world:
Testing env: https://athion.net/ISPaster/paste/view/acfd539f6c67439bbccf8cd04a9cc659

- [ ] Clearing, deleting, swapping and moving plots does not touch stuff higher than Y256
- [ ] Make sure to get the sout out of #setMiddle before merging :P
- [ ] `/plot set biome savanna`, guess any biome works, followed by `/plot set all air` removes the plot wall, border and floor, when it is supposed to remove blocks inside the plot only.
- [ ] Backups don't work atm. If you clear a plot, a backup schem is created you can load with FAWE (which works fine), but p2 has a hard time: Here's my testing schematic https://keybase.pub/notmyfault_og/plotsquaredTesting/plotworld-1%3B2-1643209005992.schem
- [ ] `/plot crs` -> `/plot debugroadregen plot` pastes the road schematic at Y126, when the specified road height is 62
The same applies to unmerging a plot. The schematic is pasted high above the road as well.
- [ ] Not sure if that is a new introduced issue, but `/plot regenallroads` fires an IllegalStateException for every unloaded chunk touched: https://pastebin.com/raw/kuFf3sQT
More below but I cut it off.
- [ ] `/plot area single` in a normal overworld at Y66 throws https://pastebin.com/raw/1bGPQN9n

Testing with Fawe 76 on an upgraded world from 6.3.0, where I changed `min_gen_height` to -30:
- [ ] `/plot set all air` followed by `/plot clear` results in the same issue outlined in the first bullet point above, however, the bedrock layer is added at Y0 while content below that line remains untouched

Mane of the above described issues apply to the bottom part as well, but I didn't list them again.
Made a quick check list out of 'em to tick things done :)

dordsor21:
FAWE will need to wait for P2 to update before it can update. Testing should be done with FAWE integration off, or with the fawe hook disabled

NotMyFault:
> FAWE will need to wait for P2 to update before it can update. Testing should be done with FAWE integration off, or with the fawe hook disabled. Some of that looks like FAWE issues, not P2.

Good to know, I rested with WorldEdit and left a Fawe prefix for Fawe-specific issues, or at least something I couldn't replicate with WorldEdit.

dordsor21:
> > FAWE will need to wait for P2 to update before it can update. Testing should be done with FAWE integration off, or with the fawe hook disabled. Some of that looks like FAWE issues, not P2.
> 
> Good to know, I rested with WorldEdit and left a Fawe prefix for Fawe-specific issues, or at least something I couldn't replicate with WorldEdit.

The FAWE ones aren't going to be able to be rectified until after the PR is merged...

NotMyFault:
Yeah that for sure, I was solely putting it here, so we don't have to test existing issues again. 

If you don't mind and want to work on Fawe in parallel, which would be great for testing, I can deploy an incremental snapshot of this branch to Sonatype Snapshots.

dordsor21:
https://pastebin.com/NgBGjmi2 when regenerating all roads, however there's nothing to indicate this is a P2 issue and is completely outside the scope of this PR

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -430,26 +431,29 @@ public boolean scheduleRoadUpdate(
...
                     Iterator<BlockVector2> iter = chunks.iterator();
+                    QueueCoordinator queue = blockQueue.getNewQueue(worldUtil.getWeWorld(area.getWorldName()));
                     while (iter.hasNext()) {
                         BlockVector2 chunk = iter.next();
                         iter.remove();
-                        boolean regenedRoad = regenerateRoad(area, chunk, extend);
+                        boolean regenedRoad = regenerateRoad(area, chunk, extend, queue);
                         if (!regenedRoad) {
-                            LOGGER.info("Failed to regenerate roads");
+                            LOGGER.info(String.format("Failed to regenerate roads in chunk %s", chunk));
```
NotMyFault, line 447: You can use `{}` as variable placeholders for Log4J, this doesn't necessarily need to be String-ified.

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -474,18 +478,35 @@ public void run() {
...
                                     Iterator<BlockVector2> iterator = chunks.iterator();
-                                    while (System.currentTimeMillis() - start < 20 && !chunks.isEmpty()) {
+                                    if (chunks.size() >= 32) {
+                                        QueueCoordinator queue = blockQueue.getNewQueue(worldUtil.getWeWorld(area.getWorldName()));
+                                        for (int i = 0; i < 32; i++) {
+                                            final BlockVector2 chunk = iterator.next();
+                                            iterator.remove();
+                                            boolean regenedRoads = regenerateRoad(area, chunk, extend, queue);
+                                            if (!regenedRoads) {
+                                                LOGGER.info(String.format("Failed to regenerate road in chunk %s", chunk));
```
NotMyFault, line 489: Same like above

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -568,7 +589,35 @@ private int get_ey(final HybridPlotWorld hpw, QueueCoordinator queue, int sx, in
...
+    }
+
+    /**
+     * Regenerate the road in a chunk in a plot area.
+     *
+     * @param area             Plot area to regenerate road for
+     * @param chunk            Chunk location to regenerate
+     * @param extend           How far to extend setting air above the road
+     * @param queueCoordinator {@link QueueCoordinator} to use to set the blocks. Null if one should be created and enqueued
+     * @return if successful
```
NotMyFault, line 613: `since`, I guess?

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -594,92 +643,100 @@ public boolean regenerateRoad(final PlotArea area, final BlockVector2 chunk, int
...
-                                        finalZ + Z + plotWorld.ROAD_OFFSET_Z,
-                                        plotWorld.getPlotBiome()
-                                );
-                            }
+                        }
+                        BiomeType biome = plotWorld.G_SCH_B.get(MathMan.pair(absX, absZ));
+                        if (biome != null) {
+                            queue.setBiome(finalX + X + plotWorld.ROAD_OFFSET_X, finalZ + Z + plotWorld.ROAD_OFFSET_Z, biome);
+                        } else {
+                            queue.setBiome(
```
NotMyFault, line 728: This uses the now deprecated `#setBiome`, looking on the amount of args. Do we not support 3D biomes here?

NotMyFault (commented):
The IllegalStateExceptions are fired still, just the Fawe part is gone by now, because I'm using WorldEdit atm https://pastebin.com/sXfiPU8x

dordsor21:
> The IllegalStateExceptions are fired still, just the Fawe part is gone by now, because I'm using WorldEdit atm https://pastebin.com/sXfiPU8x

Yeah but that doesn't make it a P2 issue... The only thing we could possibly do that might not even help is either add tickets to every single chunk before editing, including a 1 chunk border, (or only edit one chunk st a time again...) or (it might work) ensuring that the chunks locations are stored in ordered sets (it might not work)

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
dordsor21: Somewhat out of the scope of the PR; 3D biomes aren't supported anywhere else in the plugin

NotMyFault: Oic, good to know 👍🏻

dordsor21:
The latest commit reduces the occurrence of the error massively, but not entirely

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -647,7 +647,7 @@ public void onBlockDamage(BlockDamageEvent event) {
                     event.getBlock().breakNaturally();
                 }
             }
-            if (location.getY() == 0) {
+            if (location.getY() == area.getMinBuildHeight()) {
```
SirYwell, line 650: see above

Core/src/main/java/com/plotsquared/core/util/ChunkUtil.java
```diff
@@ -97,14 +94,14 @@ public static int getX(int j) {
     }
 
     /**
-     * Gets the y coordinate for specific I and J values for a Chunk 16x16x16x16 layerxyz Array[16][4096].
+     * Gets the y coordinate for specific I and J values for a Chunk 16x16x16x16 layerxyz Array[N][4096].
```
SirYwell, line 97:
```suggestion
     * Gets the y coordinate for specific I and J values for a Chunk Nx16x16x16 layerxyz Array[N][4096].
```

SirYwell (dismissed):
I also saw a few more spots where `<=` and `<` are used wrongly. I'd suggest to kepp inclusiveness of the upper bound as constistent as possible.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -277,7 +277,7 @@ public void blockCreate(BlockPlaceEvent event) {
         BukkitPlayer pp = BukkitUtil.adapt(player);
         Plot plot = area.getPlot(location);
         if (plot != null) {
-            if ((location.getY() > area.getMaxBuildHeight() || location.getY() < area
+            if ((location.getY() >= area.getMaxBuildHeight() || location.getY() < area
```
traksag, line 280: `>` because max build height is inclusive?

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -362,7 +363,7 @@ public void blockDestroy(BlockBreakEvent event) {
                     event.setCancelled(true);
                     return;
                 }
-            } else if ((location.getY() > area.getMaxBuildHeight() || location.getY() < area
+            } else if ((location.getY() >= area.getMaxBuildHeight() || location.getY() < area
```
traksag, line 366: Same as above

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
```diff
@@ -134,7 +134,7 @@ public boolean enqueue() {
                 int sx = blockVector2.getX() << 4;
                 int sz = blockVector2.getZ() << 4;
                 if (isRegenChunk) {
-                    for (int layer = 0; layer < 16; layer++) {
+                    for (int layer = (getWorld().getMinY() >> 4); layer < (getWorld().getMaxY() >> 4); layer++) {
```
traksag, line 137: WE world max y is inclusive, so this should be `<=`

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -232,10 +237,10 @@ public void analyzeRegion(final String world, final CuboidRegion region, final R
                 for (int x = 0; x < width; x++) {
                     for (int z = 0; z < length; z++) {
                         Set<BlockType> types = new HashSet<>();
-                        for (int y = 0; y < 256; y++) {
-                            BlockState old = oldBlocks[y][x][z];
+                        for (int yIndex = 0; yIndex < height; yIndex++) {
```
traksag, line 240: `<=` since max gen height is inclusive?

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -244,23 +249,23 @@ public void analyzeRegion(final String world, final CuboidRegion region, final R
                                 } else {
                                     // check vertices
                                     // modifications_adjacent
-                                    if (x > 0 && z > 0 && y > 0 && x < width - 1 && z < length - 1 && y < 255) {
-                                        if (newBlocks[y - 1][x][z].getBlockType().getMaterial().isAir()) {
+                                    if (x > 0 && z > 0 && yIndex > 0 && x < width - 1 && z < length - 1 && yIndex < (height - 1)) {
```
traksag, line 252: `< height` since height is inclusive?

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
```diff
@@ -75,12 +75,26 @@ public QueueCoordinator(@Nullable World world) {
      * @param x chunk x coordinate
      * @param z chunk z coordinate
      * @return a new {@link ScopedQueueCoordinator}
+     * @deprecated Use {@link ScopedQueueCoordinator#getForChunk(int, int, int, int)}
      */
+    @Deprecated(forRemoval = true, since = "TODO")
     public ScopedQueueCoordinator getForChunk(int x, int z) {
+        return getForChunk(x, z, 0, 255);
```
traksag, line 82: The world's min and max height (inclusive) instead of 0 & 255?

traksag: Or are you leaving it 0 - 255 cause of deprecation? There's a similar case below.

Core/src/main/java/com/plotsquared/core/util/WorldUtil.java
```diff
@@ -81,11 +82,29 @@
      */
     public static void setBiome(String world, int p1x, int p1z, int p2x, int p2z, BiomeType biome) {
         BlockVector3 pos1 = BlockVector2.at(p1x, p1z).toBlockVector3();
-        BlockVector3 pos2 = BlockVector2.at(p2x, p2z).toBlockVector3(Plot.MAX_HEIGHT - 1);
+        BlockVector3 pos2 = BlockVector2.at(p2x, p2z).toBlockVector3(255);
```
traksag, line 85: Guessing this method will be deprecated due to use of 255 & 0?

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BukkitPlotGenerator.java
```diff
@@ -159,12 +160,14 @@ public ChunkGenerator getPlatformGenerator() {
             @NonNull BiomeGrid biome
     ) {
 
-        GenChunk result = new GenChunk();
+        int minY = BukkitWorld.getMinWorldHeight(world);
+        int maxY = BukkitWorld.getMaxWorldHeight(world);
+        GenChunk result = new GenChunk(minY, maxY);
```
traksag, line 165: I think the code in `ScopedQueueCoordinator ` (subclass of `GenChunk`) assumes its max is inclusive, so the max here should be inclusive too.

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
```diff
@@ -65,7 +65,10 @@ public void populate(final @NonNull World world, final @NonNull Random random, f
             return;
         }
         final ChunkWrapper wrap = new ChunkWrapper(area.getWorldName(), source.getX(), source.getZ());
-        final ScopedQueueCoordinator chunk = this.queue.getForChunk(wrap.x, wrap.z);
+        final ScopedQueueCoordinator chunk = this.queue.getForChunk(wrap.x, wrap.z,
+                com.plotsquared.bukkit.util.BukkitWorld.getMinWorldHeight(world),
+                com.plotsquared.bukkit.util.BukkitWorld.getMaxWorldHeight(world)
```
traksag, line 70: I think the code in `ScopedQueueCoordinator` assumes its max is inclusive, so the max here should be inclusive too.

traksag:
Maybe it would be less error prone if both `versionMaxHeight` and `World#getMaxHeight` are inclusive or both are exclusive, instead of the former inclusive and the latter exclusive. Probably inclusive would be better, since that's consistent with WE's world API.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
dordsor21: That's a mistake in the javadocs for the method. It's exclusive to maintain compatibility with previous versions (256 used to be the max, not 255)

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
dordsor21: ScopedQueuedCoordinator doesn't actually use max x/y/z currently

traksag: It calculates dx, dy, dz based on min/max in its constructor and then uses those

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
traksag: Hmm, though before this PR the check was with a `>`, so that would suggest inclusive. Maybe that was a bug haha. The default value of `maxBuildHeight` in `PlotArea` is currently `PlotSquared.platform().versionMaxHeight()` by the way, which is inclusive.

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
dordsor21: it used to be inclusive, so is currently okay (given latest commits)

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
dordsor21: That's been fixed now

NotMyFault:
I can't request your review @traksag but do you mind taking a look again?

traksag:
Sure, I can take another look later today

traksag (commented):
Finished my review. Slightly confusing that `PlotArea#getMaxBuildHeight` and `BukkitWorld.getMaxWorldHeight` are exclusive, while the other `max`'s are inclusive. Though I guess keeping them exclusive avoids another round of +/- 1 checking haha.

SirYwell (commented):
I'd heavily suggest to add methods for stuff like `getWorld().getMaxY() >> 4`, `layer - localChunk.getMinSection()` and `world.getMaxY() - world.getMinY()`. They can probably moved into a non-api class, but it would really help readability.

Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
```diff
@@ -276,15 +275,18 @@ public void run(ScopedQueueCoordinator value) {
                                     PlotLoc plotLoc = new PlotLoc(bx + x1, bz + z1);
                                     BaseBlock[] ids = map.allBlocks.get(plotLoc);
                                     if (ids != null) {
-                                        for (int y = 0; y < Math.min(128, ids.length); y++) {
+                                        int minY = value.getMin().getY();
+                                        for (int yIndex = 0; yIndex < Math.min(128 - minY, ids.length); yIndex++) {
```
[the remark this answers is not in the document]
dordsor21: The 128 magical number is a Jesse thing. Seemed to have been a minor optimisation for when blocks were set via bukkit API. Seems blocks stored in ContentMap should never be null so this doesn't need to exist anymore most likely

Core/src/main/java/com/plotsquared/core/generator/AugmentedUtils.java
```diff
@@ -132,7 +132,7 @@ public static boolean generate(
                         int worldZ = z + blockZ;
                         boolean can = manager.getPlotId(worldX, 0, worldZ) == null;
                         if (can) {
-                            for (int y = 1; y < 128; y++) {
+                            for (int y = area.getMinBuildHeight(); y <= area.getMaxGenHeight(); y++) {
```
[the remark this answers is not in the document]
dordsor21: The 128 was a magical value and assumed to be the highest the terrain could be for augmented worlds. With 1.18 this clearly changes drastically thus removing blocks where the road is up to the gen height is preferable

dordsor21:
Not sure a utility method needs to be created for getting a world's height, it's not really adding anything and it's quite self-explanatory when doing max - min height

dordsor21:
Everything seems to be fine for me now

NotMyFault (commented):
- It seems that plot backups are pasted one block higher than the original plot. I was clearing an empty plot at Y63 but it has been restored at Y64.

NotMyFault (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitChunkCoordinator.java
```diff
@@ -197,9 +205,11 @@ private void requestBatch() {
         BlockVector2 chunk;
         for (int i = 0; i < this.batchSize && (chunk = this.requestedChunks.poll()) != null; i++) {
             // This required PaperLib to be bumped to version 1.0.4 to mark the request as urgent
+            loadingChunks++;
             PaperLib
                     .getChunkAtAsync(this.bukkitWorld, chunk.getX(), chunk.getZ(), true, true)
                     .whenComplete((chunkObject, throwable) -> {
+                        loadingChunks--;
```
SirYwell, line 212: If I see this correctly, `loadingChunks` is accessed from different threads? (looking at the code below)

Core/src/main/java/com/plotsquared/core/plot/PlotCluster.java
```diff
@@ -89,11 +89,15 @@ public void setP2(PlotId id) {
...
-        this.region = RegionUtil.createRegion(this.pos1.getX(), this.pos2.getX(),
+        this.region = RegionUtil.createRegion(this.pos1.getX(), this.pos2.getX(), 0, 0,
                 this.pos1.getY(), this.pos2.getY()
         );
     }
 
+    /**
+     * Returns a region of PlotIDs
+     */
+    @Deprecated
```
SirYwell, line 100: Annotated as deprecated, but for what reason?

Core/src/main/java/com/plotsquared/core/queue/LocationOffsetDelegateQueueCoordinator.java
```diff
@@ -87,11 +87,14 @@ public boolean setBlock(int x, int y, int z, @NonNull Pattern pattern) {
...
     public boolean setBiome(int x, int z, @NonNull BiomeType biome) {
-        boolean result = true;
-        for (int y = 0; y < 256; y++) {
-            result &= this.setBiome(x, y, z, biome);
+        try {
+            if (canPlace[x - blockX][z - blockZ]) {
+                return super.setBiome(x, z, biome);
+            }
+        } catch (final Exception e) {
+            throw e;
```
SirYwell, line 95: The exception just gets rethrown

Core/src/main/java/com/plotsquared/core/util/MainUtil.java
```diff
@@ -37,15 +37,21 @@
      * Cache of mapping x,y,z coordinates to the chunk array<br>
      * - Used for efficient world generation<br>
      */
+    @Deprecated(forRemoval = true, since = "6.0.0")
```
SirYwell, line 40: all the `since` values should be TODO here I guess?

Core/src/main/java/com/plotsquared/core/util/WorldUtil.java
```diff
@@ -217,11 +233,14 @@ public abstract void setSign(
...
-     * @param region Region
-     * @param biome  New biome
+     * @param worldName World name
+     * @param region    Region
+     * @param biome     New biome
      */
-    public abstract void setBiomes(@NonNull String world, @NonNull CuboidRegion region, @NonNull BiomeType biome);
+    public void setBiomes(@NonNull String worldName, @NonNull CuboidRegion region, @NonNull BiomeType biome) {
+        final World world = getWeWorld(worldName);
+        region.iterator().forEachRemaining(bv -> world.setBiome(bv, biome));
```
SirYwell, line 242: Why not just `forEach` directly?

SirYwell (dismissed):

Core/src/main/java/com/plotsquared/core/queue/LocationOffsetDelegateQueueCoordinator.java
dordsor21: Matches every other method in the class

Core/src/main/java/com/plotsquared/core/util/WorldUtil.java
dordsor21: doesn't exist for Sets iirc

Core/src/main/java/com/plotsquared/core/plot/PlotCluster.java
SirYwell: I see, yeah that's bad

Core/src/main/java/com/plotsquared/core/queue/LocationOffsetDelegateQueueCoordinator.java
SirYwell: Could be cleaned up then I guess, but we can do that separately to avoid even more diff

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitChunkCoordinator.java
SirYwell: Yeah that's what I'd expect, but then the code below should probably removed at some point

NotMyFault:
Conflicts + security scan failure

SirYwell (approved):
I didn't test it ingame, but looks good code-wise.

NotMyFault:
Next week would be a release week, I'm gonna give it a try later this week and if I don't find something new or there is no more negative feedback, we can pull it by end of the weekend I guess.

PR #3496, opened 2022-02-09
Don't open the component GUI if the actor has no permission for a specific component

NotMyFault:
## Description
If presets have a permission set and the actor has no permission to interact with any preset, an IllegalArgumentException is thrown, because the inventory we create is empty and therefore has no size.
Additionally, this makes the GUI unusable, because we don't return from that state.

## Submitter Checklist

## Maintainer checklist

- [X] Changelog entries in the PR title are correct
- [X] Proper changelog labels are set so that the changelog can be generated automatically


Citymonstret (approved):

Core/src/main/java/com/plotsquared/core/components/ComponentPresetManager.java
```diff
@@ -181,7 +181,13 @@ public ComponentPresetManager(final @NonNull EconHandler econHandler, final @Non
...
             allowedPresets.add(componentPreset);
         }
-        final int size = (int) Math.ceil((double) allowedPresets.size() / 9.0D);
+        final int size;
+        if (allowedPresets.isEmpty()) {
+            player.sendMessage(TranslatableCaption.of("preset.empty"));
+            return null;
+        } else {
+            size = (int) Math.ceil((double) allowedPresets.size() / 9.0D);
+        }
```
SirYwell, line 190: The size does not depend on the `if` part. It would be sufficient to just do the check before the old `final int size = ...` code.

SirYwell (approved):

NotMyFault:

PR #3524, opened 2022-03-02
Change the default sign orientation to better match the wall location

dordsor21:


NotMyFault:
Mind elaborating what that fixes?

dordsor21:
Floating signs facing the wrong direction. Now defaults to being "against the wall" even if it is air.

NotMyFault (changes requested):
Not... quite what we want 👀

NotMyFault:
I can replicate it constantly with WorldEdit: https://streamable.com/nal9zp
Using Fawe over WE works fine though.

NotMyFault:
Only thing I can think of is the sign being placed before the border is removed, hence the block state updates and the block drops using WE.
Fawe doesn't care about it hence it stays, I guess?

NotMyFault (approved):

SirYwell (approved):

PR #3526, opened 2022-03-06
Draft: Recalculate light levels on chunk generation

### LINKED ISSUE #3514: Is LightCleaner still necessary?
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

When generating new chunks with a custom template that includes lighting-blocks, the lighting is not updated after the chunk is placed the first time. Instead LightCleaner was advertised in the past. 

The problem is, that LightCleaner goes into an infinite loop of calculating light for unloaded chunks and this is costing a lot of performance. The author of that plugin said, that the problem relies elsewhere and it is not his fault, so I wonder where this problem comes from.

### To Reproduce

1. Use Template with lighting blocks, for example 'Bridge.template'
2. Generate Chunks
3. See no lighting

### Expected behaviour

PlotSquared should handle the lighting itself.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/ca36ed467388451b9ef4e047cf52e247

### PlotSquared Version

PlotSquared version 6.3.0

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview

Fixes #3514

## Description
Applies light updates to blocks with light values on chunk generation using a populator and WorldEdits World#applySideEffects-method.
Draft, as sometimes that doesn't seem to work and chunk borders sometimes have rough edges.

## Submitter Checklist

- [ ] I tested my changes and approved their functionality.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
@@ -33,20 +33,32 @@ import com.plotsquared.core.inject.factory.HybridPlotWorldFactory;
 import com.plotsquared.core.location.Location;
 import com.plotsquared.core.plot.PlotArea;
 import com.plotsquared.core.plot.PlotId;
+import com.plotsquared.core.queue.LightingMode;
 import com.plotsquared.core.queue.ScopedQueueCoordinator;
 import com.plotsquared.core.util.MathMan;
+import com.sk89q.worldedit.WorldEditException;
+import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.util.SideEffect;
+import com.sk89q.worldedit.util.SideEffectSet;
 import com.sk89q.worldedit.world.biome.BiomeType;
 import com.sk89q.worldedit.world.block.BaseBlock;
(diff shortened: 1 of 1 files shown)


dordsor21:
What's the performance impact of this? Also is setting blocks via the world safe to do during the populate chunk phase, given neighbour-generation-state requirements?

PierreSchwang:
> What's the performance impact of this? Also is setting blocks via the world safe to do during the populate chunk phase, given neighbour-generation-state requirements?

Before caching the light blocks it was quite laggy, but after caching them when setting them and interating over them changes nothing in performance.
I do update the light levels on population, therefor after the chunk generation itself. It does work as I stated, but sometimes chunk edges or blocks won't update properly

dordsor21:
A reasonable extension to this would be setting NBT data as well (https://github.com/IntellectualSites/PlotSquared/issues/3579)

SirYwell:
At this point, it might make sense to use the new worldgen api (and fallback to the "broken" behavior on older versions)

PierreSchwang:
Is this now superseded / implemented by https://github.com/IntellectualSites/PlotSquared/pull/3659 and can be closed?

PR #3533, opened 2022-03-08
Don't expose jcip and findbugs annotations to the world on runtime

NotMyFault:
## Description
The change proposed addresses findbugs and jcip not being relocated in the jar.
Additionally labeling as "Developer", because you never know where people pull their dependencies from...
I've also taken the shot to reorganize our internal structure of dependencies, grouping them in sub folders.

## Submitter Checklist


dordsor21:
Is it worth grouping and commenting by the subpackage (web/internal etc)?

dordsor21 (approved):
¯\_(ツ)_/¯

SirYwell:
Modifying relocations of libraries exposed in API is a breaking change. Especially guice should be considered here I think.

NotMyFault:
> Modifying relocations of libraries exposed in API is a breaking change. Especially guice should be considered here I think.

We don't publish shadow artifacts, none of the relocation is available in the API artifacts in any way.

NotMyFault:
I've reverted the commit to comply with the PR title. I'll look forward to change the relocation pattern within v7, yet to notice, this still doesn't affect API consumers, because we don't publish builds containing these libraries.

cc @SirYwell

SirYwell (approved):

PR #3584, opened 2022-04-23
adding option to disable Mojang uuid API calls and some improovements to /p list

### LINKED ISSUE #2959: option to replace invalid (offline) UUIDs

Can we add an option to delete / override invalid UUID's in the database? Or a function to hide this old UUIDs from the UUID cache on the startup?

`[21:40:48] [Craft Scheduler Thread - 94/INFO]: [PS]: (UUID) Request for 030de5ae-a083-3581-9129-d0bb4e98da7b timed out`

`[21:46:06] [pool-16-thread-585/INFO]: [ParzellenSystem]: (UUID) Found invalid UUID: 030de5ae-a083-3581-9129-d0bb4e98da7b`

Otherwise, this invalid UUID's will check with every startup. In this way this feature can get a performance improvement on servers with this problem.

Which kind of solutions are valid? Maybe a replacement from `invalid UUID` to `null`?
Maybe a problem is the check in which table the UUID stand on.

**Additional context**
I have my server since over 8 years. So, I have an amount of old / invalid users registered in my PS database. I think it's a good feature to help people with the same problem without to delete all old plot.

**Checklist**:

- [X] I made sure there are no duplicates of my suggestion [(Use search)](https://github.com/IntellectualSites/PlotSquaredSuggestions/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+)
- [X] I made sure my suggestion does not exist yet [(Check the wiki)](https://wiki.intellectualsites.com/plotsquared/home)
- [X] I made sure my suggestion is useful for the majority of PlotSquared users

Xaver106:
### Fixes #2959
With random Mojang API calls disabled the plugin should be less likely to throw a timeout error. The lookup should have enough time to check all other providers before timing out. This might still be a problem if too many invalid lookups happen at once. Replacing invalid UUIDs should not be necessary, but will still improve efficiency a bit.

## Description

- Adding option to disable random Mojang API calls and only use the cache. Unknown UUIDs will already be returned as unknown (`info.unknown` from messages_en.json)
- List command will now show if a plot is a server plot and if the owner is unknown. Also improved showcase if everyone is owner of a plot
- The color of the plot will now show correctly if plot is owned by everyone

## Submitter Checklist


dordsor21:
Whilst good ideas, this should ideally be split into several PRs for each "part", allowing for easier commenting/suggestions

Xaver106:
> Whilst good ideas, this should ideally be split into several PRs for each "part", allowing for easier commenting/suggestions

I initially didn't plan to have two different ones in here and it seemed unnecessary to change it at the end. I can separate it if needed though.

NotMyFault:
Please separate different topics in PRs to allow a faster review while making it easier to bisect regressions.

PR #3612, opened 2022-05-10
Implement chunkObject into queueing

dordsor21:
- This should only be used in generation
 - The chunk will be used as fallback to set to using bukkit API if WNA fails

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
@@ -51,6 +51,7 @@ import com.sk89q.worldedit.world.biome.BiomeType;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockState;
 import org.bukkit.Bukkit;
+import org.bukkit.Chunk;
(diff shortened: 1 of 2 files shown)


Citymonstret (approved):
lgtm

Bukkit/src/main/java/com/plotsquared/bukkit/queue/BukkitQueueCoordinator.java
```diff
@@ -266,7 +267,13 @@ private void setWorldBlock(int x, int y, int z, @NonNull BaseBlock block, @NonNu
         } catch (WorldEditException ignored) {
             // Fallback to not so nice method
             BlockData blockData = BukkitAdapter.adapt(block);
-            Block existing = getBukkitWorld().getBlockAt(x, y, z);
+            Block existing;
+            // Assume a chunk object has been given only when it should have been.
+            if (getChunkObject() instanceof Chunk) {
+                existing = ((Chunk) getChunkObject()).getBlock(x & 15, y, z & 15);
```
SirYwell, line 273:
```suggestion
            if (getChunkObject() instanceof Chunk chunkObject) {
                existing = chunkObject.getBlock(x & 15, y, z & 15);
```

SirYwell (changes requested):

NotMyFault (approved):
Nice one

PR #3614, opened 2022-05-10
Fix lag caused when generating augmented worlds with roads

dordsor21:
- Begin by implementing forceSync to the queue system as we know the chunk will be accessible to edits in some cases (i.e. population).
 - Also implement custom SideEffectSets to override any decided by the default chunk consumer, as we do NOT want to update neighbours in population (this caused infinite generation to be required causing the lag and server death). We also do not want to enqueue the QueueCoordinator in AugmentedUtils as this would write to the world and update neighbours before we might want to (plus it's just used to restrict and offset the blocks being set)
 - Then implement disabling any biomes from being saved/set to the queue to prevent augmented worlds having their biomes overriden in roads
 - Consequently fix ScopedQueueCoordinator, preventing the y value of blocks being set from needlessly being changed, fixing road heights in augmented worlds
 - Finally we do not need a method with chunkObject in the signature in AugmentedUtils as this is no longer used by the method


Core/src/main/java/com/plotsquared/core/generator/AugmentedUtils.java
```diff
@@ -172,13 +175,26 @@ public static boolean generate(
...
             queue.enqueue();
         }
         return generationResult;
     }
 
+    /**
+     * @deprecated Use {@link AugmentedUtils#generateChunk(String, int, int, QueueCoordinator)} as chunkObject is not required
+     * in the above method
+     */
+    @Deprecated(forRemoval = true)
```
NotMyFault, line 189:
```suggestion
    @Deprecated(forRemoval = true, since = "TODO")
```

Core/src/main/java/com/plotsquared/core/generator/AugmentedUtils.java
```diff
@@ -54,12 +54,22 @@ public static void bypass(boolean bypass, Runnable run) {
...
+     * Generate an augmented world chunk at the given location. If a queue is given, the data will be written to it, else a new
+     * queue will be created and written to world. Returns true if generation occurred.
+     *
+     * @param world  World name to generate data for. Must be a PlotSquared world containing one or more areas else nothing will
+     *               happen.
+     * @param chunkX Chunk X position
+     * @param chunkZ Chunk Z position
+     * @param queue  Queue to write to, if desired.
+     * @return true if generation occurred.
+     */
```
NotMyFault, line 67:
```suggestion
     * @since TODO
     */
```

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
```diff
@@ -247,6 +248,13 @@ public boolean setBlock(final int x, final int y, final int z, final @NonNull Pa
      */
     public abstract boolean isSettingBiomes();
 
+    /**
+     * If the queue should accept biome placement
+     *
+     * @param enabled If biomes should be enabled
```
NotMyFault, line 254:
```suggestion
     * @param enabled If biomes should be enabled
     * @since TODO
```

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
```diff
@@ -412,6 +420,10 @@ public boolean enqueue() {
      */
     public abstract void setLightingMode(@Nullable LightingMode mode);
 
+    public abstract SideEffectSet getSideEffectSet();
+
```
NotMyFault, line 424: Please add javadocs, as appropriately.

NotMyFault (commented):
I didn't have much luck testing this in practice, after rebooting the server https://gist.github.com/NotMyFault/f8709adf8061362fbe9a31756f2227ef

dordsor21:
Ah yeah this requires WorldEdit for the time being (or thr SideEffectSet PR in FAWE)

NotMyFault (approved):

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
```diff
@@ -412,6 +421,20 @@ public boolean enqueue() {
      */
     public abstract void setLightingMode(@Nullable LightingMode mode);
 
+    /**
+     * Get the overriding {@link SideEffectSet} to be used by the queue if it exists, else null
+     *
+     * @return Overriding {@link SideEffectSet} or null
+     */
```
SirYwell, line 428: We could probably make use of the new `{@return}` inline tag, like 
```suggestion
    /**
     * {@return the overriding {@link SideEffectSet} to be used by the queue if it exists, else null}
     */
```

SirYwell (approved):
Looks good

Core/src/main/java/com/plotsquared/core/queue/QueueCoordinator.java
NotMyFault: Keep in mind that this feature is not supported by major IDE vendors, like IntelliJ, yet and renders as plain text.

NotMyFault:
Thanks for your contribution! We will merge your PR within the next ~24 hours, as outlined in our [PR review process](https://github.com/IntellectualSites/IntellectualSites/blob/main/docs/pull-requests/pull-request-review.adoc#step-3-marking-for-merge).

PR #3618, opened 2022-05-10
Fix plot analysis

### LINKED ISSUE #3464: plot analysis / calibration returns unrelated values
### Server Implementation

Paper

### Server Version

1.18.1

### Describe the bug

The plot analysis as of now is quite unrelated to the plot itself. 
There were a few bugs which would break the plot analysis itself, which were addressed in [`29c054e` (#3452)](https://github.com/IntellectualSites/PlotSquared/pull/3452/commits/29c054e2a2f081e3a06fcad2273e52795a7d577e), but the calculation of the plot returns some weird values unrelated to the plot itself.

- Multiple empty plots have different changes / column
- Plot values seem to be affected by the plot world grid layout itself (based on testing by @iLemon, the cells in the image are plots)
grafik

### To Reproduce

1. Use build from #3452 
2. Run /plot debugexec analyze on multiple plots

### Expected behaviour

Valid values based on the plot itself

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

-/-

### PlotSquared Version

Commit 1542db6266e3b539a73048752ede57d4e3421dee (6.3.1)

### Checklist

### Anything else?

_No response_

### LINKED ISSUE #3600: plot autoclear not working
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

The autoclear task, defined in the settings, never executes. Also there is some error with plot expiration on server startup.

### To Reproduce

Scroll down my server log.

### Expected behaviour

Plots get cleared, if they match the defined task

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://gist.github.com/TheLethalGoose/fe9c9adc72facc60024ae4a5d2ba8f7a

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/9ae31304c0f342a0bdcd3d4f75ee5e3f

### PlotSquared Version

6.7.0-Premium

### Checklist

### Anything else?

_No response_

dordsor21:
- Stop using deprecated ChunkQueueCoordinator and create a new purpose-built coordinator
 - Generation is chunk-by-chunk thus the old blocks cache needs to be filled accordingly
 - Remove the **four** System#gc calls
 - Fixes #3464
 - Fix really weird dumb... maths? in ExpiryTask
 - Fixes #3600

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
@@ -83,6 +83,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class HybridUtils {
 
     private static final Logger LOGGER = LogManager.getLogger("PlotSquared/" + HybridUtils.class.getSimpleName());
+    private static final BlockState AIR = BlockTypes.AIR.getDefaultState();
 
     public static HybridUtils manager;
     public static Set<BlockVector2> regions;
@@ -157,40 +158,28 @@ public class HybridUtils {
             final int height = area.getMaxGenHeight() - area.getMinGenHeight() + 1;
             final int minHeight = area.getMinGenHeight();
 
-            final BlockState airBlock = BlockTypes.AIR.getDefaultState();
-            final BlockState[][][] oldBlocks = new BlockState[height][width][length];
             final BlockState[][][] newBlocks = new BlockState[height][width][length];
-            for (final BlockState[][] newBlock : newBlocks) {
-                for (final BlockState[] blockStates : newBlock) {
-                    Arrays.fill(blockStates, airBlock);
-                }
-            }
-            for (final BlockState[][] oldBlock : oldBlocks) {
-                for (final BlockState[] blockStates : oldBlock) {
-                    Arrays.fill(blockStates, airBlock);
-                }
-            }
 
             BlockArrayCacheScopedQueueCoordinator oldBlockQueue = new BlockArrayCacheScopedQueueCoordinator(
-                    oldBlocks,
-                    area.getMinGenHeight(),
-                    Location.at("", region.getMinimumPoint()),
-                    Location.at("", region.getMaximumPoint())
+                    Location.at("", region.getMinimumPoint().withY(hpw.getMinGenHeight())),
+                    Location.at("", region.getMaximumPoint().withY(hpw.getMaxGenHeight()))
             );
 
-            region.getChunks().forEach(blockVector2 -> {
-                int relChunkX = blockVector2.getX() - cbx;
-                int relChunkZ = blockVector2.getZ() - cbz;
+            region.getChunks().forEach(chunkPos -> {
+                int relChunkX = chunkPos.getX() - cbx;
+                int relChunkZ = chunkPos.getZ() - cbz;
                 oldBlockQueue.setOffsetX(relChunkX << 4);
                 oldBlockQueue.setOffsetZ(relChunkZ << 4);
                 hpw.getGenerator().generateChunk(oldBlockQueue, hpw);
             });
 
+            final BlockState[][][] oldBlocks = oldBlockQueue.getBlockStates();
+
             QueueCoordinator queue = area.getQueue();
             queue.addReadChunks(region.getChunks());
-            queue.setChunkConsumer(blockVector2 -> {
-                int X = blockVector2.getX();
-                int Z = blockVector2.getZ();
+            queue.setChunkConsumer(chunkPos -> {
+                int X = chunkPos.getX();
+                int Z = chunkPos.getZ();
                 int minX;
                 if (X == cbx) {
                     minX = bx & 15;
@@ -228,6 +217,9 @@ public class HybridUtils {
                         for (int yIndex = 0; yIndex < height; yIndex++) {
                             int y = yIndex + minHeight;
                             BlockState block = queue.getBlock(xx, y, zz);
+                            if (block == null) {
+                                block = AIR;
+                            }
                             int xr = xb + x;
                             int zr = zb + z;
                             newBlocks[yIndex][xr][zr] = block;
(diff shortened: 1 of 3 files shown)


Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
```diff
@@ -0,0 +1,168 @@
...
+    private int offsetX = 0;
+    private int offsetZ = 0;
+
+    /**
+     * Construct a new instance
+     *
+     * @param blockStates Array of form BlockState[y][x][z]. Must be fully initialised.
+     * @param minY        Minimum applicable y value. Used to account for negative world heights (e.g. -64). Inclusive
+     * @param min         Inclusive location of the minimum point to limit the scope to.
+     * @param max         Exclusive location of the maximum point to limit the scope to.
```
NotMyFault, line 64:
```suggestion
     * @param max         Exclusive location of the maximum point to limit the scope to.
     * @since TODO
```

NotMyFault (approved):

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
```diff
@@ -0,0 +1,169 @@
...
+        if (x >= scopeMinX && x < scopeMaxX && y >= minY && y < maxY && z >= scopeMinZ && z < scopeMaxZ) {
+            blockStates[y - minY][x][z] = id.toImmutableState();
+        }
+        return false;
+    }
+
+    @Override
+    public @Nullable BlockState getBlock(final int x, final int y, final int z) {
+        if (x >= 0 && x < width && y >= minY && y < maxY && z >= 0 && z < length) {
+            return blockStates[y - minY][x][z];
```
SirYwell, line 128: I guess you need to add the offset for x and z here too, as in setBlock. Also, the bounds check should probaböy moved to an extra method to make it more consostent.

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Construct a new instance
+     *
+     * @param blockStates Array of form BlockState[y][x][z]. Must be fully initialised.
+     * @param minY        Minimum applicable y value. Used to account for negative world heights (e.g. -64). Inclusive
+     * @param min         Inclusive location of the minimum point to limit the scope to.
+     * @param max         Exclusive location of the maximum point to limit the scope to.
+     * @since TODO
+     */
+    public BlockArrayCacheScopedQueueCoordinator(@NonNull BlockState[][][] blockStates, int minY, Location min, Location max) {
```
SirYwell, line 67: Passing the array is a little bit odd, it allows outside modification of an internal detail. Shouldn't it be enough to have min and max locations to create the array in the constructor? (note that the javadocs would need to be changed accordingly if this is changed). The name of the class could then also be changed to something more generic.

Core/src/main/java/com/plotsquared/core/generator/HybridUtils.java
```diff
@@ -175,8 +171,20 @@ public void analyzeRegion(final String world, final CuboidRegion region, final R
...
-            System.gc();
-            System.gc();
+            BlockArrayCacheScopedQueueCoordinator oldBlockQueue = new BlockArrayCacheScopedQueueCoordinator(
+                    oldBlocks,
+                    area.getMinGenHeight(),
+                    Location.at("", region.getMinimumPoint()),
+                    Location.at("", region.getMaximumPoint())
+            );
+
+            region.getChunks().forEach(blockVector2 -> {
```
SirYwell, line 181: the lambda parameter should probably be called something like `chunkPos` to avoid confusion

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Override
+    public @NonNull Location getMin() {
+        return super.getMin().add(offsetX - scopeMinX, 0, offsetZ - scopeMinZ);
+    }
+
+    @Override
+    public @NonNull Location getMax() {
+        return getMin().add(15, 0, 15).withY(maxY);
```
SirYwell, line 112: this looks wrong, we're not enforcing a 16x16 limit anywhere else in this class.

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
dordsor21: The queue isn't used for reading blocks anywhere, but I purposely left the offset out of this; writing blocks is specifically done by the generator that writes 0->15 for x/z and thus needs to be offset, and given the point of the queue is to basically aggregate multiple chunks worth of generation, I think it makes sense for any "get" calls to be normalised to the minimum point, but still able to access the entire cache regardless of the offset

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
dordsor21: The class has a very specific use in the first place. It was done to allow the array be filled by default with air (which would not be ideal for a generic queue class) as per the analyzeRegion method in HybridUtils. I suppose the oldBlocks variable doesn't necessarily need to be available there, but performance is increased at effectively no memory cost by simply referencing the array directly without checking if the coordinates are within range every single time (we know they are because that's the coordinates we iterate over in analyzeRegion). Keeping this performance improvement would require a method to retrieve the array cache from the queue anyway, and unless we add the extra overhead of cloning the entire BlockState[][][] array, it is still mutable at that point anyway.

Given generation would actually set air to the entire array anyway (and a null check can always be added to assume air), it probably would be better to have a method to simply return the array (as in the ChunkQueueCoordinator class that was used initially).

Though thinking of it, there is no reason why we couldn't just analyze chunk-by-chunk as the chunks in the world are loaded, using a off-the-main-thread single-threaded executor...

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
dordsor21: This is because of the very specific use case of the class for caching generation. I do, however, think the whole thing should be basically completely rewritten (see the short essay above) which would remove the necessity for this class and go back to the ChunkQueueCoordinator/similar

Core/src/main/java/com/plotsquared/core/queue/BlockArrayCacheScopedQueueCoordinator.java
dordsor21: Scrap that, we need to full data at once in order to properly analyse the faces

SirYwell (changes requested):
With the given context, this looks good to me. However, I think the new queue coordinator should be marked as internal.

SirYwell (approved):

PR #3627, opened 2022-05-18
feature: add flag for beacon effects of other plots

### LINKED ISSUE #3626: Feeature Request: Add flag for preventing beacon effects of other plots

A flag to toggle on your own plot that disables all effects comming from beacons that are not placed on the same plot

(No known alternative)

### Anything else?

_No response_

DerEingerostete:
## Overview
Adds a Plot Flag that enables players to disable beacon effects from other plots
Fixes #3626

## Description
Adds the new Plot Flag "beacon-effects" to toggle beacon effects from other plots.
By default beacon effects from other plots are enabled but can be toggled off by setting this flag to false.
Note: Works only on Paper (requires the BeaconEffectEvent from Paper)

## Submitter Checklist


Core/src/main/java/com/plotsquared/core/plot/flag/GlobalFlagContainer.java
```diff
@@ -200,6 +201,7 @@ private GlobalFlagContainer() {
         this.addFlag(VillagerInteractFlag.VILLAGER_INTERACT_FALSE);
         this.addFlag(VineGrowFlag.VINE_GROW_TRUE);
         this.addFlag(ProjectilesFlag.PROJECTILES_FALSE);
+        this.addFlag(BeaconEffectFlag.BEACON_EFFECT_TRUE);
```
NotMyFault, line 204: If the necessary API doesn't exist on Spigot, the flag shouldn't be added to the global flag container, if the server runs Spigot.

NotMyFault (commented):
Would you mind raising a PR to the documentations repository, documenting your change? https://github.com/IntellectualSites/plotsquared-documentation/pull/5

Core/src/main/java/com/plotsquared/core/plot/flag/GlobalFlagContainer.java
RedstoneFuture: Unfortunately, that is the problem. Even if I like this flag idea.

RedstoneFuture: Oh, sry.

dordsor21:
I think it would also be a good idea to add a global prevention for beacon effects "leaking" out of plots, able to be enabled/disabled under the paper block in settings.yml (probably enabled by default). This way, beacons can still affect players in their own plot, but they will not be affected by beacons in other plots and vice versa.

Core/src/main/java/com/plotsquared/core/plot/flag/GlobalFlagContainer.java
DerEingerostete: > If the necessary API doesn't exist on Spigot, the flag shouldn't be added to the global flag container, if the server runs Spigot.

Is there a way to use paper specific flags? Because in the Core module you can't access the PaperLib library

Core/src/main/java/com/plotsquared/core/configuration/Settings.java
```diff
@@ -666,6 +666,8 @@ public static final class Paper_Components {
         public static boolean TILE_ENTITY_CHECK = true;
         @Comment("Use Paper's async tab completion")
         public static boolean ASYNC_TAB_COMPLETION;
+        @Comment("Allow beacon effects originating from other plots affecting the own plot")
+        public static boolean BEACON_EFFECTS = false;
```
NotMyFault, line 670: This configuration option is too ambiguous, how about `cross plot beacon effects`?

DerEingerostete: You mean ``Allow cross plot beacon effects``? Okay, should I change it in the docs as well?

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
```diff
@@ -407,4 +409,25 @@ public void onAsyncTabCompletion(final AsyncTabCompleteEvent event) {
...
+        if (beaconPlot == null) {
+            return;
+        }
+
+        Player player = event.getPlayer();
+        Location playerLocation = BukkitUtil.adapt(player.getLocation());
+
+        PlotPlayer<Player> plotPlayer = BukkitUtil.adapt(player);
+        Plot playerStandingPlot = playerLocation.getPlot();
+        boolean effectDisabled = !Settings.Paper_Components.BEACON_EFFECTS ||
```
RedstoneFuture, line 426: I would not double negate the "if" so often. And the paper component check could also be implemented at the beginning. :)

DerEingerostete: With the paper comonent check do you mean the ``if (PaperLib.isPaper()) {}`` in the ``BukkitPlatform`` class?

RedstoneFuture: https://github.com/DerEingerostete/PlotSquared/blob/d6f0abfbb40aef5fa76e4ec1d6f394bb060294d6/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java#L414

Core/src/main/java/com/plotsquared/core/configuration/Settings.java
NotMyFault: Please do that, thanks.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
DerEingerostete: Thank you!

DerEingerostete: I just tested the code and the paper component check cannot be on top because it's either disabled by config or disabled by flag. If I would put it at the beginning it would never cancel the event when set to true

RedstoneFuture:
I'm not sure if the option/flag names are user-friendly. Now we have the following setup:

- Config option "BEACON_EFFECTS", `true` by default = Prevent cross plot beacon effects, if it's `true`
- Flag "BeaconEffect", `true` by default = Prevent cross plot beacon effects, if it's `true`

DerEingerostete:
> I'm not sure if the option/flag names are user-friendly. Now we have the following setup:
> 
>     * Config option "BEACON_EFFECTS", `true` by default = Prevent cross plot beacon effects, if it's `true`
> 
>     * Flag "BeaconEffect", `true` by default = Prevent cross plot beacon effects, if it's `true`

Should I remove the flag and only keep the setting?

NotMyFault (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
```diff
@@ -407,4 +409,27 @@ public void onAsyncTabCompletion(final AsyncTabCompleteEvent event) {
...
+        if (beaconPlot == null) {
+            return;
+        }
+
+        Player player = event.getPlayer();
+        Location playerLocation = BukkitUtil.adapt(player.getLocation());
+
+        PlotPlayer<Player> plotPlayer = BukkitUtil.adapt(player);
+        Plot playerStandingPlot = playerLocation.getPlot();
+        if (playerStandingPlot == null || beaconPlot.equals(playerStandingPlot)) {
```
dordsor21, line 426: It's probably ideal to check if the player is in a plot area, and thus beacon effects can be prevented from "leaking" into the road as well as adjacent plots. Having it as a road flag would also be useful, so checking the road flag container

DerEingerostete: I added a commit. Hope it fixed everything you mentioned

Core/src/main/java/com/plotsquared/core/plot/flag/implementations/DisableBeaconEffectOverflowFlag.java
```diff
@@ -29,17 +29,17 @@
 import com.plotsquared.core.plot.flag.types.BooleanFlag;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
-public class BeaconEffectFlag extends BooleanFlag<BeaconEffectFlag> {
+public class DisableBeaconEffectOverflowFlag extends BooleanFlag<DisableBeaconEffectOverflowFlag> {
```
dordsor21, line 32: I thought it reasonable for the flag itself to prevent/enable all beacon effects on the plot, and only meant the config option should be renamed. Having the flag be true to allow/false to deny all beacon effects on the plot is more what players would want to do, whilst preventing overflow from/between plots is likely to be a more admin thing to do.

DerEingerostete: Okay now I got what you meant. Changed it

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
```diff
@@ -429,18 +426,22 @@ public void onBeaconEffect(final BeaconEffectEvent event) {
...
-                    || Settings.Paper_Components.DISABLE_BEACON_EFFECT_OVERFLOW) {
+            if (!area.getRoadFlag(BeaconEffectsFlag.class) ||
+                    (beaconPlot != null && Settings.Paper_Components.DISABLE_BEACON_EFFECT_OVERFLOW)) {
                 event.setCancelled(true);
             }
             return;
         }
 
-        if (beaconPlot.equals(playerStandingPlot)) {
+        boolean plotBeaconEffects = playerStandingPlot.getFlag(BeaconEffectsFlag.class);
```
dordsor21, line 436: This would NPE on spigot platforms as we do not set the flag to the global container. You should use `playerStandingPlot.getFlagContainer().getFlag()` and then check for nullability before doing `Flag#getValue`. The same issue will occur for getting the road flag.

OR

We could still use the flag for spigot, where any beacon effect is cancelled based on the player's location, and add a new listener that's only enabled if paper is not present. Perhaps in a new class called "SpigotListener". @IntellectualSites/plotsquared-team thoughts on this?

SirYwell: I agree. Having all registrations in one place also makes the code a lot clearer.

NotMyFault:
Looks good overall, thanks for addressing the outstanding concerns.

Minor nit: The settings.yml option should consider `PlayerLeavePlotEvent`, otherwise players can step on the plot, grab the effect and leave again, which kinda defeats the purpose of the flag.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
```diff
@@ -446,4 +450,11 @@ public void onBeaconEffect(final BeaconEffectEvent event) {
         }
     }
 
+    private boolean getBooleanFlagValue(@NonNull FlagContainer container,
```
dordsor21, line 453: There's no longer a need to do this as the flag is now added to the GlobalFlagContainer

Bukkit/src/main/java/com/plotsquared/bukkit/listener/SpigotListener.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Fallback listener for paper events on spigot
+ */
+public class SpigotListener implements Listener {
+
+    @EventHandler(ignoreCancelled = true)
+    public void onEffect(@NotNull EntityPotionEffectEvent event) {
+        if (event.getCause() != EntityPotionEffectEvent.Cause.BEACON) return;
```
dordsor21: For P2 codestyle this should have braces

DerEingerostete: Oh sorry. Always do this in my own projects and tried to change it for PlotSquared. Have fallen into old habits there. Will change it

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PaperListener.java
DerEingerostete: Should I revert back to the old way there or is this way also okay?

DerEingerostete:
> Works out nicely overall, thanks for addressing the outstanding concerns.
> 
> Minor nit: The settings.yml option should consider `PlayerLeavePlotEvent`, otherwise players can step on the plot, grab the effect and leave again, which kinda defeats the purpose of the flag.

Okay will remember that and change it. Is this event an event working with the Bukkit  EventHandler or do I need a special class/listener?

dordsor21:
You can add it to the plotExit method in `com.plotsquared.core.listener.PlotListener`

DerEingerostete:
> You can add it to the plotExit method in `com.plotsquared.core.listener.PlotListener`

This saidly dosen't work because the player abstraction dosen't have a remove effect mathod. Should I add one or rather register a own listener in the Bukkit module?

dordsor21:
It would probably be best to add it to the PlotPlayer

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -204,6 +206,29 @@ public PlayerEventListener(
...
+            return;
+        }
+
+        UUID uuid = player.getUniqueId();
+        PotionEffect effect = event.getNewEffect();
+        if (effect == null) {
+            PotionEffect oldEffect = event.getOldEffect();
+            if (oldEffect != null) {
+                String name = oldEffect.getType().getName();
+                this.addEffect(uuid, name, -1);
```
NotMyFault, line 223: How did you test this out? #addEffect isn't accessible in the Bukkit module the way you implemented it in the Core module.

Core/src/main/java/com/plotsquared/core/listener/PlotListener.java
```diff
@@ -468,6 +494,16 @@ private void notifyPlotOwner(final PlotPlayer<?> player, final Plot plot, final
     public void logout(UUID uuid) {
         feedRunnable.remove(uuid);
         healRunnable.remove(uuid);
+        playerEffects.remove(uuid);
+    }
+
+    public void addEffect(@NonNull UUID uuid, @NonNull String name, long expiresAt) {
```
NotMyFault, line 500: See my previous comment.

If this is intended to be API, javadocs are needed.

Core/src/main/java/com/plotsquared/core/listener/PlotListener.java
```diff
@@ -484,5 +520,15 @@ private static class Interval {
         }
 
     }
+    private static class StatusEffect {
```
NotMyFault, line 523: Can be record-inized if you don't mind.

Core/src/main/java/com/plotsquared/core/player/PlotPlayer.java
```diff
@@ -1016,6 +1016,13 @@ public boolean equals(final Object obj) {
         return this.lockRepository;
     }
 
+    /**
+     * Removes any effects present of the given type.
+     *
+     * @param name the name of the type to remove
```
NotMyFault, line 1022:
```suggestion
     * @param name the name of the type to remove
     * @since TODO
```

NotMyFault:
Thanks for addressing the outstandign concerns, looks promising so far!

NotMyFault:
What's the status of this PR? The changes requested have been addressed and there is no more blocking feedback that prevents a merge. Shall we include this in next week's release?

NotMyFault:
No further reviews after an explicit request imply this PR is ready to go, thanks for your contribution!

PR #3645, opened 2022-06-01
feature: Add language files to debugpaste

### LINKED ISSUE #3176: Include `/lang/messages_%` in debugpaste
Doing so will make debugging caption and placeholder issues much easier.

Xaver106:
## Overview

Fixes #3176

## Description

This should add all language files to the debugpaste.

### This currently doesn't work!
The response is always, that a *file list* needs to be provided in the request, which is obviously the case. I suspect, that there need to be server-side changes for it to accept those files.

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/DebugPaste.java
+++ b/Core/src/main/java/com/plotsquared/core/command/DebugPaste.java
@@ -78,9 +78,9 @@ public class DebugPaste extends SubCommand {
                 StringBuilder b = new StringBuilder();
                 b.append(
                         """
-                                # Welcome to this paste
-                                # It is meant to provide us at IntellectualSites with better information about your problem
-                                """
+                         # Welcome to this paste
+                         # It is meant to provide us at IntellectualSites with better information about your problem
+                         """
                 );
                 b.append("# PlotSquared Information\n");
                 b.append("PlotSquared Version: ").append(PlotSquared.get().getVersion())


SirYwell (dismissed):
I'd prefer the usage of java nio `Files` and `Path`, and with proper null checks rather than catching NPEs.

Xaver106:
@SirYwell What would be the benefit of `Files` and `Path`?

SirYwell:
> @SirYwell What would be the benefit of `Files` and `Path`?

It's the more modern API, it has better exception handling and less magic null values. In best case, we could get rid of any java.io.File usages altogether, but that's clearly out of scope for this PR.

In this case, the code could be written like 
```java
Path langDir = PlotSquared.platform().getDirectory().toPath().resolve("lang");
if (Files.isDirectory(langDir)) {
    Files.list(langDir)
        .filter(Files::isRegularFile)
        .filter(path -> path.getFileName().startsWith("message_") && path.getFileName().toString().endsWith(".json"))
        .forEach(path -> incendoPaster.addFile(path.toFile()));
```
or something like that (I just wrote it down quickly).

Xaver106:
You never stop learning. Thanks for explaining @SirYwell

Core/src/main/java/com/plotsquared/core/command/DebugPaste.java
```diff
@@ -76,9 +78,9 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
                 StringBuilder b = new StringBuilder();
                 b.append(
                         """
-                         # Welcome to this paste
-                         # It is meant to provide us at IntellectualSites with better information about your problem
-                         """
+                                # Welcome to this paste
+                                # It is meant to provide us at IntellectualSites with better information about your problem
+                                """
```
NotMyFault, line 83: Unneeded change?

NotMyFault (commented):
The change proposed doesn't work at all:
> [P2] Failed to create the debugpaste: request must contain a file list

Core/src/main/java/com/plotsquared/core/command/DebugPaste.java
Xaver106: Looks like Intellij did that while reformatting. Can change it back

SirYwell:
> The change proposed doesn't work at all:

That looks like a bug with the IncendoPaster or the web service.

NotMyFault:
> > The change proposed doesn't work at all:
> 
> That looks like a bug with the IncendoPaster or the web service.

If that is true, this PR must remain on hold until a mandatory PR has been filed against the [Paster repository](https://github.com/IntellectualSites/Paster) to address this issue, if that is actually an issue, or implement it as feature.

Xaver106:
> > > The change proposed doesn't work at all:
> > 
> > 
> > That looks like a bug with the IncendoPaster or the web service.
> 
> If that is true, this PR must remain on hold until a mandatory PR has been filed against the [Paster repository](https://github.com/IntellectualSites/Paster) to address this issue, if that is actually an issue, or implement it as feature.

@NotMyFault That is the repository for the Java library. Is the code of the web service accessible? 
The error code is not from the `Paster` but sent from the Server. The files were all added successfully. I listed all files with `Paster.listFiles()` and were all added correctly.

NotMyFault:
There is https://github.com/IntellectualSites/IncendoPasteViewer I guess?

NotMyFault:
It looks like that is an issue in the IncendoPasteViewer itself: https://github.com/IntellectualSites/IncendoPasteViewer/blob/master/src/main/java/org/incendo/paste/rest/UploadService.java#L95

Xaver106:
Is the server software accesible? The issue seems to be on the server side.

NotMyFault:
> Is the server software accesible? The issue seems to be on the server side.

See
> The web server itself runs on Kvantum: https://github.com/Citymonstret/Kvantum

Xaver106:
@NotMyFault How is the raw code of the web server relevant? As the server only natively serves static content, you either have another software handling incoming files or Kvantum does more than it states in the [Readme](https://github.com/Citymonstret/Kvantum/blob/master/README.md).
And I also don't understand why I'm supposed to debug it. I'm missing vital information to even have a hope of fixing this issue. I have no information on how the server is set up. You need to ask the server owner to debug, not me.

@dordsor21 seems to be involved in the server, maybe he can help. Or if it is true, that kvantum is handling both serving and receiving files, then maybe @Citymonstret can help.

I have done what was asked of me. I implemented the change in PlotSquares code. Everything else is the Problem of IntellectualSites. Right now I feel like I wasted my time on this.

PR #3657, opened 2022-06-07
Ensure forceSync is provided when constructing chunk coordinators

### LINKED ISSUE #3667: Async catcher prevents plot clear
### Server Implementation

Paper

### Server Version

1.19

### Describe the bug

Async catcher prevents plot clearing on 1.19

### To Reproduce

- Create a server on 1.19 using the newest Paper, WorldEdit and P2 build (specific version information can be obtained from the debugpaste below)
- Create a plot world
- Run `/plot clear`

### Expected behaviour

Proceed fine

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://gist.github.com/NotMyFault/e78359794d04f904e38a7af8cb3cb768

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/355edd9b906d4cc9bba7d87ed6b562e0

### PlotSquared Version

6.8.1

### Checklist

### Anything else?

_No response_

dordsor21:
- Fixes asyncCatchOp errors due to chunk coordinators defaulting to forceSync "true"
 - Details that forceSync is *not* necessarily synchronous to the server main thread, and instead sets that enqueue/start methods are synchronous (blocking)
 - Fixes #3667

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/queue/ChunkCoordinatorBuilder.java
+++ b/Core/src/main/java/com/plotsquared/core/queue/ChunkCoordinatorBuilder.java
@@ -58,6 +58,7 @@ public class ChunkCoordinatorBuilder {
     private long maxIterationTime = Settings.QUEUE.MAX_ITERATION_TIME; // A little over 1 tick;
     private int initialBatchSize = Settings.QUEUE.INITIAL_BATCH_SIZE;
     private boolean unloadAfter = true;
+    private boolean forceSync = false;
 
     @Inject
     public ChunkCoordinatorBuilder(@NonNull ChunkCoordinatorFactory chunkCoordinatorFactory) {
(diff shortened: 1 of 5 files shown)


Core/src/main/java/com/plotsquared/core/queue/ChunkCoordinatorBuilder.java
```diff
@@ -227,7 +240,8 @@ public ChunkCoordinatorBuilder(@NonNull ChunkCoordinatorFactory chunkCoordinator
                         this.whenDone,
                         this.throwableConsumer,
                         this.unloadAfter,
-                        this.progressSubscribers
+                        this.progressSubscribers,
+                        this.forceSync
```
NotMyFault, line 244: It would make sense to deprecate the existing builder and create a new method, because it's API, which now takes an additional parameter.

SirYwell (approved):

PR #3660, opened 2022-06-08
Chore: General deprecations

dordsor21:


NotMyFault (commented):
That should likely be done before v7 releases. There will be more stuff we may deprecate, if you are fine with resolving merge conflicts every now and then.

NotMyFault (approved):

SirYwell (approved):

PR #3663, opened 2022-06-09
Default flags using BlockCategories don't throw an error anymore

PierreSchwang:
## Description
Loads all worlds **after** the WorldEdit-Platform is ready to not fail loading default flags using block categories and accessing empty registries.
Hacky workaround for
> 06.06 23:15:39 [Server] WARN com.plotsquared.core.plot.flag.FlagParseException: Failed to parse flag of type 'use'. Value '#signs' was not accepted.
06.06 23:15:39 [Server] WARN [PlotSquared/PlotArea] Failed to parse default flag with key 'use' and value '#signs'. Reason: com.plotsquared.core.configuration.caption.TranslatableCaption@6ebec81d. This flag will not be added as a default flag.

(The `Reason: Class@Hash` is fixed as well) 

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/PlotSquared.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotSquared.java
@@ -73,6 +73,7 @@ import com.plotsquared.core.util.task.TaskManager;
 import com.plotsquared.core.uuid.UUIDPipeline;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.math.BlockVector2;
+import com.sk89q.worldedit.world.block.BlockCategory;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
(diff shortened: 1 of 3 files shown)


SirYwell:
Running the given code later means that the config is saved before the changes are applied - we can't do that this way.

PierreSchwang:
> Running the given code later means that the config is saved before the changes are applied - we can't do that this way.

Good point, haven't thought about that

NotMyFault:
> Loads all worlds after the WorldEdit-Platform is ready to not fail loading default flags using block categories and accessing empty registries.

Is there a test case or step-by-step guide providing sufficient information to replicate this issue in a constant manner?
I fail to reproduce the issue this PR attempts to resolve and don't see why and especially where this would be an issue.

PierreSchwang:
> > Loads all worlds after the WorldEdit-Platform is ready to not fail loading default flags using block categories and accessing empty registries.
> 
> Is there a test case or step-by-step guide providing sufficient information to replicate this issue in a constant manner? I fail to reproduce the issue this PR attempts to resolve and don't see why and especially where this would be an issue.

Set the use flag as a default flag with a block category as a value.

NotMyFault:
> Set the use flag as a default flag with a block category as a value.

That is not a suitable test case. The issue occurs only on plotworlds set as default world in the server.properties file.

PierreSchwang:
To be addressed in v7

PR #3665, opened 2022-06-09
Implement tile entities to generation using Populators

### LINKED ISSUE #3051: Skulls not saving correctly in road schematics
### Server Implementation

Paper

### Server Version

1.16.5

### Describe the bug

**Copy of https://issues.intellectualsites.com/issue/PS-195 by Tannianna**

 Using FAWE or normal Worldedit, creating a road schematic and setting it to generate that schematic does not save/load skulls correctly, they are all Steve.

Tested that this is not an issue with FAWE or normal Worldedit by creating a schematic of a skull, loading and saving the schematic works just fine and the skull is as saved.

### To Reproduce

Place some skulls in the road, create a new schematic using /plot createroadschematic, fly out of the new chunks or using /plot debugroadregen to load that schematic and you'll see all the skulls are Steve.

### Expected behaviour

-

### Screenshots / Videos

grafik

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/d4c53bcd3c6442c0bf237e6b44442b7f

### PlotSquared Version

5.13.0

### Checklist

### Anything else?

_No response_

### LINKED ISSUE #3664: Can't use default flags with block categories for standard world defined by server.properties
### Server Implementation

Paper

### Server Version

1.18.2

### Describe the bug

When using the servers default world as a plot world with a default flag containing a block category (e.g. `#signs`), the PlotArea for that world is loaded before the world itself is loaded.
WorldEdit initializes the registries containing the categories etc after the first WorldInitEvent. Therefor the registries aren't filled at that moment and the default flag parsing throws an error.

### To Reproduce

1. Set the world `world` as the default word in server.properties
2. Use `world` as a plot world
3. Configure the plotworld to have a default use-flag with a block category (e.g. #signs)

### Expected behaviour

The flag should be parsed

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

-

### PlotSquared Version

-

### Checklist

### Anything else?

_No response_

dordsor21:
Sort of fix for #3051 though doesn't support _very_ custom heads that would be placed via other plugins.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
@@ -45,6 +45,9 @@ final class BlockStatePopulator extends BlockPopulator {
 
     private final IndependentPlotGenerator plotGenerator;
 
+    /**
+     * @since TODO
+     */
     public BlockStatePopulator(
             final @NonNull IndependentPlotGenerator plotGenerator
     ) {
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
@@ -78,7 +78,7 @@ final class BlockStatePopulator extends BlockPopulator {
             return;
         }
         LimitedRegionWrapperQueue wrapped = new LimitedRegionWrapperQueue(limitedRegion);
-        // It is possible for the region to be larger than the chunk, but there is not reason for P2 to need to populate
+        // It is possible for the region to be larger than the chunk, but there is no reason for P2 to need to populate
         // outside of the actual chunk area.
         Location min = UncheckedWorldLocation.at(worldInfo.getName(), chunkX << 4, worldInfo.getMinHeight(), chunkZ << 4);
         Location max = UncheckedWorldLocation.at(
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator116.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator116.java
@@ -25,67 +25,54 @@
  */
 package com.plotsquared.bukkit.generator;
 
-import com.plotsquared.bukkit.queue.LimitedRegionWrapperQueue;
+import com.plotsquared.core.PlotSquared;
 import com.plotsquared.core.generator.HybridPlotWorld;
 import com.plotsquared.core.generator.IndependentPlotGenerator;
 import com.plotsquared.core.location.Location;
 import com.plotsquared.core.location.UncheckedWorldLocation;
 import com.plotsquared.core.plot.PlotArea;
-import com.plotsquared.core.plot.world.PlotAreaManager;
 import com.plotsquared.core.plot.world.SinglePlotArea;
+import com.plotsquared.core.queue.QueueCoordinator;
 import com.plotsquared.core.queue.ScopedQueueCoordinator;
+import com.sk89q.worldedit.bukkit.BukkitWorld;
+import com.sk89q.worldedit.util.SideEffectSet;
+import org.bukkit.Chunk;
+import org.bukkit.World;
 import org.bukkit.generator.BlockPopulator;
-import org.bukkit.generator.LimitedRegion;
-import org.bukkit.generator.WorldInfo;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Random;
 
-final class BlockStatePopulator extends BlockPopulator {
+final class BlockStatePopulator116 extends BlockPopulator {
 
     private final IndependentPlotGenerator plotGenerator;
 
-    public BlockStatePopulator(
+    public BlockStatePopulator116(
             final @NonNull IndependentPlotGenerator plotGenerator
     ) {
         this.plotGenerator = plotGenerator;
     }
 
-    /**
-     * @deprecated Use {@link BlockStatePopulator#BlockStatePopulator(IndependentPlotGenerator)} as plotAreManager is unused
-     */
-    @Deprecated(forRemoval = true, since = "TODO")
-    public BlockStatePopulator(
-            final @NonNull IndependentPlotGenerator plotGenerator,
-            final @NonNull PlotAreaManager plotAreaManager
-    ) {
-        this.plotGenerator = plotGenerator;
-    }
-
     @Override
-    public void populate(
-            @NonNull final WorldInfo worldInfo,
-            @NonNull final Random random,
-            final int chunkX,
-            final int chunkZ,
-            @NonNull final LimitedRegion limitedRegion
-    ) {
-        PlotArea area = UncheckedWorldLocation.at(worldInfo.getName(), chunkX << 4, 0, chunkZ << 4).getPlotArea();
+    public void populate(@NotNull final World world, @NotNull final Random random, @NotNull final Chunk source) {
+        int chunkMinX = source.getX() << 4;
+        int chunkMinZ = source.getZ() << 4;
+        PlotArea area = Location.at(world.getName(), chunkMinX, 0, chunkMinZ).getPlotArea();
         if (area == null || (area instanceof HybridPlotWorld hpw && !hpw.populationNeeded()) || area instanceof SinglePlotArea) {
             return;
         }
-        LimitedRegionWrapperQueue wrapped = new LimitedRegionWrapperQueue(limitedRegion);
-        // It is possible for the region to be larger than the chunk, but there is not reason for P2 to need to populate
-        // outside of the actual chunk area.
-        Location min = UncheckedWorldLocation.at(worldInfo.getName(), chunkX << 4, worldInfo.getMinHeight(), chunkZ << 4);
-        Location max = UncheckedWorldLocation.at(
-                worldInfo.getName(),
-                (chunkX << 4) + 15,
-                worldInfo.getMaxHeight(),
-                (chunkZ << 4) + 15
-        );
-        ScopedQueueCoordinator offsetChunkQueue = new ScopedQueueCoordinator(wrapped, min, max);
+
+        QueueCoordinator queue = PlotSquared.platform().globalBlockQueue().getNewQueue(new BukkitWorld(world));
+        queue.setForceSync(true);
+        queue.setSideEffectSet(SideEffectSet.none());
+        queue.setBiomesEnabled(false);
+        queue.setChunkObject(source);
+        Location min = UncheckedWorldLocation.at(world.getName(), chunkMinX, world.getMinHeight(), chunkMinZ);
+        Location max = UncheckedWorldLocation.at(world.getName(), chunkMinX + 15, world.getMaxHeight(), chunkMinZ + 15);
+        ScopedQueueCoordinator offsetChunkQueue = new ScopedQueueCoordinator(queue, min, max);
         this.plotGenerator.populateChunk(offsetChunkQueue, area);
+        queue.enqueue();
     }
 
 }
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator116.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator116.java
@@ -48,6 +48,9 @@ final class BlockStatePopulator116 extends BlockPopulator {
 
     private final IndependentPlotGenerator plotGenerator;
 
+    /**
+     * @since TODO
+     */
     public BlockStatePopulator116(
             final @NonNull IndependentPlotGenerator plotGenerator
     ) {
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
@@ -48,7 +48,7 @@ import org.checkerframework.checker.nullness.qual.NonNull;
  */
 public class LimitedRegionWrapperQueue extends DelegateQueueCoordinator {
 
-    private static final Logger LOGGER = LogManager.getLogger();
+    private static final Logger LOGGER = LogManager.getLogger("PlotSquared/" + LimitedRegionWrapperQueue.class.getSimpleName());
 
     private final LimitedRegion limitedRegion;
 
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
@@ -34,6 +34,8 @@ import com.sk89q.worldedit.function.pattern.Pattern;
 import com.sk89q.worldedit.math.BlockVector3;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockState;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.bukkit.Location;
 import org.bukkit.entity.EntityType;
 import org.bukkit.generator.LimitedRegion;
@@ -45,6 +47,8 @@ import org.checkerframework.checker.nullness.qual.NonNull;
  * @since TODO
  */
 public class LimitedRegionWrapperQueue extends DelegateQueueCoordinator {
+
+    private static final Logger LOGGER = LogManager.getLogger();
 
     private final LimitedRegion limitedRegion;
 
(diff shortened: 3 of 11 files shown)


NotMyFault:
Do you mind outsourcing the fix for #3664 into its own PR? Based on a quick look it doesn't really belong here and I feel it may slow down reviews, because it covers a different topic, as seen on different PRs before.

dordsor21:
What head did you use?

NotMyFault:
> What head did you use?

As the issue shows on the screenshot, I've used a bunch of heads provided by HeadDatabase placed on all four sides of the road on #wool, #leaves, #logs and #ice as base material.

dordsor21:
That's not able to be fixed without going via NMS

NotMyFault:
> That's not able to be fixed without going via NMS. I can remove the error but they will still be steve heads.

Oic, I agree, the hard crash should - if possible - be avoided at all lol

Is there a list or categories of supported TEs I can use to test this PR with?

dordsor21:
The aim would be to support all. There shouldn't be errors for TEs that aren't supported by StateWrapper currently though

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator.java
```diff
@@ -25,53 +25,70 @@
...
+        PlotArea area = UncheckedWorldLocation.at(worldInfo.getName(), chunkX << 4, 0, chunkZ << 4).getPlotArea();
+        if (area == null || (area instanceof HybridPlotWorld hpw && !hpw.populationNeeded()) || area instanceof SinglePlotArea) {
             return;
         }
-        final ChunkWrapper wrap = new ChunkWrapper(area.getWorldName(), source.getX(), source.getZ());
-        final ScopedQueueCoordinator chunk = this.queue.getForChunk(wrap.x, wrap.z,
-                com.plotsquared.bukkit.util.BukkitWorld.getMinWorldHeight(world),
-                com.plotsquared.bukkit.util.BukkitWorld.getMaxWorldHeight(world) - 1
+        LimitedRegionWrapperQueue wrapped = new LimitedRegionWrapperQueue(limitedRegion);
+        // It is possible for the region to be larger than the chunk, but there is not reason for P2 to need to populate
```
NotMyFault, line 81:
```suggestion
        // It is possible for the region to be larger than the chunk, but there is no reason for P2 to need to populate
```

Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
```diff
@@ -0,0 +1,114 @@
...
+import com.sk89q.worldedit.math.BlockVector3;
+import com.sk89q.worldedit.world.block.BaseBlock;
+import com.sk89q.worldedit.world.block.BlockState;
+import org.bukkit.Location;
+import org.bukkit.entity.EntityType;
+import org.bukkit.generator.LimitedRegion;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Wraps a {@link LimitedRegion} inside a P2 {@link com.plotsquared.core.queue.QueueCoordinator} so it can be written to.
```
NotMyFault, line 43:
```suggestion
 * Wraps a {@link LimitedRegion} inside a {@link com.plotsquared.core.queue.QueueCoordinator} so it can be written to.
```

Bukkit/src/main/java/com/plotsquared/bukkit/schematic/StateWrapper.java
```diff
@@ -57,6 +59,10 @@ public class StateWrapper {
     public org.bukkit.block.BlockState state = null;
     public CompoundTag tag = null;
 
+    /**
+     * @deprecated in favour of using WE methods for obtaining NBT
```
NotMyFault, line 63:
```suggestion
     * @deprecated in favour of using WE methods for obtaining NBT
```
For sake of completeness we should link a method or a link to the docs outlining a method to use.

Bukkit/src/main/java/com/plotsquared/bukkit/schematic/StateWrapper.java
```diff
@@ -230,10 +236,34 @@ public boolean restoreTag(@NonNull Block block) {
                 }
                 return false;
             }
+            case "skull" -> {
+                if (state instanceof Skull skull) {
+                    CompoundTag skullOwner = ((CompoundTag) this.tag.getValue().get("SkullOwner"));
```
NotMyFault, line 241:
```suggestion
                    CompoundTag skullOwner = ((CompoundTag) this.tag.getValue().get("SkullOwner"));
```
Do we still support 1.15.2? If so, this needs a special handling, the filed used to be called "Owner" in prior versions.

Bukkit/src/main/java/com/plotsquared/bukkit/schematic/StateWrapper.java
```diff
@@ -230,10 +236,34 @@ public boolean restoreTag(@NonNull Block block) {
...
+                return false;
+            }
         }
         return false;
     }
 
+    /**
+     * @deprecated in favour of using WE methods for obtaining NBT
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 266: See above

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
```diff
@@ -456,8 +490,52 @@ public void addOverlayBiome(short x, short z, BiomeType id) {
...
+    }
+
+    /**
+     * Get if post-generation population of chunks with tiles/entities is needed for this world. Not for public API use.
+     *
+     * @since TODO
+     */
+    @AnnotationHelper.ApiDescription(info = "Internal use only. Subject to changes at any time.")
+    public boolean populationNeeded() {
+        return schem1PopulationNeeded || schem2PopulationNeeded || schem3PopulationNeeded;
```
NotMyFault, line 520: Either we change the signature name and/or invert the return, otherwise the name is quite misleading, as the outcome is always inverted.

Core/src/main/java/com/plotsquared/core/generator/IndependentPlotGenerator.java
```diff
@@ -59,10 +57,15 @@ public abstract class IndependentPlotGenerator {
     public abstract void generateChunk(ScopedQueueCoordinator result, PlotArea settings);
 
     /**
+     * Populates the queue representing a chunk area with tile entities and entities
+     *
+     * @param result   Queue to write to
+     * @param settings PlotArea (settings)
+     * @return True if any population occured
```
NotMyFault, line 64:
```suggestion
     * @return True if any population occurred
```

Core/src/main/java/com/plotsquared/core/location/UncheckedWorldLocation.java
```diff
@@ -0,0 +1,75 @@
...
+import com.sk89q.worldedit.math.BlockVector3;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Used internally for generation to reference locations in worlds that "don't exist yet". There is no guarantee that the world
+ * name provided by {@link UncheckedWorldLocation#getWorldName()} exists on the server.
+ *
+ * @since TODO
+ */
+@AnnotationHelper.ApiDescription(info = "Internal use only. Subject to changes at any time.")
```
NotMyFault, line 38: Probably makes sense to apply that on the public methods too, IJ isn't very generous with propagating class annotations on methods when browsing source jars.

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
```diff
@@ -217,34 +222,170 @@ public void generateChunk(@NonNull ScopedQueueCoordinator result, @NonNull PlotA
...
     }
 
+    @Override
+    public boolean populateChunk(final ScopedQueueCoordinator result, final PlotArea settings) {
+        HybridPlotWorld hybridPlotWorld = (HybridPlotWorld) settings;
+        if (!hybridPlotWorld.populationNeeded()) {
+            return false;
+        }
+        // Coords
+        Location min = result.getMin();
```
NotMyFault, line 252: Any objections about refactoring the new introduced stuff into private methods to reuse them if needed? I feel like we end up copy/pasting more than needed here lol

Bukkit/src/main/java/com/plotsquared/bukkit/schematic/StateWrapper.java
dordsor21: 1.15 can't run java 17 at all

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
dordsor21: The intillij warning is because the only place it's used is in a negation

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
dordsor21: I thought about it but imo it actually reduces the readability of the code given the overall differences, or requirements for multiple return values

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
```diff
@@ -217,34 +222,170 @@ public void generateChunk(@NonNull ScopedQueueCoordinator result, @NonNull PlotA
...
+
+        // The Z-coordinate of a given Z coordinate, relative to the
+        // plot (Counting from the corner with the least positive
+        // coordinates)
+        short[] relativeZ = new short[16];
+        boolean[] insideRoadZ = new boolean[16];
+        boolean[] insideWallZ = new boolean[16];
+        short offsetZ = relativeOffsetZ;
+        while (offsetZ >= hybridPlotWorld.SIZE) {
+            offsetZ -= hybridPlotWorld.SIZE;
```
SirYwell, line 308: Can't this replaced by an if and a modulo operation?

Core/src/main/java/com/plotsquared/core/location/Location.java
```diff
@@ -41,14 +41,17 @@
  * An unmodifiable 6-tuple (world,x,y,z,yaw,pitch)
  */
 @SuppressWarnings("unused")
-public final class Location extends BlockLoc implements Comparable<Location> {
+public class Location extends BlockLoc implements Comparable<Location> {
```
SirYwell, line 44: this should be sealed then.

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
```diff
@@ -456,8 +490,52 @@ public void addOverlayBiome(short x, short z, BiomeType id) {
         this.G_SCH_B.put(pair, id);
     }
 
+    /**
+     * Get the entities contained within the plot schematic for generation. Intended for internal use only.
+     *
+     * @since TODO
```
SirYwell, line 496: Should be possible to hide these methods from generated javadoc with `@hidden`. Not sure if we want that though.

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
```diff
@@ -217,34 +222,170 @@ public void generateChunk(@NonNull ScopedQueueCoordinator result, @NonNull PlotA
...
+                    for (short z = 0; z < 16; z++) {
+                        placeSchem(hybridPlotWorld, result, relativeX[x], relativeZ[z], x, z, true, true);
+                    }
+                }
+            } else if (insideWallX[x]) {
+                if (hybridPlotWorld.ROAD_SCHEMATIC_ENABLED) {
+                    for (short z = 0; z < 16; z++) {
+                        placeSchem(hybridPlotWorld, result, relativeX[x], relativeZ[z], x, z, true, true);
+                    }
+                }
```
SirYwell, line 337: Those two branches contain the same code, can we collapse it?

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
```diff
@@ -217,34 +222,170 @@ public void generateChunk(@NonNull ScopedQueueCoordinator result, @NonNull PlotA
...
+
+        // The X-coordinate of a given X coordinate, relative to the
+        // plot (Counting from the corner with the least positive
+        // coordinates)
+        short[] relativeX = new short[16];
+        boolean[] insideRoadX = new boolean[16];
+        boolean[] insideWallX = new boolean[16];
+        short offsetX = relativeOffsetX;
+        while (offsetX >= hybridPlotWorld.SIZE) {
+            offsetX -= hybridPlotWorld.SIZE;
```
SirYwell, line 282: See comment below about modulo.
Why is overlap not set here?

Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
```diff
@@ -0,0 +1,114 @@
...
+        return result;
+    }
+
+    @Override
+    public boolean setBlock(final int x, final int y, final int z, @NonNull final BlockState id) {
+        try {
+            limitedRegion.setType(x, y, z, BukkitAdapter.adapt(id.getBlockType()));
+            limitedRegion.setBlockData(x, y, z, BukkitAdapter.adapt(id));
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
```
SirYwell, line 81: If we can ignore the exception, this should be removed. If we can't ignore it, it should be logged via a logger instead.

Bukkit/src/main/java/com/plotsquared/bukkit/queue/LimitedRegionWrapperQueue.java
```diff
@@ -0,0 +1,114 @@
...
+    public boolean setEntity(@NonNull final Entity entity) {
+        try {
+            EntityType type = BukkitAdapter.adapt(entity.getState().getType());
+            double x = entity.getLocation().getX();
+            double y = entity.getLocation().getY();
+            double z = entity.getLocation().getZ();
+            Location location = new Location(limitedRegion.getWorld(), x, y, z);
+            limitedRegion.spawnEntity(location, type);
+        } catch (IllegalArgumentException e) {
+            e.printStackTrace();
```
SirYwell, line 97: See above

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
dordsor21: People would still be able to access the method from the API though, so it'd be worth telling them what the method does still tbh

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
dordsor21: Yeah it should be able to be... I had extracted it out of the for loop below and intended to do just that but forgot

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
NotMyFault: -0 for hiding methods from javadocs, we should rather outline and document which methods can and should be used while outlining which stuff is internal or should rather not be used.

Core/src/main/java/com/plotsquared/core/generator/HybridGen.java
dordsor21: Overlap should have only been set in the for loop for the chunk itself, not the while loop outside

NotMyFault (commented):
Otherwise, looks good so far.

NotMyFault (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/generator/BlockStatePopulator116.java
```diff
@@ -0,0 +1,81 @@
...
+import com.sk89q.worldedit.util.SideEffectSet;
+import org.bukkit.Chunk;
+import org.bukkit.World;
+import org.bukkit.generator.BlockPopulator;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Random;
+
+final class BlockStatePopulator116 extends BlockPopulator {
```
SirYwell, line 47: The naming here might be confusing, we have a `BlockEventListener117` and a `PaperListener113` which both declare the "since" while this one declares the "until". Maybe we can change it to `...Pre117`?

dordsor21: Might be better named "LegacyBlockStatePopulator"?

SirYwell: Sure, that works too

SirYwell (approved):

PR #3666, opened 2022-06-09
Only load world cofigurations if WorldEdit has fully enabled (safe to do)

### LINKED ISSUE #3664: Can't use default flags with block categories for standard world defined by server.properties

dordsor21:
- Fixes #3664

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/PlotSquared.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotSquared.java
@@ -1611,6 +1611,7 @@ public class PlotSquared {
 
     private final class WEPlatformReadyListener {
 
+        @SuppressWarnings("unused")
         @Subscribe(priority = EventHandler.Priority.VERY_EARLY)
         public void onPlatformReady(PlatformReadyEvent event) {
             weInitialised = true;
--- a/Core/src/main/java/com/plotsquared/core/PlotSquared.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotSquared.java
@@ -72,7 +72,10 @@ import com.plotsquared.core.util.ReflectionUtils;
 import com.plotsquared.core.util.task.TaskManager;
 import com.plotsquared.core.uuid.UUIDPipeline;
 import com.sk89q.worldedit.WorldEdit;
+import com.sk89q.worldedit.event.platform.PlatformReadyEvent;
 import com.sk89q.worldedit.math.BlockVector2;
+import com.sk89q.worldedit.util.eventbus.EventHandler;
+import com.sk89q.worldedit.util.eventbus.Subscribe;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


NotMyFault:
Using the following setup and having a plotworld set in the server.properties file, I receive the stacktrace still: https://athion.net/ISPaster/paste/view/e1d19d707217430b9c7e57593783c743

dordsor21:
I asked pierre to test and didn't say it didn't work ¯\_(ツ)_/¯

dordsor21:
So the whole queryCapability stuff is just completely useless for seeing if capabilities are actually valid yet

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -1574,6 +1580,13 @@ public void registerCaptionMap(
         return this.plotListener;
     }
 
+    /**
+     * Get if the {@link PlatformReadyEvent} has been sent by WE. There is no way to query this within WE itself.
```
NotMyFault, line 1584:
```suggestion
     * Get if the {@link PlatformReadyEvent} has been sent by WorldEdit. There is no way to query this within WorldEdit itself.
     * @since TODO
```

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -1596,4 +1609,14 @@ public enum SortType {
         DISTANCE_FROM_ORIGIN
     }
 
+    private final class WEPlatformReadyListener {
+
+        @Subscribe(priority = EventHandler.Priority.VERY_EARLY)
```
NotMyFault, line 1614:
```suggestion
        @SuppressWarnings("unused")
        @Subscribe(priority = EventHandler.Priority.VERY_EARLY)
```

NotMyFault (commented):
Otherwise, LGTM, working as expected.

PierreSchwang (approved):
Works perfectly fine for me

NotMyFault (approved):
Addressing my own concern here, this is now good to go.

PR #3673, opened 2022-06-12
refactor: replace usages of switch over keyed enums

SirYwell:
## Description

As [announced by md_5](https://www.spigotmc.org/threads/spigot-bungeecord-1-19.559742/#post-4422235) the API around Keyed enums is planned to change. To prevent issues in future, this PR replaces usages of switch over such enums.

The affected places were identified using the Search Structurally feature from IntelliJ.

There also are a few places which I assume were wrong before (mostly missing breaks/returns), so this should be reviewed carefully.

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -233,10 +233,17 @@ public class BlockEventListener implements Listener {
             plot.debug("Prevented block physics and resent block change because disable-physics = true");
             return;
         }
-        if (event.getChangedType() == Material.COMPARATOR || PHYSICS_BLOCKS.contains(event.getChangedType())) {
+        if (event.getChangedType() == Material.COMPARATOR) {
             if (!plot.getFlag(RedstoneFlag.class)) {
                 event.setCancelled(true);
                 plot.debug("Prevented comparator update because redstone = false");
+            }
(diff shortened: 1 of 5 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -223,48 +233,31 @@ public void onPhysicsEvent(BlockPhysicsEvent event) {
...
-            case DRAGON_EGG:
-            case GRAVEL:
-            case SAND:
-            case TURTLE_EGG:
-            case TURTLE_HELMET:
-            case TURTLE_SPAWN_EGG: {
-                if (plot.getFlag(DisablePhysicsFlag.class)) {
+            return;
+        }
+        if (PHYSICS_BLOCKS.contains(event.getChangedType())) {
```
dordsor21, line 243: this set doesn't have all the blocks that used to be in the switch?

SirYwell: `ANVIL`, `DRAGON_EGG`, `GRAVEL`, and `SAND` should be covered by the `if (event.getChangedType().hasGravity() ...` above, `TURTLE_HELMET` is an item only

dordsor21 (approved):

NotMyFault (approved):
Untested, but looks good.

PR #3681, opened 2022-06-13
Adjust schematic height logic when pasting and actually set air as well

### LINKED ISSUE #3504: Road Schematics are not properly created
### Server Implementation

Paper

### Server Version

1.18.1

### Describe the bug

When creating road(only replicated with roads) schematics with the setting "paste-on-top: false" will not only paste the respective schematic one block higher then it should, it also cuts away the most upper block layer. This might be due to a misconfiguration, as the bedrock is included in each schematic and it should therefore be pasted at y=0 and not y=1 as stated in multiple settings. 

Why you want to rise the road from y=0 to your street height is obvious: Signs and Hologramms will otherwise  generate at y=0 and the schematic WILL be pasted in the right way, but it is not the right way to do things.

Right now you are expected to set the road height to the wall height, so signs (and holos) generate properly. The Road blocks are set to air and "paste-on-top" must be disabled. Then we generate a world, paste the roads and intersections in as described in the documentation, select EVERYTHING and move it one block down. Next we place a Block on each road and intersection, forming a highest block layer consisting of ONLY this block and are now able to select ALL important blocks with FAWE (selecting plots as well doesnt matter, as long as you copy the walls as well as they are part of the schematics as well). Next, we move everything one block down.

NOW, you are able to make a proper schematic with properly aligned signs and holos. This took me almost a week to figure out.

As suggested, the paste level should either be put to y=0, or the schematic should be copied at the respective level and with its full height (which might be an even better solution). Another solution would be an extra option for the sign height, but there may be other solutions as well.

This is just a suggestion after all, but something must be done about the current system - it is not working as expected.

### To Reproduce

1. Create a World with `/p2 setup`, with a road height of 0.
2. Create road schematics as usual with standard settings.
3. Try out worldgen and see signs being shifted, but roads being as expected.
4. Change the setting `paste-on-top` to `false` and change the height of the road to wall-height in `worlds.yml`.
5. Reload PlotSquared and execute just `createroadschematic` on the same model as before.
6. Fly into newly generated chunks and enjoy the choped and 1-Block-too-high roads!

### Expected behaviour

No chopped roads!

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/41142093e1994cffab8bd8e9cb18c11a

### PlotSquared Version

PlotSquared-6.5.0

### Checklist

### Anything else?

As I am just a developer, I am personally interested in this topic, but not affected. As I built this project from source, I do not insist neither push for this change, but I consider it as a valuable bug report nobody else has seen yet. Hope I can help in some way.

dordsor21:
- Should fix #3504

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
@@ -81,6 +81,9 @@ public class HybridPlotWorld extends ClassicPlotWorld {
     public short PATH_WIDTH_UPPER;
     public HashMap<Integer, BaseBlock[]> G_SCH;
     public HashMap<Integer, BiomeType> G_SCH_B;
+    /**
+     * The Y level at which schematic generation will start, lowest of either road or plot schematic generation.
+     */
     public int SCHEM_Y;
     private Location SIGN_LOCATION;
     private File root = null;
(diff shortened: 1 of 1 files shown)


NotMyFault (approved):
Works out fine so far, if I understood the issue correctly.

Minor, non blocking nit: You can drop the `@since` from the deprecation javadoc, it's already part of `@Deprecated`.

NotMyFault:
Causes https://github.com/IntellectualSites/PlotSquared/issues/3712

PR #3691, opened 2022-06-18
Implement build height limits into more events

dordsor21:


Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -846,6 +848,11 @@ public void onGrow(BlockGrowEvent event) {
             return;
         }
 
+        if (location.getY() >= area.getMaxBuildHeight() || location.getY() < area.getMinBuildHeight()) {
```
SirYwell, line 851: Can we introduce a helper method for this somewhere?

SirYwell (changes requested):

SirYwell (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -1068,6 +1083,16 @@ public void onBlockIgnite(BlockIgniteEvent event) {
         Plot plot = area.getOwnedPlot(location1);
         if (player != null) {
             BukkitPlayer pp = BukkitUtil.adapt(player);
+            if (!area.buildRangeContainsY(location1.getY()) && !Permissions
```
NotMyFault, line 1086: Any objections about refactoring this? This is now the sixth time we reuse this message part, this could live on as protected method helper.

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
```diff
@@ -624,6 +624,16 @@ public boolean contains(final @NonNull Location location) {
                 getRegionAbs() == null || this.region.contains(location.getBlockVector3()));
     }
 
+    /**
+     * Get if the PlotArea's build range (min build height -> max build height) contains the given y value
```
NotMyFault, line 628:
```suggestion
     * Get if the {@code PlotArea}'s build range (min build height -> max build height) contains the given y value
```

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
dordsor21: It's sent mostly in BlockEventListener but a protected static method in PlayerEventListener might make more sense since it's specifically for players?

NotMyFault: Based on a quick look over "duplicated methods" in IJ earlier, this is used here and one time in the BlockEventListener117 class.
+1 for adding it to the PlayerEventListener.

PR #3695, opened 2022-06-18
Deprecate the static Permissions "util" class that is almost entirely superseded by methods in PermissionHolder instances themselves

dordsor21:
Given our proclivity for moving away from static-access "util" classes for v7


NotMyFault (approved):
I fear this class can't be removed with ease on a final deprecation cleanup of v7, #hasPermission(caller, permissions) e.g. has more than 100 usages needing to migrate them to the named methods living elsewhere in a different PR, to address the forRemoval part sanely.

PR #3707, opened 2022-06-21
chore: Use WorldEdit logger interface

NotMyFault:
## Description
Use WorldEdit logger implementation, inspired by a recent PR which mistakenly used it.

Untested, because sed goes brrr, however, the format change goes from `PlotSquared/<caller class>` to `FQN.<caller class>`.

## Submitter Checklist


dordsor21 (approved):

SirYwell (changes requested):
The WE class is in an internal package, and while we already use internal API in some places, we should probably keep it to a minimum. In this case, we can simply implement something similar ourselves, making use of the `StackWalker` API.

PR #3709, opened 2022-06-23
Ignore cancelled projectile launch events

### LINKED ISSUE #3592: Flying around with elytra and fireworks

Some way to allow players to fly around with elytra and fireworks anywhere in the plot world. This is something many of our players enjoyed, but has not been possible since we switched to 1.18. Players can't use fireworks to boost themselves anymore while on the road or on other people's plots. I don't know what changed to break this.

It seems booster fireworks are spawned in with a ProjectileLaunchEvent, which causes PlotSquared to cancel the event in ProjectileEventListener. The permissions for projectile launching don't seem fine-grained enough to change this behaviour specifically for booster fireworks.

Perhaps the launch event listener could be modified, so it ignores cancelled events? That is, just slap a `(ignoreCancelled = true)` after the `@EventHandler` tag. Then a simple custom plugin could just cancel the launch event for booster fireworks before PlotSquared receives the event, and then afterwards uncancel the event.

### Anything else?

_No response_

traksag:
## Overview

Fixes #3592

## Description

Changed it so PlotSquared ignores cancelled projectile launch events. Other plugins can then allow projectiles to be launched by cancelling the event, and then later uncancelling the event.

It looks like most of PlotSquared's event listeners use the event priority HIGHEST, but I left that untouched in case any other plugins rely on the current (default) priority.

I could also make it so elytra can be used by anyone anywhere in plot worlds, but I don't know if you guys find that desirable.

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/ProjectileEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/ProjectileEventListener.java
@@ -89,7 +89,7 @@ public class ProjectileEventListener implements Listener {
         }
     }
 
-    @EventHandler
+    @EventHandler(ignoreCancelled = true)
     public void onProjectileLaunch(ProjectileLaunchEvent event) {
         Projectile entity = event.getEntity();
         ProjectileSource shooter = entity.getShooter();


NotMyFault (approved):
Untested.

PR #3717, opened 2022-06-30
Prevent headless pistons deleting walls

### LINKED ISSUE #3716: Pistons can delete plot walls
### Server Implementation

Paper

### Server Version

1.19

### Describe the bug

Sticky pistons can be used to delete plot wall blocks.

Note: happens on 1.18.2 and presumably earlier versions as well.

https://user-images.githubusercontent.com/58818927/176791278-f442055d-3a6c-4d15-b08c-46555df58ed9.mov

### To Reproduce

1. Place sticky piston at the edge of your plot pointing to the road.
2. Turn it into a headless piston using the debug stick (set the extended block property to true).
3. Update the piston by placing a block next to it.
4. Wall block disappears.

### Expected behaviour

I expected the wall block to not disappear.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/d44d61891e3c4dc98c2d26a1ed3e442b

### PlotSquared Version

6.9.1-SNAPSHOT

### Checklist

### Anything else?

_No response_

traksag:
## Overview

Fixes #3716

## Description

Cancel the retract event when a sticky piston faces the plot wall (or into a plot area) and retracts with no blocks attached to it. The effect: the plot wall block doesn't get deleted and the headless sticky piston will turn into a normal sticky piston.

Some things to take note of:
- If a headless sticky piston has more than 1 block in front of it, the piston retract event doesn't seem to fire. In that case the block directly in front of the headless piston also doesn't get deleted. So we don't need to worry about this case.
- Headless non-sticky pistons don't seem to delete the block in front of them. So we can ignore non-stickies. Note that non-stickies have block type `PISTON` instead of `MOVING_PISTON` in this event.
- It is somewhat important to only handle sticky pistons with no blocks in front of them, because `BlockPistonRetractEvent#getDirection` is the direction the blocks move in if the block list is non-empty, and otherwise the facing direction of the piston.

## Submitter Checklist


NotMyFault (approved):
Untested, but LGTM

dordsor21:
A check for the block being outside the min/max build heights of the plot area would also be an idea. See the usages of `PlotArea#buildRangeContainsY` in other events (I'm assuming the same issue will be present for this case as well)

NotMyFault:
@traksag Do you want to address the PR feedback?

traksag:
Sure, I kinda forgot about it. Will probably crack something out over the weekend.

NotMyFault:
> A check for the block being outside the min/max build heights of the plot area would also be an idea. See the usages of `PlotArea#buildRangeContainsY` in other events (I'm assuming the same issue will be present for this case as well)

@Copilot can you address this feedback?

PR #3718, opened 2022-07-03
fix: improve handling of missing message files

### LINKED ISSUE #3411: Don't fail startup if user specifies invalid language code
See https://paste.gg/p/anonymous/eb692c7938dc4eab8a595c8cd2e94fa1 for reference

If a caption is missing, the exception is thrown endlessly but only for one specific missing caption. In this case, the user didn't download the spanish translation, hence it can't find any caption for 'es', because only 'en' is present.
That specific case could be handled separately as well; if no translation file for a caption is present, don't try to lookup captions at all, because you can't find them anyway.

SirYwell:
## Overview

Fixes #3411

## Description

This is basically a two step protection against issues with loading of message files:
1. If the file does not exist, we create a new one and copy the default values into it (will be english only currently).
2. If any other exception occurs during loading (e.g. missing permissions to create/load the file, race conditions on file system level etc) we stop the server to prevent annoying damage to the world.

## Submitter Checklist


NotMyFault (approved):

NotMyFault:
If there's no further demand to review this PR, I'll include it in next week's release.

PR #3722, opened 2022-07-05
Revert #3681

NotMyFault:
#3681 caused a bunch of regressions and improper created road schematics.

Evidently the PR lacked proper testing in production environments, I'm reverting it. The underlaying issue needs to be addressed properly.


NotMyFault:
I tested your PR with ##wool as 1 layer pattern, but apparently that wasn't enough to trigger the regression :/

PR #3729, opened 2022-07-16
Allow pistons to break blocks at borders

CommandDan:
## Description

Pistons are now able to break blocks when pushing against a plot border or plot area border

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -69,6 +69,7 @@ import org.bukkit.Tag;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.block.PistonMoveReaction;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Fireball;
@@ -828,6 +829,7 @@ public class BlockEventListener implements Listener {
             }
             for (Block block1 : event.getBlocks()) {
                 Location bloc = BukkitUtil.adapt(block1.getLocation());
+                if (block1.getPistonMoveReaction() == PistonMoveReaction.BREAK) return;
                 if (bloc.isPlotArea() || bloc
                         .add(relative.getBlockX(), relative.getBlockY(), relative.getBlockZ())
                         .isPlotArea()) {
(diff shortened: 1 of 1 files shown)


NotMyFault (changes requested):
The plot border is meant to be immutable by in game edits, unless you have certain permission.
This change circumvenst the intended limitation.

What is the purpose of this PR? 

**Blocking review**

CommandDan:
It does not affect the plot border. Nothing will be pushed outside the plot area as the block is broken when being pushed.

This change fixes the current issue that breakable blocks, in my situation sugarcane, next to the border can not be broken by pistons facing against the border.

It is currently possible to break these blocks using pistons as long as the breakable block is not the last in the row.

CommandDan:
Here are previews of the change
[Current behaviour](https://imgur.com/ENPTLl3)
[New behaviour](https://imgur.com/0jtm9C2)
[Top Bottom comparison](https://imgur.com/a/oc8WZHV)

CommandDan:
Just realised where my change has a flaw. Am going on camping vacation so I won't be able to make the changes in the coming 2 weeks but will look at fixing it when I am back home.

But yea I see it would have an issue if the border contains a breakable block

CommandDan:
Oops forgot to remove comments

CommandDan:
Should be good now

NotMyFault:
The change looks good now, but justification is needed still. Items dropped on the last block on the plot inevetibly land on the border or road, which possibly leads to unwanted side effects for servers.
More feedback is needed.

CommandDan:
Yea that is true but wouldn't that be the case either way? I guess, if not working like that already, we can implement so blocks breaking keep their drops inside the plot or outside the area

RedstoneFuture:
I like the idea - thanks [CommandDan](https://github.com/CommandDan)!

> The change looks good now, but justification is needed still. Items dropped on the last block on the plot inevetibly land on the border or road, which possibly leads to unwanted side effects for servers. More feedback is needed.

Dropping items will not be the problem, I think. You can use the plot and way flags and the `kill-road-items` option.

CommandDan:
Np. And yeah that is what I meant for both points

CommandDan:
> The change looks good now, but justification is needed still. Items dropped on the last block on the plot inevetibly land on the border or road, which possibly leads to unwanted side effects for servers.
> More feedback is needed.

What would you want to be done regards to the items?

If anything should be done

PierreSchwang:
> > The change looks good now, but justification is needed still. Items dropped on the last block on the plot inevetibly land on the border or road, which possibly leads to unwanted side effects for servers.
> > More feedback is needed.
> 
> What would you want to be done regards to the items?
> 
> If anything should be done

I'd say, if `kill-road-items` is set to `true` in the config, kill the item when dropping onto the road / border.
Not sure if the position of the final item can be predicted tbh - could be quite hard.

Otherwise, what might be unsafe, is dropping the item based on the first solid block underneath the piston retraction (Using dropItem instead of dropItemNaturally). Basically, accessing the Blocks from the event, calling getDrops on those and drop these in the middle of the lowest block.
But feedback regarding that would be highly appreciated, as that solution isn't the best probably.

CommandDan:
Yea, I can add the check for kill-road-items

PR #3735, opened 2022-07-22
Add PlayerPlotChatEvent

jaapieaapie1:
## Description
This pull request adds a new event that will be called after a player chats in plot chat.

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -778,6 +778,8 @@ public class PlayerEventListener implements Listener {
                     spymessageTemplate
             );
         }
+
+        this.eventDispatcher.callChat(plotPlayer, plot, message, plotRecipients, spies);
     }
 
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
--- a/Core/src/main/java/com/plotsquared/core/events/PlayerPlotChatEvent.java
+++ b/Core/src/main/java/com/plotsquared/core/events/PlayerPlotChatEvent.java
@@ -1,0 +1,67 @@
+package com.plotsquared.core.events;
+
+import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.Plot;
+
+import java.util.Set;
+
+/**
+ * @since TODO
+ */
+public class PlayerPlotChatEvent extends PlotEvent {
+
+    private final PlotPlayer<?> player;
+    private final String message;
+    private final Set<PlotPlayer<?>> recipients;
+    private final Set<PlotPlayer<?>> spies;
+
+    public PlayerPlotChatEvent(
+            PlotPlayer<?> player,
+            Plot plot,
+            String message,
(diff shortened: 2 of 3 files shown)


PierreSchwang:
Looks good so far - Would be great to implement the [CancellablePlotEvent](https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/events/CancellablePlotEvent.java) as well. Therefor plugins using that event are able to force / cancel plot messages on their own, if that event already exist.

Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
```diff
@@ -241,6 +243,22 @@ public PlayerLeavePlotEvent callLeave(PlotPlayer<?> player, Plot plot) {
         return event;
     }
 
+    /**
+     * @since TODO
+     */
```
NotMyFault, line 248:
```suggestion
```
This is an internal class.

Core/src/main/java/com/plotsquared/core/events/PlayerPlotChatEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+package com.plotsquared.core.events;
+
+import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.Plot;
+
+import java.util.Set;
+
+/**
+ * @since TODO
```
NotMyFault, line 27: Please add since todo to the added methods too.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -732,6 +730,22 @@ public void onChat(AsyncPlayerChatEvent event) {
...
+
+        PlayerPlotChatEvent plotChatEvent = this.eventDispatcher.callChat(plotPlayer, plot, message, plotRecipients, spies);
+
+        if (plotChatEvent.getEventResult() == Result.DENY) {
+            return;
+        }
+
+        boolean force = plotChatEvent.getEventResult() == Result.FORCE;
+
+        if (force || (plot.isDenied(plotPlayer.getUUID()) && !Permissions
```
SirYwell, line 742: I think this should be `!force`, as the plot chat should happen if the result is `FORCE`

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/events/PlayerPlotChatEvent.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public class PlayerPlotChatEvent extends PlotEvent implements CancellablePlotEvent {
+
+    private final PlotPlayer<?> player;
+    private final String message;
```
PierreSchwang, line 32: Maybe rather use an adventure component for the message?
The passed message as a string is before the translation of color codes etc happened. Would be more suitable to use the component which is used as the "msg" placeholder.

PierreSchwang (changes requested):

NotMyFault:
@jaapieaapie1 Would you like to address the feedback?

PR #3746, opened 2022-07-28
Don't allow wall corals to dry

traksag:
## Overview

Currently wall corals dry up when the plot flag `coral-dry` is set to false (the default). Simply get yourself a coral fan, place it on the side of a wall and wait for it to dry up (should happen within a couple of seconds).

## Description

The plugin didn't consider wall corals as corals (they don't fall under the two used block tags). Also check for the `WALL_CORALS` block tag to fix this.

## Submitter Checklist


NotMyFault (approved):
Good catch

PierreSchwang (approved):

PR #3749, opened 2022-07-29
perf: avoid quadratic overhead for UUID collection

SirYwell:
## Description

The previous way to collect UUID's into the queue caused long startup times when having a lot of plots. By putting all UUID's in a set first, we can go from quadratic to linear complexity.

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
@@ -656,20 +656,15 @@ public final class BukkitPlatform extends JavaPlugin implements Listener, PlotPl
             final @NonNull SQLiteUUIDService sqLiteUUIDService,
             final @NonNull CacheUUIDService cacheUUIDService
     ) {
-        // Load all uuids into a big chunky boi queue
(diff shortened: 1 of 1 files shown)


PierreSchwang:
Thought the database querying was the critical point, which took most of the time. Do you have any comparisons regarding startup time? Maybe by seeding the database with a few thousand plots & uuids? I'd be highly interested in that

SirYwell:
I generated user data with 10000 and 50000 entries and started the server wih each of them 3 times and wrote down the time in the `Done` message.

Baseline
| UUID count | startup time avg |
| ------------ | ----------------- |
| 10000         | 5.576s                |
| 50000         | 12.8s                  |

Patch
| UUID count | startup time avg |
| ------------ | ----------------- |
| 10000         | 5.633s                |
| 50000         | 5.795s                |

For fun, I also generated a file with 250000 entries. With that, the patched version takes `7.574s` and the baseline takes `515.268s`...

NotMyFault (approved):
Untested, but sounds reasonable.

PierreSchwang (approved):
I'm gonna trust you, that this works 👀

PR #3819, opened 2022-10-01
Only show visible players in `/plot near`

### LINKED ISSUE #3800: /plot near should show only visible players (vanish)

/plot near command does not respect vanish plugins, so players can see if they are being watched by staff members.
i think this command should check and display players only if command sender can see them.

-

### Anything else?

_No response_

PierreSchwang:
## Overview
Fixes #3800

## Description
Checks for the visibility when running `/plot near` (respects "vanish" plugins etc)

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/Plot.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/Plot.java
@@ -457,7 +457,7 @@ public class Plot {
      * that could alter the de facto owner of the plot.
      *
      * @return The plot owner of this particular (sub-)plot
-     * as stored in the database, if one exists. Else, null.
+     *         as stored in the database, if one exists. Else, null.
      */
     public @Nullable UUID getOwnerAbs() {
         return this.owner;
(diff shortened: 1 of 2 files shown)


NotMyFault (approved):
Untested, but looks reasonable.

Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -554,6 +555,16 @@ public void deleteMeta(final @NonNull String key) {
         return players;
     }
 
+    /**
+     * @param executor The player for whom the players in the plot should be retrieved
+     * @return list of visible PlotPlayer(s) for the executor inside of {@code this} plot
+     * @see #getPlayersInPlot()
+     * @since TODO
+     */
+    public @NonNull List<PlotPlayer<?>> getPlayersInPlotVisible(@NonNull PlotPlayer<?> executor) {
```
SirYwell, line 564: Not sure if a public method in `Plot` is worth it. I'd prefer to have a (private) method in the command as long as it's only used there.

Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -554,6 +555,16 @@ public void deleteMeta(final @NonNull String key) {
...
+        return getPlayersInPlot().stream().filter(executor::canSee).collect(Collectors.toList());
```
SirYwell, line 565:
```suggestion
        return getPlayersInPlot().stream().filter(executor::canSee).toList();
```

SirYwell (changes requested):

SirYwell (approved):

PR #3829, opened 2022-10-06
Don't kill owned road entities (leashed / named)

### LINKED ISSUE #3824: Named Animals or Ridden/Leashed horses get killed on road
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

Animals, Mobs, Horses etc. are being killed on road when they enter the road, even when the configs say they dont. They just dont get removed when they are actually ridden (horses).

### To Reproduce

Create World, Make the Setup (see attachment of "settings.yml") and yea thats it.

### Expected behaviour

Mobs not being removed when leashed or named.

### Screenshots / Videos

https://gyazo.com/fdaa63976eff81e64f001ef0c97e7b3a

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/3b9f5aad8b5c409382b59b6876bf7fda

### PlotSquared Version

6.10.0-Premium

### Checklist

### Anything else?

_No response_

BlockyTheDev:
## Overview

Fixes #3824

## Description
Fixes a bug, where leashed entities got killed on the road, with "KILL_ROAD_MOBS" set to true and "KILL_OWNED_ROAD_MOBS" set to false. This PR restores the behaviour described in the settings comment at the value "KILL_OWNED_ROAD_MOBS".

## Submitter Checklist

- [ ] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [ ] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md).

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
@@ -759,6 +759,8 @@ public class Settings extends Config {
         @Comment("Also kill any road mobs that are being ridden, or are leashed")
         public static boolean
                 KILL_OWNED_ROAD_MOBS = false;
+        @Comment("Also kill any road mobs that are being named")
+        public static boolean KILL_NAMED_ROAD_MOBS = false;
         @Comment("Kill items on roads (Stick, Paper, etc.)")
         public static boolean KILL_ROAD_ITEMS = false;
         @Comment("Kill vehicles on roads (Boat, Minecart, etc.)")
--- a/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
+++ b/Core/src/main/java/com/plotsquared/core/configuration/Settings.java
@@ -759,7 +759,7 @@ public class Settings extends Config {
         @Comment("Also kill any road mobs that are being ridden, or are leashed")
         public static boolean
                 KILL_OWNED_ROAD_MOBS = false;
-        @Comment("Also kill any road mobs that are being named")
+        @Comment("Also kill any road mobs that are named")
         public static boolean KILL_NAMED_ROAD_MOBS = false;
         @Comment("Kill items on roads (Stick, Paper, etc.)")
         public static boolean KILL_ROAD_ITEMS = false;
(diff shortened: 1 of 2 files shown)


PierreSchwang:
Would be great if you could adress the 2nd part of that issue as well ("Mobs not being removed when leashed **or named**") - Otherwise that issue would only be partly solved

BlockyTheDev:
> Would be great if you could adress the 2nd part of that issue as well ("Mobs not being removed when leashed **or named**") - Otherwise that issue would only be partly solved

Ok. I will do it tomorrow. I didn't added the name thing, because I oriented me on the settings comment. Should I add the thing with the name to the settings comment, too?

BlockyTheDev:
Or should there be a extra setting for named items?

The problem, I am seeing, when the named thing is the same setting like the leashed and ridden one, is primarily on servers where nametags are available in a large number. So players can bring very much entities to the road. This is not always wanted, because players could bypass things like plot caps with things like villagers by adding a nametag to them and moving them to the road and switching them after trading, so they can have much more, as planned.

I hope I wrote it in an understandable way.

BlockyTheDev:
> > Or should there be a extra setting for named items?
> > The problem, I am seeing, when the named thing is the same setting like the leashed and ridden one, is primarily on servers where nametags are available in a large number. So players can bring very much entities to the road. This is not always wanted, because players could bypass things like plot caps with things like villagers by adding a nametag to them and moving them to the road and switching them after trading, so they can have much more, as planned.
> > I hope I wrote it in an understandable way.
> 
> Good point - I'd say you are free to decide that. You can add a new config option for that or keep it as it is

I added a setting to kill named mobs. In addition, I added a missing check on shulkers, where shulkers are removed when they are leashed but the setting `KILL_OWNED_ROAD_MOBS` is set to false.

Core/src/main/java/com/plotsquared/core/configuration/Settings.java
```diff
@@ -759,6 +759,8 @@ public static final class Enabled_Components { // Group the following values int
         @Comment("Also kill any road mobs that are being ridden, or are leashed")
         public static boolean
                 KILL_OWNED_ROAD_MOBS = false;
+        @Comment("Also kill any road mobs that are being named")
```
dordsor21, line 762:
```suggestion
        @Comment("Also kill any road mobs that are named")
```

dordsor21 (approved):

NotMyFault (approved):

PR #3841, opened 2022-10-09
feat: add cost for the merge when using plot auto-merge

### LINKED ISSUE #3814: /p auto 2,2 economy no merge costs
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

/p auto 2,2 economy no merge costs 

### To Reproduce

have 
```
    economy:
      prices:
        merge: 500
```
make /p auto 2,2
No money will be deducted

### Expected behaviour

The money is deducted

### Screenshots / Videos

_No response_

### Error log (if applicable)

no error

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/fdc556d1985e402697e376b9369bfe4b

### PlotSquared Version

6.10.1 Build from Source

### Checklist

### Anything else?

_No response_

dordsor21:
- Closes #3814


NotMyFault (approved):
Untested

Core/src/main/java/com/plotsquared/core/command/Auto.java
```diff
@@ -293,10 +293,12 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
         }
         if (this.econHandler != null && plotarea.useEconomy()) {
             PlotExpression costExp = plotarea.getPrices().get("claim");
+            PlotExpression mergeCostExp = plotarea.getPrices().get("merge");
+            double mergeCost = mergeCostExp == null ? 0d : mergeCostExp.evaluate(sizeX * sizeZ);
             double cost = costExp.evaluate(Settings.Limit.GLOBAL ?
                     player.getPlotCount() :
                     player.getPlotCount(plotarea.getWorldName()));
-            cost = (sizeX * sizeZ) * cost;
+            cost = (sizeX * sizeZ) * cost + mergeCost;
```
PierreSchwang, line 301: Not tested yet, but based on my quick read: Wouldn't the current implementation charge for a merge, even if no merge has to be made (sizeX = sizeZ = 1).
Either way, the current implementation should be modified either way, as linear costs (e.g. `if (plots == 0) {return 0;} else { result = 4000; for (cnt = 2, plots) result *= 2; return result;}`) wouldn't be respected. The cost for each plot should be calculated on their own and added in the end. Same goes for (eventual) merges of course.

dordsor21: It would currently charge for a merge, even if the PlotMergeEvent is cancelled, and it would charge for a single plot...

However, RE: the second part, the implementation for how merge cost is calculated is kept the same as a normal `/plot merge`, and change to this would be out of scope

dordsor21:
@PierreSchwang changes made

SirYwell (approved):

PR #3845, opened 2022-10-10
feat: Move ExpireManager to Guice by removing legacy IMP references

### LINKED ISSUE #3358: Move `ExpireManager` to Guice

The `ExpireManager` still uses the legacy `IMP` static to get an instance of the class, PlotSquared v6 got a rid of it everywhere and uses Guice's inject to get an implementation of the class. Looks like the expire manager has been forgotten about but should be ported as well.

:)

### Anything else?

_No response_

michizhou:
## Overview

Closes #3358 

## Description

This PR adds a new method to `PlotPlatform` that allows it to use Guice's inject to get an implementation of the class and replaces all references to `ExpireManager.IMP` accordingly. Instances of `ExpireManager.IMP` have also been replaced with appropriate instances of `ExpireManager` to avoid confusion with the `PlotPlatform` method.

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/PlotPlatform.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotPlatform.java
@@ -32,6 +32,7 @@ import com.plotsquared.core.inject.annotations.DefaultGenerator;
 import com.plotsquared.core.location.World;
 import com.plotsquared.core.permissions.PermissionHandler;
 import com.plotsquared.core.player.PlotPlayer;
+import com.plotsquared.core.plot.expiration.ExpireManager;
 import com.plotsquared.core.plot.world.PlotAreaManager;
 import com.plotsquared.core.queue.GlobalBlockQueue;
 import com.plotsquared.core.util.ChunkManager;
@@ -282,6 +283,15 @@ public interface PlotPlatform<P> extends LocaleHolder {
      */
     default @NonNull ChunkManager chunkManager() {
         return injector().getInstance(ChunkManager.class);
+    }
+
+    /**
+     * Get the {@link ExpireManager} implementation for the platform
+     *
+     * @return Expire manager
+     */
+    default @NonNull ExpireManager expireManager() {
+        return injector().getInstance(ExpireManager.class);
     }
 
     /**
--- a/Core/src/main/java/com/plotsquared/core/PlotPlatform.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotPlatform.java
@@ -289,6 +289,7 @@ public interface PlotPlatform<P> extends LocaleHolder {
      * Get the {@link ExpireManager} implementation for the platform
      *
      * @return Expire manager
+     * @since TODO
      */
     default @NonNull ExpireManager expireManager() {
         return injector().getInstance(ExpireManager.class);
--- a/Core/src/main/java/com/plotsquared/core/PlotSquared.java
+++ b/Core/src/main/java/com/plotsquared/core/PlotSquared.java
@@ -290,11 +290,11 @@ public class PlotSquared {
 
     public void startExpiryTasks() {
         if (Settings.Enabled_Components.PLOT_EXPIRY) {
-            ExpireManager manager = new ExpireManager(this.eventDispatcher);
-            manager.runAutomatedTask();
+            ExpireManager expireManager = PlotSquared.platform().expireManager();
+            expireManager.runAutomatedTask();
             for (Settings.Auto_Clear settings : Settings.AUTO_CLEAR.getInstances()) {
                 ExpiryTask task = new ExpiryTask(settings, this.getPlotAreaManager());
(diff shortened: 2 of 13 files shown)


Core/src/main/java/com/plotsquared/core/plot/expiration/ExpireManager.java
```diff
@@ -60,7 +60,6 @@
 
 public class ExpireManager {
 
-    public static ExpireManager IMP;
```
NotMyFault, line 63:
```suggestion
/**
* @deprecated Use {@link PlotPlatform#expireManager()} instead
*/
@Deprecated(forRemoval = true, since = "TODO")
public static ExpireManager IMP;
```

This is public API and can't be removed without causing possible regressions for third party consumers.

The import for the class for the javadoc is needed still.

Core/src/main/java/com/plotsquared/core/PlotPlatform.java
```diff
@@ -284,6 +285,15 @@
         return injector().getInstance(ChunkManager.class);
     }
 
+    /**
+     * Get the {@link ExpireManager} implementation for the platform
+     *
+     * @return Expire manager
```
NotMyFault, line 291:
```suggestion
     * @return Expire manager
     * @since TODO
```

NotMyFault (commented):
Untested, but looks good so far. A few minor nits are up to address, but thanks for your PR!

Core/src/main/java/com/plotsquared/core/plot/expiration/ExpireManager.java
```diff
@@ -60,6 +60,10 @@
 
 public class ExpireManager {
 
+    /**
+     * @deprecated Use {@link PlotPlatform#expireManager()} instead
```
dordsor21, line 64: Build is failing here `error: reference not found`. Ensure it's imported or is referenced by FQN

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -290,11 +290,11 @@ public void loadCaptionMap() throws Exception {
 
     public void startExpiryTasks() {
         if (Settings.Enabled_Components.PLOT_EXPIRY) {
-            ExpireManager.IMP = new ExpireManager(this.eventDispatcher);
-            ExpireManager.IMP.runAutomatedTask();
+            ExpireManager manager = new ExpireManager(this.eventDispatcher);
```
dordsor21, line 293: Should this not be accessing the expiry manager from guice?

Core/src/main/java/com/plotsquared/core/PlotSquared.java
```diff
@@ -645,7 +645,7 @@ private void sortPlotsByHash(final @NonNull Plot @NonNull [] input) {
         } else {
             list = new ArrayList<>(input);
         }
-        list.sort(Comparator.comparingLong(a -> ExpireManager.IMP.getTimestamp(a.getOwnerAbs())));
+        list.sort(Comparator.comparingLong(a -> PlotSquared.platform().expireManager().getTimestamp(a.getOwnerAbs())));
```
SirYwell, line 648: We've had performance issues with guice before (https://github.com/IntellectualSites/PlotSquared/pull/3315), so I think frequent accesses should be avoided, e.g.
```suggestion
        ExpireManager expireManager = PlotSquared.platform().expireManager();
        list.sort(Comparator.comparingLong(a -> expireManager.getTimestamp(a.getOwnerAbs())));
```

Core/src/main/java/com/plotsquared/core/plot/expiration/ExpiryTask.java
```diff
@@ -73,7 +74,7 @@ public boolean applies(PlotArea area) {
                         diff = plots.size() - settings.REQUIRED_PLOTS;
                     }
                     List<Long> entireList =
-                            plots.stream().map(plot -> ExpireManager.IMP.getAge(plot, settings.DELETE_IF_OWNER_IS_UNKNOWN))
+                            plots.stream().map(plot -> PlotSquared.platform().expireManager().getAge(plot, settings.DELETE_IF_OWNER_IS_UNKNOWN))
```
SirYwell, line 77: See my other comment

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/PlotSquared.java
michizhou: @dordsor21 I didn't want to call the `ExpireManager` from guice since I don't believe it allows for `eventDispatcher` to be explicitly set, which could impact `runAutomatedTask()`. If this isn't an actual concern, then I'll can revert this change or call `PlotSquared.platform().expireManager()` here instead.

SirYwell (approved):
Looks good.

PR #3854, opened 2022-10-23
fix: account for mismatched road-schematic heights

dordsor21:
Intersection and sideroad schematics can be of different heights. Ensure we get the maximum of the two for road schematic height


SirYwell (approved):

PierreSchwang (approved):
pog

NotMyFault (approved):
Untested

PR #3870, opened 2022-11-12
feature: implement road flags as widely as seems reasonable

### LINKED ISSUE #3047: Allow blocked actions on roads and unowned

Some actions, like throwing a splash potion, using a fireworkrocket or shooting a bow, are blocked on roads and unowned plots. There is no flag or bypass permission to allow players to use it and not even a message to let the player now, that Plotsquared blocked the action.
Some of these actions, like shooting a bow, work on plots you are not added/trusted on, even if the arrows dissapear on the ground.
Most of them can be bypassed with admin permissions, but splash potions can't (see: #2986). 

A bypass permission for all of these would be nice.

No. Plotsquared just blocks certain actions on roads/unowned regardless of the permissions/flags.

### Anything else?

_No response_

dordsor21:
- Closes #3047
 - Any further road flag implementations would be very specific from what I can gather, and thus the gist of #3047 is now implemented

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
@@ -46,7 +46,6 @@ import com.plotsquared.core.plot.flag.FlagParseException;
 import com.plotsquared.core.plot.flag.GlobalFlagContainer;
 import com.plotsquared.core.plot.flag.PlotFlag;
 import com.plotsquared.core.plot.flag.implementations.DoneFlag;
-import com.plotsquared.core.plot.flag.types.BooleanFlag;
 import com.plotsquared.core.queue.GlobalBlockQueue;
 import com.plotsquared.core.queue.QueueCoordinator;
 import com.plotsquared.core.util.MathMan;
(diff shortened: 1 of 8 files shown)


Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
```diff
@@ -1242,6 +1243,18 @@ public <T> T getRoadFlag(final Class<? extends PlotFlag<T, ?>> flagClass) {
         return this.roadFlagContainer.getFlag(flagClass).getValue();
     }
 
+    /**
+     * Check if the value of a {@link BooleanFlag} matches the given boolean. If
+     * road flags are disabled, returns false.
+     *
+     * @param flagClass boolean flag to get value of
+     * @param value     boolean value to check flag value against
+     * @return Flag value or false if road flags disabled
```
NotMyFault, line 1252:
```suggestion
     * @return Flag value or false if road flags disabled
     * @since TODO
```

NotMyFault (approved):

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
```diff
@@ -1242,6 +1243,19 @@ public <T> T getRoadFlag(final Class<? extends PlotFlag<T, ?>> flagClass) {
...
+     * @since TODO
+     */
+    public boolean isRoadFlagsAndFlagEquals(final Class<? extends BooleanFlag<?>> flagClass, boolean value) {
```
SirYwell, line 1255: This method is somewhat specific, could it be moved to some (non-API) util class instead?

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
dordsor21: There aren't currently any suitable util classes that I can find

SirYwell: I guess we can add a new one, like `PlotFlagSupport` or whatever, and mark it as internal? Maybe we can also extract some more code related to flags in future.

PR #3871, opened 2022-11-14
feat: allow the world border to be larger than the road surrounding the outermost plots

### LINKED ISSUE #2962: Wrap world border around furthest plot + 1
Currently, the world border wraps around the furthest plot. Players who own plots on the border are unable to "take a step back" and look at their plot from a distance because they can't fly beyond the border.

A solution to this problem is to wrap the border around the furthest plot, plus 1, allowing unobstructed movement around the edge of the entire world.

In each direction, the world border should wrap around the furthest plot + 1.
The outermost ring of plots should be unclaimable* as in the current system. This prevents users from expanding the world border arbitrarily by abusing this change. The only difference should be that users are now able to fly a short distance beyond the world border.

**Additional context**
\* As an additional feature, the outermost ring of "unclaimable" plots could be made claimable by users with a certain admin permission. This allows the creation of merged/megaplots at the world border, which was not possible before.

**Checklist**:

- [X] I made sure there are no duplicates of my suggestion [(Use search)](https://github.com/IntellectualSites/PlotSquaredSuggestions/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+)
- [X] I made sure my suggestion does not exist yet [(Check the wiki)](https://wiki.intellectualsites.com/plotsquared/home)
- [X] I made sure my suggestion is useful for the majority of PlotSquared users

dordsor21:
- Closes #2962

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -596,7 +596,7 @@ public class PlayerEventListener implements Listener {
                 this.tmpTeleport = true;
                 return;
             }
-            int border = area.getBorder();
+            int border = area.getBorder(true);
             int x1;
             if (x2 > border && this.tmpTeleport) {
                 if (!Permissions.hasPermission(pp, Permission.PERMISSION_ADMIN_BYPASS_BORDER)) {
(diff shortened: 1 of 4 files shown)


NotMyFault (approved):
Since annotations would be nice to have

SirYwell (changes requested):
I think this is confusing on multiple levels:

- the effects of the two different border "types" are unclear. What's their meaning? What's their behavior?
- the getExt(r)aBorder method returns the value that is added, not the result of the addition.

Besides that:

- the extra border value should always be nonnegative

I think it would make more sense to have e.g. an enum `BorderType` with constants `CLAIMABLE` and `WALKABLE` (or something like that). The config representation of that doesn't need to be exposed like that, in the API it could just look like two different border types (without public getter for that extra value).

Issue #3875, opened 2022-11-18
Players cannot claim plots on PlotSquared-Bukkit-6.10.4-Premium, if no merge permissions are set


SecretlyJealous:
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

I noticed that new players couldn't claim a plot when I updated to PlotSquared-Bukkit-6.10.4-Premium, it just gives the error "You are lacking the permission node plots.merge.1" For some reason the plugin thinks claiming a new plot means you're trying to perform a plot merge?

![2022-11-18_12 38 14](https://user-images.githubusercontent.com/10485646/202706770-e4260f37-1574-487f-b8d8-ca7e7c708640.png)

I think this commit is causing the issue but I could be wrong: https://github.com/IntellectualSites/PlotSquared/commit/25ce7a83f134921c1fc2c69f5d92ecb23edf1761

### To Reproduce

1. Give players the permission for plots.auto
2. Try to claim a plot with /plot auto

### Expected behaviour

A new plot should be claimed by the user

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/907a608b7f1549378468c275c8f51bb8

### PlotSquared Version

PlotSquared version 6.10.4-Premium

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

TruaYT:
PLEASE FIX!!!!!!!!

PR #3876, opened 2022-11-20
Revert "feat: check merge limits when completing auto-merge (#3868)" - Fixes a bug in 6.10.4

NotMyFault:
The prior PR merged only works if the actor has basic merge permissions, which may not always be set.

Closes https://github.com/IntellectualSites/PlotSquared/issues/3875

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Auto.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Auto.java
@@ -142,14 +142,6 @@ public class Auto extends SubCommand {
                 }
             }
         }
-        int maxMerge = Permissions.hasPermissionRange(player, Permission.PERMISSION_MERGE, Settings.Limit.MAX_PLOTS);
-        if (sizeX * sizeZ > maxMerge) {
-            player.sendMessage(
-                    TranslatableCaption.of("permission.no_permission"),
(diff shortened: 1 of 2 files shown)


Core/src/main/java/com/plotsquared/core/command/Auto.java
```diff
@@ -142,14 +142,6 @@ public static boolean checkAllowedPlots(
                 }
             }
         }
-        int maxMerge = Permissions.hasPermissionRange(player, Permission.PERMISSION_MERGE, Settings.Limit.MAX_PLOTS);
-        if (sizeX * sizeZ > maxMerge) {
```
SirYwell, line 146: We could simply check `if (sizeX * sizeZ > Math.max(1, maxMerge))` here instead of reverting

SirYwell (changes requested):

Core/src/main/java/com/plotsquared/core/command/Auto.java
NotMyFault: That would have been one alternative. Given this breaks a serious feature, my aim targets reverting to status quo and redo stuff from here, rather of leaving things unfixed for longer than needed.

PR #3915, opened 2023-01-08
Remove unused editorconfig rules

NotMyFault:
These rules apply to web frontends.


.editorconfig
```diff
@@ -963,47 +336,6 @@ ij_kotlin_wrap_elvis_expressions = 1
 ij_kotlin_wrap_expression_body_functions = 0
 ij_kotlin_wrap_first_method_in_call_chain = false
 
-[{*.har, *.jsb2, *.jsb3, *.json, .babelrc, .eslintrc, .stylelintrc, bowerrc, jest.config, mcmod.info}]
```
SirYwell, line 966: We have `.json` resources, maybe it makes sense to keep that?

SirYwell (changes requested):

SirYwell (approved):

PR #3916, opened 2023-01-09
Drop HTTP4J

NotMyFault:
There's no trace that this is still needed or used. Initially, I thought this would be needed for Arkitektonika or the paste service, but both work without, too.


dordsor21 (approved):
Given it builds it must be fine ¯\\_(ツ)_/¯

PR #3917, opened 2023-01-09
Add source documentation to javadocs and add dependency tab

NotMyFault:
- isLinkSource generates HTML files for all classes to facilitate browsing javadocs without the need to change between source code and back
- isUse adds a dedicated "USE" tab to the navbar to visualize class dependencies
- keywords adds some optional metadata for search engines
- bottom adds a footer, to be considered, but I found it a funny idea :)


dordsor21 (approved):

SirYwell (approved):
Cool

PR #3921, opened 2023-01-10
Reduce `/plot home` overhead

SirYwell:
## Description

This fixes issues with extremly slow `/plot home` command usages when having many plots and large merged plots.
The `/plot home` command requires to check for every existing plot whether the player is an owner, means the caches were recalculated for almost every plot.

There are also less significant but not insignificant performance improvements by using an ArrayList instead of a HashSet to collect all the plots (we do not have duplicates there anyways) and using getOwnerAbs instead of getOwner. getOwner additionally checks the server-plot flag, but we don't need to do that for each connected plot as we do it for the current plot before already, avoiding HashMap lookups.

Note these changes aren't really perfect, in best case we can come up with a solution to query plots from a specific owner more efficiently at some point, without having to go through all plots.

## Submitter Checklist


dordsor21 (approved):

NotMyFault (approved):
Thanks!

PR #3922, opened 2023-01-10
Officially deprecated BukkitPlayer#BukkitPlayer()

NotMyFault:
According to the javadocs, this method is already discouraged. Let's make it official.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
@@ -67,14 +67,15 @@ public class BukkitPlayer extends PlotPlayer<Player> {
     private String name;
 
     /**
-     * <p>Please do not use this method. Instead use
-     * BukkitUtil.getPlayer(Player), as it caches player objects.</p>
-     *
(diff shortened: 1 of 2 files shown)


dordsor21 (approved):

SirYwell (changes requested):
The comment basically applies to the other ctor too. However, these ctors are requiredn in PlotSquared so we can't just remove them. In best case, there wouldn't be public ctors for this class, but from a quick look there's no nice solution to avoid that.

NotMyFault:
> so we can't just remove them

Nobody said they will be removed? Take a look at my javadoc comment.

SirYwell:
Well, change their visibility, whatever. Both are used in PlotSquared code right now, and a more restricted visibility isn't possible with the current structure.

Also, this causes a new warning in the class where this constructor is used. It should be suppressed if it does not apply or the code should be changed to not use a deprecated method.

NotMyFault:
> a more restricted visibility isn't possible with the current structure.

I had no issue weakening the access level at compile time.

SirYwell:
I see, I misread the code. Still, the deprecation should apply to the other constructor aswell and the warnings should be suppressed in BukkitPlayerManager.

SirYwell (approved):

PR #3930, opened 2023-01-15
Remove uses of deprecated `Permissions` class and add javadoc notes

### LINKED ISSUE #3907: [v7] Clarify state of deprecation for Permission class
@dordsor21 You deprecated https://github.com/IntellectualSites/PlotSquared/blob/v6/Core/src/main/java/com/plotsquared/core/util/Permissions.java without pointing to a method end users can migrate to.
Could you add a method to the javadocs, so users can use to avoid using the deprecated methods? Otherwise, I'd recommend removing the deprecation again on the v6 branch and add general javadocs that this method _may_ change in the future.

dordsor21:
- Closes #3907


NotMyFault (approved):
Thanks a lot!

Issue #3878, opened 2022-11-24
Chunk#isLoaded does not always return true after a chunk is loaded with PaperLib#getChunkAtAsync causing "holes" in an edit queue


DevJoey:
### Server Implementation

Paper

### Server Version

1.19.2

### Describe the bug

Hart to describe look at the images ;(
![image](https://user-images.githubusercontent.com/38475631/203857779-f5c2a9f4-fff7-4c64-a42a-8d5543c0ff45.png)

### To Reproduce

Clear a big plot

### Expected behaviour

All roads have to be correct

### Screenshots / Videos

![image](https://user-images.githubusercontent.com/38475631/203857815-8875d057-433d-4160-9833-48a2e78a51fd.png)

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/441cf229e03c4e878ac7e94ee343079e

### PlotSquared Version

6.10.4-Premium

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

SirYwell:
Could you test this with WorldEdit instead of FastAsyncWorldEdit?

DevJoey:
> Could you test this with WorldEdit instead of FastAsyncWorldEdit?

With normal we all works fine

SirYwell:
Do you know with which FAWE build it worked last? Doing binary search here is likely the fastest way to find the cause.

DevJoey:
> Do you know with which FAWE build it worked last? Doing binary search here is likely the fastest way to find the cause.

No sorry! I setup a server for my friend and deleted a 12er merge. I cannot say which version has worked. I think it has something to do with the async block-placing from fawe.

Incluuu:
I have the same problem, i use fawe.

cosmohdx:
Is here any update?

Still having the same issue with 1.19.2 latest build and FAWE 2.4.4 and also with FAWE 2.5.0

dordsor21:
This is not limited to FAWE and is not a WE or FAWE issue.

The issue is `Chunk#isLoaded` doesn't necessarily return true when called within `PaperLib#getChunkAtAsync#whenComplete`

PR #3933, opened 2023-01-15
fix: ensure plots are fully unlinked before the clear itself is run

dordsor21:
- Possibly helps with #3878


SirYwell (approved):

Core/src/main/java/com/plotsquared/core/plot/PlotModificationManager.java
```diff
@@ -321,7 +324,22 @@ public void run() {
...
+        return unlinkPlot(createRoad, createSign, null);
+    }
+
+    /**
+     * Unlink the plot and all connected plots.
+     *
+     * @param createRoad whether to recreate road
+     * @param createSign whether to recreate signs
+     * @param whenDone   Task to run when unlink is complete
+     * @return success/!cancelled
```
NotMyFault, line 336:
```suggestion
     * @return success/!cancelled
     * @since TODO
```

NotMyFault (approved):

PR #3939, opened 2023-01-17
refactor: perform various code cleanup

dordsor21:


dordsor21:
Will raise a PR to v6 addressing method removals soon

NotMyFault:
I don't think including https://github.com/IntellectualSites/PlotSquared/pull/3939/commits/8a7c65999269b6bc2ac45e2d61ace9fb0ea37655 right now is a great idea.

This will cause future merge conflicts from v6, we should probably keep that part on hold until we're done on the v6 branch.
Preferably, we transfer what we find of use from editorconfig to spotless rules.
While editorconfig manages code style at your IDE, there is no check at compile time to warrant your changes actually follow editorconfig, which makes it useless, as proven evidently by the vast amount of reformatting changes.

PR #3941, opened 2023-01-20
chore: Remove things marked as for removal

dordsor21:


Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayerManager.java
```diff
@@ -66,13 +66,12 @@ public BukkitPlayer getPlayer(final @NonNull Player object) {
     }
 
     @Override
-    @SuppressWarnings("deprecation")
```
SirYwell, line 69: I can't link the method above, but the suppression should be redundant there too.

SirYwell (approved):
One minor nit, but looks good overall.

NotMyFault (approved):

Issue #397, opened 2015-07-14
Couple more chat color inconsistancies


Log1x:
Hi. Reporting a couple more inconsistencies with chat coloring. This list may be expanded on as I do configuration, so I will continue to update this Issue until it is assigned.

![Plot Flags](http://log1x.com/screenshots/2015-07-13_22-15-32.png)
/p set flag needs to use the style.yml colors.

![Plot Comments](http://log1x.com/screenshots/2015-07-13_21-47-01.png)
%amount% doesn't appear to be working as well as the username color is hardcoded.

![PlotSquared](http://log1x.com/screenshots/2015-07-13_22-19-47.png)
"or" was accidentally typo'd &2 instead of $2 in the default translation file

![Debug Language](http://log1x.com/screenshots/2015-07-14_00-22-19.png)
Small typo for 'debug' on line 213 of the translation file.

![Owner Language](http://log1x.com/screenshots/2015-07-14_00-48-23.png)
%owner% for plot_info has a white-space at the beginning of it making it not uniform with the rest of the variables. Easily fixable visually in the translation file, but the variable its self might as well have that prepended white-space removed.

![/p list](http://log1x.com/screenshots/2015-07-14_01-04-19.png)
/p list's Possible Values return doesn't seem to be translatable.

![/p list 2](http://log1x.com/screenshots/2015-07-14_01-07-10.png)
Above is the closest value I can find, but it does not seem to affect it.

![whitespace](http://log1x.com/screenshots/2015-07-14_01-19-54.png)
Line 69 'no_plot_inbox' has an extra white-space.

<3

Sauilitired:
Fixed:

* Flag List Colours
* "or" typo
* debug_header typo
* no_plot_inbox whitespace

For the list values, the argument list is based on your permissions. 
I couldn't find the info owner whitespace issue.

PR #3942, opened 2023-01-20
chore: deprecate and add javadoc notes for methods to be removed in v7

dordsor21:


Core/src/main/java/com/plotsquared/core/components/ComponentPreset.java
```diff
@@ -74,30 +74,65 @@ public static ComponentPreset deserialize(final @NonNull Map<String, Object> map
         );
     }
 
+    /**
+     * @deprecated {@link ComponentPreset} class will be replaced by a record in the next major update. This method will become
+     *         {@code ComponentPreset#component()}
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 81: The same deprecation should be placed on the class itself too.

Core/src/main/java/com/plotsquared/core/location/ChunkWrapper.java
```diff
@@ -23,8 +23,25 @@
 
 public class ChunkWrapper {
 
+    /**
+     * @deprecated {@link ChunkWrapper} class will be replaced by a record in the next major update. This variable will be
+     *         accessible using {@code ChunkWrapper#x()}
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 30: The same deprecation should be placed on the class itself too.

Core/src/main/java/com/plotsquared/core/plot/comment/PlotComment.java
```diff
@@ -22,11 +22,45 @@
 
 public class PlotComment {
 
+    /**
+     * @deprecated {@link PlotComment} class will be replaced by a record in the next major update. This variable will be
+     *         accessible using {@code PlotComment#comment()}
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 29: The same deprecation should be placed on the class itself too.

Core/src/main/java/com/plotsquared/core/util/FileBytes.java
```diff
@@ -20,7 +20,18 @@
 
 public class FileBytes {
 
+    /**
+     * @deprecated {@link FileBytes} class will be replaced by a record in the next major update. This variable will be
+     *         accessible using {@code FileBytes#path()}
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 27: The same deprecation should be placed on the class itself too.

Core/src/main/java/com/plotsquared/core/uuid/UUIDMapping.java
```diff
@@ -36,10 +36,20 @@ public UUIDMapping(final @NonNull UUID uuid, final @NonNull String username) {
         this.username = username;
     }
 
+    /**
+     * @deprecated {@link UUIDMapping} class will be replaced by a record in the next major update. This method will become
+     *         {@code UUIDMapping#username()}
+     */
+    @Deprecated(forRemoval = true, since = "TODO")
```
NotMyFault, line 43: The same deprecation should be placed on the class itself too.

SirYwell (commented):
I'm not sure about the deprecations without existing replacement. Those will show up as warnings or errrors in tools, but developers can't do anything about it as there is no alternative (yet), and if they want to still provide support v6, they're stuck with those warnings.

For getters in records, I'd say we either
- just remove them in v7 without previous deprecation
- add fluent accessors as a replacement in v6 already
- add explicit getters to v7

What's a good solution might depend on the actual class, especially for classes that are meant to be used internally only, I'd just mark them as internal and just make them records in v7 without changing anything else. For classes that are meant to be consumed by API users, allowing to use fluent accessors in v6 already might be a better solution.

I don't have a strong opinion on the actual solution, but I don't think it#s useful for anyone to deprecate the methods that will only get replacements in v7.

NotMyFault:
> add fluent accessors as a replacement in v6 already

That may be the solution with the least maintenance burden.

NotMyFault:
I separated this PR into several ones, making it easier to build consensus: https://github.com/IntellectualSites/PlotSquared/pull/397, https://github.com/IntellectualSites/PlotSquared/pull/3977 and https://github.com/IntellectualSites/PlotSquared/pull/3976

PR #3944, opened 2023-01-21
PS-3908: Cleanup deprecated methods to improve code maturity.

### LINKED ISSUE #3908: Cleanup deprecated methods to improve code maturity
Below, I've assembled a non-exhaustive list of methods we can switch away from, others may be kept for Spigot and backwards compatibility.

- https://github.com/IntellectualSites/PlotSquared/blob/339ca8e30fda26f2f1fa2522d692872cd02ad701/Core/src/main/java/com/plotsquared/core/command/Area.java#L232-L235 is deprecated in favor of `EditSessionBuilder#newEditSessionBuilder()`
- https://github.com/IntellectualSites/PlotSquared/blob/339ca8e30fda26f2f1fa2522d692872cd02ad701/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java#L207 is deprecated in favor of `#canAccess`

mkaulfers:
## Overview
https://github.com/IntellectualSites/PlotSquared/issues/3908

Fixes #3908 
- Deprecated `isAccessible` and `getEditSessionFactory()`

## Description
Removes the deprecations and updates to the newer `editSessionBuilder` and `canAccess(...)`

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
@@ -204,7 +204,7 @@ public class HybridPlotWorld extends ClassicPlotWorld {
                 }
                 Object value;
                 try {
-                    final boolean accessible = field.isAccessible();
+                    final boolean accessible = field.canAccess(field);
                     field.setAccessible(true);
                     value = field.get(this);
                     field.setAccessible(accessible);
--- a/Core/src/main/java/com/plotsquared/core/command/Area.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Area.java
@@ -56,6 +56,7 @@ import com.plotsquared.core.util.TabCompletions;
 import com.plotsquared.core.util.WorldUtil;
 import com.plotsquared.core.util.task.RunnableVal3;
 import com.sk89q.worldedit.EditSession;
+import com.sk89q.worldedit.EditSessionBuilder;
 import com.sk89q.worldedit.LocalSession;
 import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.entity.Player;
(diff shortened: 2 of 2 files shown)


mkaulfers:
On the `Area.java` changes from `.isAccessible()` to `field.canAccessField(field)` just feels... wrong? This is some weird reflection going on with this one and if anyone cares to teach some poor soul how this reflection check should work, please be my guest. 🤣

NotMyFault (approved):
Thanks! The reflection change is fine so far, #canAccess(object) is a drop-in replacement for #isAccessibler We already use it in a few other spots.

Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java
```diff
@@ -204,7 +204,7 @@ public void loadConfiguration(ConfigurationSection config) {
                 }
                 Object value;
                 try {
-                    final boolean accessible = field.isAccessible();
+                    final boolean accessible = field.canAccess(field);
```
SirYwell, line 207: This will throw an `IllegalArgumentException`. However, we do not need the accessible stuff at all: `getField` only returns public fields and those fields all belong to the current class or superclasses. `setAccessible` would only be needed for `private` fields from superclasses, but as we do not access them, this code isn't needed.

(Also, we wouldn't need to "reset" the accessible state of the field, as it is bound to the `Field` object rather than the underlying field, and the `Field` object isn't referenced outside of this method - it will be just garbage collected)

mkaulfers: See that's why I thought this whole thing looked funny hah. Basically this whole thing can be cleaned up and removed if I'm understanding correctly?

SirYwell: Yes, only the `field.get(this)` call is needed.

mkaulfers: I'll get a new PR up for that :D

PR #3952, opened 2023-01-29
Make farmland keep its moisture

traksag:
## Overview

The `soil-dry` flag is false by default and prevents farmland from turning into dirt. However, moisturised farmland, i.e. farmland with the block property `moisture=7`, still loses its moisture over time and turns into 'dry' farm land. 

## Description
Cancel `MoistureChangeEvent` when needed.

## Submitter Checklist


NotMyFault (approved):
Good catch, thanks!

SirYwell (approved):
Looks reasonable

ByteExceptionM:
This update ensures that farmland now no longer gets wet. However, the flag should only prevent drying. The following PR fixes that: #3978

PR #3955, opened 2023-02-03
feat: Add event firing on remove road entity

ByteExceptionM:
## Description
Firing the cancellable `RemoveRoadEntityEvent` when an entity is removed from the road

## Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
@@ -1000,11 +1000,11 @@ public final class BukkitPlatform extends JavaPlugin implements Listener, PlotPl
     private void removeRoadEntity(Entity entity, Iterator<Entity> entityIterator) {
         RemoveRoadEntityEvent event = eventDispatcher.callRemoveRoadEntity(BukkitAdapter.adapt(entity));
 
-        entityIterator.remove();
-
-        if (event.getEventResult() == Result.DENY)
+        if (event.getEventResult() == Result.DENY) {
             return;
+        }
 
+        entityIterator.remove();
         entity.remove();
     }
 
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
@@ -71,6 +71,7 @@ import com.plotsquared.core.configuration.Storage;
 import com.plotsquared.core.configuration.caption.ChatFormatter;
 import com.plotsquared.core.configuration.file.YamlConfiguration;
 import com.plotsquared.core.database.DBFunc;
+import com.plotsquared.core.events.Result;
 import com.plotsquared.core.generator.GeneratorWrapper;
 import com.plotsquared.core.generator.IndependentPlotGenerator;
 import com.plotsquared.core.generator.SingleWorldGenerator;
@@ -109,6 +110,7 @@ import com.plotsquared.core.uuid.CacheUUIDService;
 import com.plotsquared.core.uuid.UUIDPipeline;
 import com.plotsquared.core.uuid.offline.OfflineModeUUIDService;
 import com.sk89q.worldedit.WorldEdit;
+import com.sk89q.worldedit.bukkit.BukkitAdapter;
 import io.papermc.lib.PaperLib;
 import net.kyori.adventure.audience.Audience;
 import net.kyori.adventure.text.Component;
@@ -148,6 +150,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Queue;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


Core/src/main/java/com/plotsquared/core/events/EntityEvent.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private final Entity entity;
+    private final EntityType entityType;
+
+    private final UUID entityUuid;
+    private final int entityId;
+
+    private String name;
+
+    public EntityEvent(Entity entity, EntityType entityType, UUID entityUuid, int entityId) {
```
SirYwell, line 37: Actually only the WE Entity should be needed. The type can be directly accessed through the BaseEntity, and the UUID is part of the NBT data. The int entityId might not be exposed on other platforms at all, so I think it's better to leave it out.

Core/src/main/java/com/plotsquared/core/events/EntityEvent.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Obtain the entity id of the entity involved in the event
+     *
+     * @return id of entity
+     */
+    public int getEntityId() {
+        return this.entityId;
+    }
+
+    public String getName() {
```
SirYwell, line 80: This method kinda duplicates the method below, with rather unintended behavior. Is it needed?

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
```diff
@@ -811,7 +815,7 @@ private void runEntityTask() {
                                             continue;
                                         }
                                         iterator.remove();
-                                        entity.remove();
```
SirYwell, line 814: The `iterator.remove()` call should be moved to the new method too

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
```diff
@@ -999,6 +1003,20 @@ private void runEntityTask() {
...
 
+    private void removeRoadEntity(Entity entity) {
+        RemoveRoadEntityEvent event = eventDispatcher.callRemoveRoadEntity(
+                BukkitAdapter.adapt(entity),
+                BukkitAdapter.adapt(entity.getType()),
+                entity.getUniqueId(),
+                entity.getEntityId()
+        );
+
+        if (Objects.equals(event.getEventResult(), Result.DENY))
```
SirYwell, line 1014: `==` is fine here as we're comparing enums.

SirYwell (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
ByteExceptionM: Since the EventResult can also be null, I use 'Objects#equals' here to prevent a NPE and to perform a null check through it

PierreSchwang: that's should be no issue tho - if the result is no the check would simplify to `null == DENY` -> `false`, which would be the expected behaviour.

Core/src/main/java/com/plotsquared/core/events/EntityEvent.java
ByteExceptionM: Does the WE Entity class contain the NBT data of the entity? Have actually not seen this

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
ByteExceptionM: Ah okay. I don't use `==` normally, think it doesn't really look nice. But sounds logical, I'll change it.

Core/src/main/java/com/plotsquared/core/events/EntityEvent.java
ByteExceptionM: Got it, thanks. Going to change that

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
```diff
@@ -999,6 +997,17 @@ private void runEntityTask() {
         }), TaskTime.seconds(1L));
     }
 
+    private void removeRoadEntity(Entity entity, Iterator<Entity> entityIterator) {
+        RemoveRoadEntityEvent event = eventDispatcher.callRemoveRoadEntity(BukkitAdapter.adapt(entity));
+
+        entityIterator.remove();
```
SirYwell, line 1003: The removal should happen only if `Entity#remove` is called. Also please add braces to the if below.

SirYwell (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/BukkitPlatform.java
ByteExceptionM: I don't quite understand the not removing from the iterator, but have adjusted it anyway.

SirYwell (approved):
Looks good!

NotMyFault (approved):
Thanks for your contribution!

PR #3961, opened 2023-02-12
feat: Add PostPlotClearEvent

ByteExceptionM:
## Description
This PR brings in that after a plot clear, the event `PostPlotClearEvent` is fired.

## Submitter Checklist


dordsor21:
I wonder if we should add a superclass for "post" events as we're getting a few of them now, or some other differentiation given we do not need to do anything with the event once fired

PR #3976, opened 2023-03-04
Record-inize methods

NotMyFault:
This is an alternative approach to https://github.com/IntellectualSites/PlotSquared/pull/3942/, without deprecations in v6 because the methods are subject to removal anyway. Therefore, deprecating them pointing to the drop-in methods is a bit _point_less.

I don't plan to spend more time implementing static accessors in v6 at this point.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/BlockBucket.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/BlockBucket.java
@@ -215,66 +215,44 @@ public final class BlockBucket implements ConfigurationSerializable {
         return result;
     }
 
-    private static final class Range {
-
-        private final int min;
-        private final int max;
-        private final boolean automatic;
-
-        public Range(int min, int max, boolean automatic) {
-            this.min = min;
-            this.max = max;
-            this.automatic = automatic;
-        }
+    private record Range(int min, int max, boolean automatic) {
 
         public int getWeight() {
-            return max - min;
-        }
+                return max - min;
+            }
 
-        public boolean isInRange(final int num) {
-            return num <= max && num >= min;
-        }
+            public boolean isInRange(final int num) {
+                return num <= max && num >= min;
+            }
 
-        public int getMin() {
-            return this.min;
-        }
-
-        public int getMax() {
-            return this.max;
-        }
-
-        public boolean equals(final Object o) {
-            if (o == this) {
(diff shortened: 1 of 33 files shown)


NotMyFault:
@SirYwell Your favorite checkstyle has been applied.

Core/src/main/java/com/plotsquared/core/plot/BlockBucket.java
```diff
@@ -215,66 +215,48 @@ public int hashCode() {
...
-        public int getMax() {
-            return this.max;
-        }
+            public boolean isInRange(final int num) {
+                return num <= max && num >= min;
+            }
 
-        public boolean equals(final Object o) {
-            if (o == this) {
+            public boolean equals(final Object o) {
```
SirYwell, line 232: I think we can fully remove the equals and hashCode methods here

Core/src/main/java/com/plotsquared/core/services/plots/AutoQuery.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Crate a new auto query
+     *
+     * @param player   Player to claim for
+     * @param startId  Plot ID to start searching from
+     * @param sizeX    Number of plots along the X axis
+     * @param sizeZ    Number of plots along the Z axis
+     * @param plotArea Plot area to search in
+     */
+    public AutoQuery(
```
SirYwell, line 44: Annotations can be added to the record components, and the `@param` tag is allowed at the record header too. Can we remove the constructor as well as the methods?

Core/src/main/java/com/plotsquared/core/uuid/UUIDMapping.java
```diff
@@ -20,29 +20,15 @@
...
-    public @NonNull UUID getUuid() {
-        return this.uuid;
-    }
+public record UUIDMapping(
+        @NonNull UUID uuid,
+        @NonNull String username
+) {
 
     @Override
     public boolean equals(final Object o) {
```
SirYwell, line 34: This method can be dropped too I guess?

SirYwell (changes requested):
There are a few places were we can probably remove even more code, but looks nice!

SirYwell (approved):

PR #3991, opened 2023-03-17
fix: clear above and below plot gen heights if build heights are larger

### LINKED ISSUE #3985: Blocks above 255 are not removed when a plot is cleared
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

When you do a /plot clear, it doesn't clear blocks above 255.

### To Reproduce

https://pastes.dev/WY9kThu8KK

Use these world settings.

### Expected behaviour

@PierreSchwang in discord said "Seems like the max_gen_height is used. Create an issue for that" so I assume clearing needs to use max_height instead.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/d154f3e8d89f4e89bfe6d1291f265577

### PlotSquared Version

6.11.0-Premium

### Checklist

### Anything else?

_No response_

dordsor21:
Fixes #3985 (also requires IntellectualSites/FastAsyncWorldEdit#2153)

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/HybridPlotManager.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/HybridPlotManager.java
@@ -288,6 +288,13 @@ public class HybridPlotManager extends ClassicPlotManager {
             queue.setCompleteTask(whenDone);
         }
         if (!canRegen) {
+            if (hybridPlotWorld.getMinBuildHeight() < hybridPlotWorld.getMinGenHeight()) {
+                queue.setCuboid(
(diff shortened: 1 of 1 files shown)


NotMyFault:
> also requires https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2153

But it also works, or in this case, doesn't with older versions of FAWE, no?

dordsor21:
Well, if using FAWE you need to use the latest version as FAWE replaces P2 clear rather than being the queue

NotMyFault (approved):
Thanks!

SirYwell (approved):

PR #4017, opened 2023-04-18
Fix custom biome error when a plot is cleared

### LINKED ISSUE #4012: /p clear has trouble with custom biomes
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

When I try to clear a plot that has a custom registered biome with Volans (or any other custom biome plugin), errors happen in the console.

### To Reproduce

1. Create a custom biome using a biome plugin
2. Use for example /p setbiome $volans:blue (normally: /p setbiome $yourplugin:yourbiome)
3. Try to clear the plot with /p clear
4. Wait for errors / result

### Expected behaviour

It should remove what was build on the plot, maybe also the custom biome, but you could always override the biome with /p setbiome forest. It does unfortunately not work for this case, as the backup process became stuck and overriding biomes only helps for the /p clear process after a server restart.

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/24d5a1203890463289561c3db3905d61

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/833d3df57c844b21be02b4d6953709f4

### PlotSquared Version

6.10.4

### Checklist

### Anything else?

I am not sure if it is 100% a plotsquared or a FAWE / Worldedit issue because there are different problems mixing in together

TheMeinerLP:
## Overview

Fixes #4012

## Description
One of my plugins are register custom biomes. But PS have some trouble to clean that from a plot

## Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/SchematicHandler.java
+++ b/Core/src/main/java/com/plotsquared/core/util/SchematicHandler.java
@@ -62,6 +62,7 @@ import com.sk89q.worldedit.regions.Region;
 import com.sk89q.worldedit.regions.RegionIntersection;
 import com.sk89q.worldedit.world.World;
 import com.sk89q.worldedit.world.biome.BiomeType;
+import com.sk89q.worldedit.world.biome.BiomeTypes;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockTypes;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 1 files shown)


Core/src/main/java/com/plotsquared/core/util/SchematicHandler.java
```diff
@@ -771,6 +772,10 @@ public void run() {
                                     }
                                     BlockVector2 pt = BlockVector2.at(currentX, currentZ);
                                     BiomeType biome = aabb.getWorld().getBiome(pt);
+                                    // Replace empty biome with plains
+                                    if(biome == null) {
```
dordsor21, line 776: Space required

Core/src/main/java/com/plotsquared/core/util/SchematicHandler.java
```diff
@@ -771,6 +772,10 @@ public void run() {
...
+                                        biome = BiomeTypes.PLAINS;
```
dordsor21, line 777: Unsure if we should default to the plotworld's biome (as set in worlds.yml) rather than forcing plains

TheMeinerLP: Be a deal

TheJonstone:
Hi, this issue has already been fixed.
https://github.com/IntellectualSites/FastAsyncWorldEdit/pull/2510

PR #4099, opened 2023-07-02
feat: improvement of kick messages

RedstoneFuture:
## Description
This PR differentiates the messages for the kick in case of an error:
- If the player is online but not on the plot, a new message `kick.player_not_in_plot` is now executed.
- If a player has admin permission, a separate message is addressed when using the normal plot commands (not cluster concept).
- Minor change to the standard texts.
- Using the placeholder "player" instead of "name".

```[tasklist]
### Submitter Checklist
```


Core/src/main/java/com/plotsquared/core/command/Kick.java
```diff
@@ -112,15 +112,15 @@ public boolean onCommand(PlotPlayer<?> player, String[] args) {
...
+                                TagResolver.resolver("player", Tag.inserting(Component.text(player2.getName())))
                         );
                         return;
                     }
                     if (player2.hasPermission(Permission.PERMISSION_ADMIN_ENTRY_DENIED)) {
                         player.sendMessage(
-                                TranslatableCaption.of("cluster.cannot_kick_player"),
-                                TagResolver.resolver("name", Tag.inserting(Component.text(player2.getName())))
+                                TranslatableCaption.of("kick.cannot_kick_player"),
+                                TagResolver.resolver("player", Tag.inserting(Component.text(player2.getName())))
```
NotMyFault, line 123: What's the motivation behind duplicating the string when the context stays the same? We reuse the same message, even if their prefix is different.

I'm not fond of changing 'name' to 'player' either because fresh installations only get the update, existing users will need to upgrade their localization file manually, causing more work than worth it.

RedstoneFuture: Sorry for the confusion.

**Custom configuration for the plot kick message:**
I have separated the "normal" plot-based kick message from the cluster-based kick message, because the message in that case has nothing to do with the cluster concept.

Whether the cluster needs its own kick message is another question. Anyway, I find it strange that some plot commands in the `Cluster.jar` have been reprogrammed again, instead of referring to the basic plot commands.

Currently, the "normal" kick messages are in the `Messages.yml` now under "kick". Alternative, I can refer to the normal kick messages in the `Cluster.jar` as desired. Then "cluster.cannot_kick_player" and "cluster.cluster_kicked_user" are omitted.

**Changing the placeholder name:**
I changed the placeholder to "player" to have continuity with other messages where the player name of a valid player should also be displayed (e.g. "member.already_added"). So I change the message configuration, but a consistent configuration should have a higher priority. (It can also be pulled as part of the V7 Change.) Otherwise, unsorted wordings accumulate over the years, which makes a new message reconfiguration difficult.

If IntellectualSites's programming ethics prohibit modifying wordings in configurations to make version changes easier, perhaps this should be explicitly stated in [Contribute Instruction](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md). Otherwise, the goal should be to improve the project in terms of continuity and readability.

---

I hope that this was understandable so far and that I was able to justify my motives. If, contrary to my reasoning, corresponding changes from the PR are to be rejected, please indicate this.

NotMyFault (changes requested):
> but a consistent configuration should have a higher priority.

By that justification, existing language files must be automatically upgraded.
Leaving the maintenance burden to the user by manually updating the localization keys between updates is not desirable.

NotMyFault (approved):

PR #4106, opened 2023-07-10
Fix: remove everyone command

RedstoneFuture:
## Overview
With this PR the Remove command is fixed. When using `/p remove *` now really every player who is geadded / trusted / denied will be removed from the plot.

## Description
Revised the delete process for EVERYONE so that all trusted / added / denied players are processed. In addition, the counting of the removed players has been adjusted.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Remove.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Remove.java
@@ -100,15 +100,17 @@ public class Remove extends SubCommand {
                             count++;
                         }
                     } else if (uuid == DBFunc.EVERYONE) {
+                        count += plot.getTrusted().size();
                         if (plot.removeTrusted(uuid)) {
                             this.eventDispatcher.callTrusted(player, plot, uuid, false);
-                            count++;
-                        } else if (plot.removeMember(uuid)) {
+                        }
+                        count += plot.getMembers().size();
+                        if (plot.removeMember(uuid)) {
                             this.eventDispatcher.callMember(player, plot, uuid, false);
-                            count++;
-                        } else if (plot.removeDenied(uuid)) {
+                        }
+                        count += plot.getDenied().size();
+                        if (plot.removeDenied(uuid)) {
                             this.eventDispatcher.callDenied(player, plot, uuid, false);
-                            count++;
                         }
                     }
                 }


dordsor21 (approved):

NotMyFault (changes requested):
> With this PR the Remove command is fixed.

Such a statement implies the current behavior is broken when it's not.
`/plot remove *` removes users in a hierarchy. Removing all added, trusted and denied users in the same run are potentially dangerous, when used too quickly without a way to recover.

RedstoneFuture:
> > With this PR the Remove command is fixed.
> 
> Such a statement implies the current behavior is broken when it's not. `/plot remove *` removes users in a hierarchy. Removing all added, trusted and denied users in the same run is potentially dangerous, when used too quickly without a way to recover.

If `/p a dd *` adds everyone, the `/p remove *` should also remove everyone from the plot. The current behavior is flawed for me from a logical point of view, yes.

NotMyFault:
> If `/p a dd *` adds everyone, the `/p remove *` should also remove everyone from the plot.

`/plot add *` adds everyone to "Added" scope of the plot. It does not add them to the "Trusted" or "Denied" scope.

RedstoneFuture:
> `/plot add *` adds everyone to "Added" scope of the plot. It does not add them to the "Trusted" or "Denied" scope.

Yes, but in practice, an ADD * is bad enough. The focus on the ADD command does not matter then.

We could also force a "/p confirm" command here. Or what do you suggest?

RedstoneFuture:
@IntellectualSites/plotsquared-team 

I would like to point out again that my changes fix the logic that corresponds to the **official command syntax**: 
`/p remove <name | *>`.

The command `/p remove <name>` removes `<name>`, regardless of whether it is saved as _ADD_, _TRUST_ or _DENIED_. Then `*` must also remove _all_ players from ADD, TRUST and DENIED.

As long as the player does not remove all players according to the syntax or can choose whether he wants to remove all players from ADD, TRUST or DENIED, this "long-time behavior" will be seen as a bug by the users.

SirYwell (approved):
I think this change makes sense. Note that functionality only changes if `*` is not added, trusted, or denied. In such case, the code above the change will already cover it.
Otherwise, all players will be removed from added, trusted, and denied, which makes more sense that rather unpredictable behavior.

SirYwell:
I don't see how the current implementation is any better. In my opinion, a confirmation dialog should be addressed separately, as it is a separate problem. The current behavior is extremely flaky and inconsistent, no one would figure out how it works without reading the code. So I can't see how anyone could rely on that behavior or expect it work like that.

SirYwell:
> You missed the point I made. I outlined the scenario of a user running the command with the proposed change, now resulting in all scopes being wiped, which hasn't been the case before.
> 
> While the proposal is correct from a logical aspect, it creates a major discrepancy between the current and proposed behavior, I consider harmful. This PR shouldn't focus on addressing logic issues only, totally neglecting the user-facing impact the change submitted actually has.

I fully understand your point. But there isn't much to it, because currently the command, from a user perspective, might fully clear a random scope, leaving others untouched. This unexpected, confusing behavior isn't less harmful at all, so this PR does not neglect the user-facing impact, it makes sure the command does work reliably, independenly of the state of specific scopes.

NotMyFault:
> it makes sure the command does work reliably, independenly of the state of specific scopes.

Reliability cannot be considered a higher priority than safety, especially when introducing irreversible actions without confirmation, and therefore this point can be disregarded.

To move on with the PR I propose:
- to implement a `/plot confirm` dialogue to all `/plot remove` actions, refactoring the remove command and ironing out the logic issue.
- to implement a `/plot confirm` dialogue subject to asterisks only
- removing the functionality to wipe all scopes

SirYwell:
> Reliability cannot be considered a higher priority than safety, especially when introducing irreversible actions without confirmation, and therefore this point can be disregarded.

We don't give up any safety here. We don't introduce any new irreversible action.

I agree that a confirmation dialog would make sense when removing `*`, though the behavior and impact of removing `*` depends on whether `*` is part of any scope already, and it might be weird to have a confirmation dialog for `*` in some cases only, so there is still room for discussion.

Generally, I hope we can reintroduce an improved version of the removed commands configuration, potentially allowing such fine-grained control over confirmation dialogs configurable by the user. But that basically depends on cloud.

NotMyFault (approved):
> Generally, I hope we can reintroduce an improved version of the removed commands configuration, potentially allowing such fine-grained control over confirmation dialogs configurable by the user. But that basically depends on cloud.

I agree, the remove functionality needs to be revised at some point. In the best case, refactored to a level the user can determine the part the remove command should affect.

Issue #3940, opened 2023-01-19
Unable to create single plot worlds on 1.19.3


NotMyFault:
### Server Implementation

Paper

### Server Version

1.19.3

### Describe the bug

I'm unable to create a single plot world using the build in generator.

### To Reproduce

[09:15:22] [Server thread/INFO]: NotMyFault issued server command: /plot setup
[09:15:26] [Server thread/INFO]: NotMyFault issued server command: /plot setup PlotSquared:single
[09:15:31] [Server thread/INFO]: NotMyFault issued server command: /plot setup normal
[09:15:37] [Server thread/INFO]: NotMyFault issued server command: /plot setup false
[09:15:42] [Server thread/INFO]: NotMyFault issued server command: /plot setup flatworld

### Expected behaviour

The world should be generated as expected.

### Error log (if applicable)

https://gist.github.com/NotMyFault/5c68ef0f84e8a3bbaade6cb85b6ed689

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/06e630d0dba04096bd9c3bfcf6df5d0d

### PlotSquared Version

6.10.9

### Checklist

- [X] I have included a Plot debugpaste.
- [X] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

NotMyFault:
Loading in 1.20 using b7, yields:
```
[10:08:51 WARN]: java.lang.NoSuchMethodException: org.bukkit.craftbukkit.v1_20_R1.CraftChunk.getHandle()
[10:08:51 WARN]: 	at java.base/java.lang.Class.getDeclaredMethod(Class.java:2772)
[10:08:51 WARN]: 	at plotsquared-bukkit-7.0.0-SNAPSHOT.jar//com.plotsquared.core.util.ReflectionUtils$RefClass.getMethod(ReflectionUtils.java:130)
[10:08:51 WARN]: 	at plotsquared-bukkit-7.0.0-SNAPSHOT.jar//com.plotsquared.bukkit.listener.SingleWorldListener.<init>(SingleWorldListener.java:46)
[10:08:51 WARN]: 	at plotsquared-bukkit-7.0.0-SNAPSHOT.jar//com.plotsquared.bukkit.inject.BukkitModule.configure(BukkitModule.java:114)
[...]

PR #4112, opened 2023-07-16
fix: add missing method in UncheckedWorldLocation

dordsor21:
- Allows single plot worlds to be generated

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/location/UncheckedWorldLocation.java
+++ b/Core/src/main/java/com/plotsquared/core/location/UncheckedWorldLocation.java
@@ -60,6 +60,19 @@ public final class UncheckedWorldLocation extends Location {
         return new UncheckedWorldLocation(world, x, y, z);
     }
 
+    /**
+     * Construct a new location with yaw and pitch equal to 0
+     *
+     * @param world World
+     * @param loc   Coordinates
+     * @return New location
+     * @since 6.9.0
+     */
+    @DoNotUse
+    public static @NonNull UncheckedWorldLocation at(final @NonNull String world, BlockVector3 loc) {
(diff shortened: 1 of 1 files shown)


NotMyFault:
Does that aim to fix https://github.com/IntellectualSites/PlotSquared/issues/3940? Should be linked as `Fixes...` then 👀

dordsor21:
I guess, though I never got the linked issue in any testing before or after these changes

NotMyFault:
I'll check later if I can reproduce it on 1.20.1 still.

Core/src/main/java/com/plotsquared/core/location/UncheckedWorldLocation.java
```diff
@@ -60,6 +60,19 @@ private UncheckedWorldLocation(
         return new UncheckedWorldLocation(world, x, y, z);
     }
 
+    /**
+     * Construct a new location with yaw and pitch equal to 0
+     *
+     * @param world World
+     * @param loc   Coordinates
+     * @return New location
+     * @since 6.9.0
```
SirYwell, line 69: This is likely wrong?

SirYwell (changes requested):
(Just a comment) It's somewhat bad that we have shadowing static methods that can cause such bugs, but I also don't have a nice and simple solution to that

Core/src/main/java/com/plotsquared/core/location/UncheckedWorldLocation.java
dordsor21: Ah yeah that'll be because I copy-pasted it

SirYwell (approved):

NotMyFault (approved):

NotMyFault:
> I'll check later if I can reproduce it on 1.20.1 still.

Amendment: It does not. I've updated the issue with the new exceptions.

PR #4195, opened 2023-10-01
Migrate 'EntityUtil#capNumeral' to an enhanced switch

NotMyFault:
This must have been missed when cleaning up for v7, because all other switches follow this pattern.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/EntityUtil.java
+++ b/Core/src/main/java/com/plotsquared/core/util/EntityUtil.java
@@ -42,7 +42,7 @@ public class EntityUtil {
     }
 
     private static int capNumeral(final @NonNull String flagName) {
-        int i = switch (flagName) {
+        return switch (flagName) {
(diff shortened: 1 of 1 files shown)


Core/src/main/java/com/plotsquared/core/util/EntityUtil.java
```diff
@@ -42,27 +42,14 @@ private EntityUtil() {
...
-                i = CAP_VEHICLE;
-                break;
-            case "misc-cap":
-                i = CAP_MISC;
-                break;
-            case "entity-cap":
-            default:
-                i = CAP_ENTITY;
-        }
+        int i = switch (flagName) {
```
PierreSchwang, line 45: Would be easier to directly return the switch instead of assigning it to a variable with the only goal to immediately return it. Otherwise, not much that can be broken with that change lol

PierreSchwang (approved):

PR #4196, opened 2023-10-02
Feat(API): Mutable Location in PlayerTeleportToPlotEvent

PierreSchwang:
## Overview
In case of some more abstract plot layouts (e.g. floating balloons in the sky), the event now provides the ability to set the location for any teleport programmatically. 

Example:
```java
@Subscribe
public void scootRightOver(PlayerTeleportToPlotEvent event) {
    if (event.getCause() == TeleportCause.COMMAND_HOME) {
        event.setLocation(event.getLocation().add(-15, 0, 0));
    }
}
```

I did not deprecate the constructor nor the event dispatcher method itself. Events should not be called by end-users either way. If no location is passed, weird behavior can occur when trying to mutate the original location (which is null in case of those deprecated methods)

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/events/PlayerTeleportToPlotEvent.java
+++ b/Core/src/main/java/com/plotsquared/core/events/PlayerTeleportToPlotEvent.java
@@ -21,10 +21,7 @@ package com.plotsquared.core.events;
 import com.plotsquared.core.location.Location;
 import com.plotsquared.core.player.PlotPlayer;
 import com.plotsquared.core.plot.Plot;
-import org.jetbrains.annotations.NotNull;
-
-import java.util.Objects;
-import java.util.function.Consumer;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Called when a player teleports to a plot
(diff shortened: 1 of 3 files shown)


dordsor21:
I wonder if two events for teleport would be good. If the event is cancelled it would be good to not search for a location and possibly load chunks needlessly. Perhaps add a new "PlayerTeleportToPlotLocationChosenEvent" :D

PierreSchwang:
> I wonder if two events for teleport would be better. If the event is cancelled it would be good to not search for a location and possibly load chunks needlessly. Perhaps add a new "PlayerTeleportToPlotLocationChosenEvent" :D

I may have an idea to mitigate that issue. Tbh, I'm not that big of a fan for an additional event (especially with that very java-ish name :D)

Core/src/main/java/com/plotsquared/core/events/PlayerTeleportToPlotEvent.java
```diff
@@ -57,14 +60,36 @@ public TeleportCause getCause() {
...
         return this.from;
     }
 
+    /**
+     * Gets the currently applied {@link LocationTransformer} or null, if none was set
+     *
+     * @return LocationTransformer
+     * @since TODO
+     */
+    public @Nullable LocationTransformer getLocationTransformer() {
```
dordsor21, line 78: Is it not easier just to use `Function<Location, Location>`?

SirYwell: or even `UnaryOperator<Location>`

SirYwell (approved):
Looks good. The only thing we could potentially change is to make it non-null and use UnaryOperator.identity() as default value, but I don't have a strong opinion on that.

dordsor21 (approved):

PR #4075, opened 2023-06-13
Fixes many database issues and add scale for bigger propose

TheMeinerLP:
## Description
Currently our server get some connection lost issues or resets and plotsqured will after that be down.
This pr fixes generally very old code from PS² and is a modern way too go.

## Disclaimer
**This PR breaks other plugin compatibilty** 

```[tasklist]
### Submitter Checklist
- [X] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [X] Ensure that the pull request title represents the desired changelog entry.
- [X] New public fields and methods are annotated with `@since TODO`.
- [X] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md).
```

PR #4214, opened 2023-10-21
feature: connection pooling & SQL refactoring

Citymonstret:


RedstoneFuture:
Here is another PR with the reference to DB management:
https://github.com/IntellectualSites/PlotSquared/pull/4075

PR #4216, opened 2023-10-26
Delay economy initialization to server load

SirYwell:
## Description

Guice initializes Singletons eagerly, i.e. as soon as we initialize guice itself when enabling the plugin. However, economy implementations might load after PlotSquared (both an economy plugin and P2 will load after Vault, but their load order is not defined).

As a solution, we can wrap the EconHandler and only initialize it once the server is ready. This ensures that all plugins are loaded. We can still fall back to the NullEconHandler if no plugin provides economy by then.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/inject/BukkitModule.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/inject/BukkitModule.java
@@ -23,13 +23,13 @@ import com.google.inject.Provides;
 import com.google.inject.Singleton;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
 import com.plotsquared.bukkit.BukkitPlatform;
+import com.plotsquared.bukkit.listener.ServerListener;
 import com.plotsquared.bukkit.listener.SingleWorldListener;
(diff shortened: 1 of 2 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/inject/BukkitModule.java
```diff
@@ -128,21 +133,64 @@ protected void configure() {
     @Provides
     @Singleton
     @NonNull EconHandler provideEconHandler() {
-        if (!Settings.Enabled_Components.ECONOMY) {
+        if (!Settings.Enabled_Components.ECONOMY || !Bukkit.getPluginManager().isPluginEnabled("Vault")) {
```
dordsor21, line 136: Do we definitely always require Vault?

dordsor21: Seems so

dordsor21 (approved):

dordsor21 (approved):

NotMyFault (approved):

PR #4219, opened 2023-10-28
Feat: adding "last" argument for Visit cmd to choose the highest number

RedstoneFuture:
## Overview
With this PR it is allowed to use the `last` or `n` argument to choose the highest possible (last) plot number of the target. For this purpose, parts of the algorithm were also outsourced to a private method and extended.

If this is accepted so far, I can build the same for the Home command: `/p h last`.

**EDIT (21.12.2023):** Support for inverted plot numbers has also been added.

## Description
**Examples:**
- `/p v <username> last`
- `/p v <username> <area> last`
--> Teleports the player to the **x** plot of `<username>` if he has a total of **x**.

**EDIT (21.12.2023):**
- `/p v <username> -#`
- `/p v <username> <area> -#`

grafik

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Visit.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Visit.java
@@ -318,14 +318,20 @@ public class Visit extends Command {
     }
 
     private boolean isInvalidPageNr(String arg) {
-        if (MathMan.isInteger(arg)) return false;
-        if (arg.equals("last") || arg.equals("n")) return false;
+        if (MathMan.isInteger(arg)) {
+            return false;
+        } else if (arg.equals("last") || arg.equals("n")) {
+            return false;
+        }
         return true;
     }
 
     private int getPageNr(String arg) {
-        if (MathMan.isInteger(arg)) return Integer.parseInt(arg);
-        if (arg.equals("last") || arg.equals("n")) return -1;
+        if (MathMan.isInteger(arg)) {
+            return Integer.parseInt(arg);
+        } else if (arg.equals("last") || arg.equals("n")) {
+            return -1;
+        }
         return Integer.MIN_VALUE;
     }
 
--- a/Core/src/main/java/com/plotsquared/core/command/Visit.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Visit.java
@@ -348,20 +348,22 @@ public class Visit extends Command {
             case 1 -> {
                 completions.addAll(
                         TabCompletions.completeAreas(args[1]));
+                completions.addAll(TabCompletions.asCompletions("last"));
                 if (args[1].isEmpty()) {
                     // if no input is given, only suggest 1 - 3
                     completions.addAll(
-                            TabCompletions.asCompletions("1", "2", "3", "last"));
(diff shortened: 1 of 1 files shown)


dordsor21:
What if as well as adding `last` we allowed adding a `-` to the number to indicate the number-from-last to visit?

RedstoneFuture:
> What if as well as adding `last` we allowed adding a `-` to the number to indicate the number-from-last to visit?

Do you mean the argument `last` or `n` is the same as `-1`. And `-2` is the second plot from behind?

**Example:**
Total plots: 4x

- `/p v last` = `/p v -1` --> `/p v 4`
- `/p v -2` --> `/p v 3`

dordsor21:
I'd say `last` is still the very last plot, and then `-1` is the plot before `last`. So `-0` is basically `last`

RedstoneFuture:
> I'd say `last` is still the very last plot, and then `-1` is the plot before `last`. So `-0` is basically `last`

Ah ok. Yea, both versions make sense.

grafik

grafik

SirYwell:
I'm not sure about the negative values. It might make sense for us, but I don't know if it is user-friendly. Python has a similar feature for lists, but there `-1` is tne last element. Having `-0` looks very confusing.

Maybe it would be better to have a flag (e.g. `-r` for reverse?), so `/p v -r` brings you to the last plot, `/p v -r 2` or `/p v 2 -r` brings you to the second last, etc. WDYT?

RedstoneFuture:
I'll add the `-#` spelling.

But I don't think that a "reverse flag" is very user-friendly, as the command already has a very complex syntax. Especially because the order of the arguments also plays a role here, so that we can classify the individual command inputs.

`/plot visit <player> | <alias> | <plot> [area]|[#] [#]`

RedstoneFuture:
> I'll add the `-#` spelling.
> 
> But I don't think that a "reverse flag" is much more very user-friendly, as the command already has a very complex syntax. Especially because the order of the arguments also plays a role here, so that we can classify the individual command inputs. In the end, the `-#` spelling is the same in short, with the difference that not even the syntax "changes" for the user.
> 
> `/plot visit <player> | <alias> | <plot> [area]|[#] [#]`

`-#` spelling was added

Core/src/main/java/com/plotsquared/core/command/Visit.java
```diff
@@ -326,6 +317,29 @@ public CompletableFuture<Boolean> execute(
         return CompletableFuture.completedFuture(true);
     }
 
+    private boolean isInvalidPageNr(String arg) {
+        if (MathMan.isInteger(arg)) return false;
```
dordsor21, line 321: I think the code style requires all ifs etc. these have the braces applied

RedstoneFuture: Ok. I have changed it.

dordsor21 (approved):

RedstoneFuture:
I'm looking forward to the merge. I can make good use of the command. When this has been merged, I'll add a "last" to the "/p home" command too.

SirYwell (approved):

PR #4230, opened 2023-11-04
Fix:  frost walker on untrusted plots

ch4ika:


SirYwell:
I can't reproduce the issue. We already preventing what you describe two methods below via the `ice-form` flag. If you set that flag to `true`, players can form ice, as intended, but by default, that's not possible.

ch4ika:
mb then

Tamikaschu:
Setting the flag `ice-form` to `false` would also disable ice forming because of the plotbiom. It doesnt affect the plotowner when wearing boots with frostwalker though.

SirYwell:
I can see how you want this behavior to be more fine-granular. However, we're getting in the area of combinatorial explosion with that (we could add thousands of config options for things like that, and someone would still want something to work differently).

I think this is something that can be solved by a custom third-party plugin easily (i.e., you can basically use the code of this PR in your own plugin). This also allows to change behavior whenever you want, without breaking it for everyone else.

PR #4236, opened 2023-11-16
Introduce `edit-sign` flag

### LINKED ISSUE #4205: use: #signs flag allows editing signs anywhere
### Server Implementation

Paper

### Server Version

1.20.2

### Describe the bug

Players can edit and save signs on other plots without having permissions there.

This is possible because the use: #signs flag seems to allow them to do so.

### To Reproduce

1. add flag use: #signs
2. edit signs as you like and save

### Expected behaviour

Shouldn't be able to edit signs or save it.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/c71cab563c9047d7b42d67e799ac6ed2

### PlotSquared Version

PlotSquared Version 7.1.0 Premium

### Checklist

### Anything else?

_No response_

SirYwell:
## Overview

Fixes #4205 

## Description

Introduces a flag to allow sign editing separately from the `use` flag. The use flag is typically used to allow interactions with shops etc, but players should not be able to just edit every sign if the use flag is needed.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener120.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener120.java
@@ -1,0 +1,62 @@
+/*
+ * PlotSquared, a land and world management plugin for Minecraft.
+ * Copyright (C) IntellectualSites <https://intellectualsites.com>
+ * Copyright (C) IntellectualSites team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package com.plotsquared.bukkit.listener;
+
+import org.bukkit.block.Sign;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.PlayerSignOpenEvent;
+import com.plotsquared.bukkit.util.BukkitUtil;
+import com.plotsquared.core.location.Location;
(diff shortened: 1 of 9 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener120.java
```diff
@@ -0,0 +1,63 @@
...
+import org.bukkit.block.Sign;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.PlayerSignOpenEvent;
+
+/**
+ * For events since 1.20.1
+ * @since TODO
+ */
+public class PlayerEventListener120 implements Listener {
```
dordsor21, line 36: The class should be called 1201 no?

Core/src/main/java/com/plotsquared/core/plot/flag/GlobalFlagContainer.java
```diff
@@ -153,6 +154,7 @@ private GlobalFlagContainer() {
         this.addFlag(DeviceInteractFlag.DEVICE_INTERACT_FALSE);
         this.addFlag(DisablePhysicsFlag.DISABLE_PHYSICS_FALSE);
         this.addFlag(DropProtectionFlag.DROP_PROTECTION_FALSE);
+        this.addFlag(EditSignFlag.EDIT_SIGN_FALSE);
```
dordsor21, line 157: Is it worth only applying the flag for > 1.20.1?

SirYwell: I think this might cause issues when trying to access the flag value on other versions, which normally shouldn't happen, but I don't think it's worth it (and we also didn't do that for other flags previously).

dordsor21 (approved):

NotMyFault:
The flag addition should also be documented in https://github.com/IntellectualSites/documentation/tree/main/plotsquared

NotMyFault (approved):

Tamikaschu:
Wouldn't it makes sense to add a bypass permission to the flag? Admins who are not added on the plot cannot create signs with text. They all appear empty when placed since the PlayerSignOpenEvent is cancelled.

Leahcimkrob:
I'm already looking for that too. You can get around this either way if the admins trust each other for a moment. But it doesn't make sense that an admin can't edit or set the signs.

PR #4252, opened 2023-12-01
Fix: disable sign-coloring when edit-sign flag is false

### LINKED ISSUE #4247: Coloring signs
### Server Implementation

Paper

### Server Version

1.20.2

### Describe the bug

Players are able to color signs on other plots when enabling `use` for `#wall_signs` even though `edit-sign` is set to `false`. This flag is essential for several ChestShop plugins. 

### To Reproduce

- Set use flag for #wall_signs
- Set edit-sign to false
- Let another player click on the sign with dye

### Expected behaviour

Signs should only be colored by other players if edit-sign is set to true on the plot. Since #4236 was implemented to allow interactions with shops and prevent players from grieving, players should also not be able to color signs on other plots with dye by clicking on them, when edit-sings is set to false.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/edd8af15c76741de97c88c92e7878dac

### PlotSquared Version

PlotSquared version 7.2.1-SNAPSHOT

### Checklist

### Anything else?

_No response_

Tamikaschu:
## Overview

Fixes #4247 

## Description

#4236 Allows sign coloring to edit-sign flag separately from the use flag. The use flag is typically used to allow interactions with shops etc, but players should not be able to just color every sign if the use flag is needed.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -50,6 +50,7 @@ import com.plotsquared.core.plot.flag.implementations.DenyPortalsFlag;
 import com.plotsquared.core.plot.flag.implementations.DenyTeleportFlag;
 import com.plotsquared.core.plot.flag.implementations.DoneFlag;
 import com.plotsquared.core.plot.flag.implementations.DropProtectionFlag;
+import com.plotsquared.core.plot.flag.implementations.EditSignFlag;
 import com.plotsquared.core.plot.flag.implementations.HangingBreakFlag;
 import com.plotsquared.core.plot.flag.implementations.HangingPlaceFlag;
 import com.plotsquared.core.plot.flag.implementations.HostileInteractFlag;
@@ -87,6 +88,7 @@ import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.block.Sign;
 import org.bukkit.block.data.Waterlogged;
 import org.bukkit.command.PluginCommand;
 import org.bukkit.entity.ArmorStand;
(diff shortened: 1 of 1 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -207,6 +229,38 @@ public PlayerEventListener(
         this.plotListener = plotListener;
     }
 
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onPlayerDyeSign(PlayerInteractEvent event) {
+        ItemStack itemStack = event.getItem();
+        Block block = event.getClickedBlock();
+        if (block != null && block.getState() instanceof Sign) {
+            if(itemStack == null) {
```
powercasgamer, line 237:
```suggestion
            if (itemStack == null) {
```

powercasgamer (changes requested):

Tamikaschu:
I will test it out but we have to keep in mind that not only dyes but also Honeycombs and Glow Ink Sack can be applied on signs. Will they also have the dye tag?

PierreSchwang:
Yeah, forget what I said. Seems like my online source listed that as a item tag, but scavenging through the paper / minecraft 1.20.2 source does not contain such tag nor is there any other kind of tag to be respected in those Items or general sign manipulation.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -207,6 +229,38 @@ public PlayerEventListener(
         this.plotListener = plotListener;
     }
 
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onPlayerDyeSign(PlayerInteractEvent event) {
+        ItemStack itemStack = event.getItem();
+        Block block = event.getClickedBlock();
+        if (block != null && block.getState() instanceof Sign) {
+            if (itemStack == null) {
```
dordsor21, line 237: Might as well move this to just after getting the item

dordsor21:
Is there potentially a better event to use here rather than cancelling the interact? The block changes state so presumably that'll fire an event

Tamikaschu:
Did some research and asked around on PaperMC discord. It seems that there is no good alternative event to detect players applying dye to signs or waxing them apart from PlayerInteractEvent

dordsor21:
> Did some research and asked around on PaperMC discord. It seems that there is no good alternative event to detect players applying dye to signs or waxing them apart from PlayerInteractEvent. Sign coloring was added 1.16 and waxing 1.19. Should that maybe be checked in two separat events for the versions?

Were any of the items added posted 1.16? If so it will need to done using strings as older servers will not have the Materials and so it will cause issues. No need for separate events though

Tamikaschu:
> Were any of the items added posted 1.16? If so it will need to done using strings as older servers will not have the Materials and so it will cause issues. No need for separate events though

No but if we would cancel PlayerInteractEvent when a player applies honey to a sign in versions below 1.20 it would cancel the Interact even though waxing is not even possible due to the version

dordsor21:
In that case just have a version check when building the list of items and only add depending of version. No need to do that for "new" items though as it's all strings

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -175,6 +177,32 @@ public class PlayerEventListener implements Listener {
             Material.WRITABLE_BOOK,
             Material.WRITTEN_BOOK
     );
+    private final Set<String> DYES;
```
SirYwell, line 180: This should be static final, and in best case you can first create the mutable set and then use Set.copyOf to store an immutable variant

SirYwell (approved):
Looks good, thanks.

PR #4259, opened 2023-12-03
feat: add HasOwner PlotFilter filter

### LINKED ISSUE #3831: Add "hasOwner" method to PlotQuery

Title says it all really. Would be nice to be able to filter plots that either are or aren't owned.

Currently using `Plot#hasOwner` on every plot returned from my current query to check if it has an owner. I'm not sure if having this method in the query itself would affect performance at all, but it would certainly be neater.

### Anything else?

_No response_

dordsor21:
- closes #3831

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/query/PlotQuery.java
+++ b/Core/src/main/java/com/plotsquared/core/util/query/PlotQuery.java
@@ -219,13 +219,13 @@ public final class PlotQuery implements Iterable<Plot> {
         return this.addFilter(new OwnersIncludeFilter(owner.getUUID()));
     }
 
-
     /**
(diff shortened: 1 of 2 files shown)


SirYwell (approved):

Core/src/main/java/com/plotsquared/core/util/query/PlotQuery.java
```diff
@@ -219,6 +219,16 @@ public static PlotQuery newQuery() {
         return this.addFilter(new OwnersIncludeFilter(owner.getUUID()));
     }
 
+
```
Citymonstret, line 222:
```suggestion
```

PR #4261, opened 2023-12-04
feat: allow for custom plot limit handling

PierreSchwang:
## Overview
As the title states, plot limits are now not limited to permission nodes but also allowed to be overridden in an event. This helps integrating other systems, where plots may be added via vouchers etc.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/events/PlayerPlotLimitEvent.java
+++ b/Core/src/main/java/com/plotsquared/core/events/PlayerPlotLimitEvent.java
@@ -46,7 +46,7 @@ public class PlayerPlotLimitEvent {
      * @param limit The amount of plots a player may claim. Must be {@code 0} or greater.
      * @since TODO
      */
-    public void setLimit(@NonNegative final int limit) {
+    public void limit(@NonNegative final int limit) {
         if (limit < 0) {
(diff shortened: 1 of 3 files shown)


Core/src/main/java/com/plotsquared/core/events/PlayerPlotLimitEvent.java
```diff
@@ -0,0 +1,77 @@
...
+        this.limit = limit;
+    }
+
+    /**
+     * Overrides the previously calculated or set plot limit for {@link #player()}.
+     *
+     * @param limit The amount of plots a player may claim. Must be {@code 0} or greater.
+     * @since TODO
+     */
+    public void setLimit(@NonNegative final int limit) {
```
Citymonstret, line 49: The name should probably be `limit` if we're using fluent accessors elsewhere in the class?

PierreSchwang: Totally, yeah - Just used IntelliJ's builder and ignored everything else

NotMyFault (approved):

dordsor21 (approved):

PR #4266, opened 2023-12-12
feat: configurable accounting for bedrock layer when setting components

dordsor21:
- Add configuration option to force plot components to be set only above bedrock level
 - Account for build height < gen height where it is not wanted for components to be set at/below bedrock

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
@@ -1454,6 +1454,8 @@ public abstract class PlotArea implements ComponentLike {
 
     /**
      * Get the maximum height that changes to plot components (wall filling, air, all etc.) may operate to
+     *
+     * @since TODO
      */
     public int getMaxComponentHeight() {
         return this.maxBuildHeight;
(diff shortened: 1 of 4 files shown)


NotMyFault (approved):

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
```diff
@@ -1452,6 +1452,20 @@ protected void setDefaultHome(BlockLoc defaultHome) {
         this.defaultHome = defaultHome;
     }
 
+    /**
+     * Get the maximum height that changes to plot components (wall filling, air, all etc.) may operate to
+     */
+    public int getMixComponentHeight() {
```
SirYwell, line 1458:
```suggestion
    public int getMaxComponentHeight() {
```

is this method even used?

SirYwell (dismissed):
New methods are missing `@since` tags

Core/src/main/java/com/plotsquared/core/plot/PlotArea.java
dordsor21: Unused but added to match the new min method for parity

PierreSchwang (approved):

SirYwell (approved):

PR #4267, opened 2023-12-12
feat: Allow admin done permission to override requiring plot complexity calculation

MrJoshuaT:
## Overview
Let admins always be able to set plots as done

## Description
The permission `plots.admin.command.done` already exists but is only used to check if you have plot perms, it's very useful to allow admins with the permissions to not require doing plot calculations.

Also removed `PlotSquared.platform().expireManager() == null` condition on same line because it can never return null.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Done.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Done.java
@@ -94,7 +94,7 @@ public class Done extends SubCommand {
                 TagResolver.resolver("plot", Tag.inserting(Component.text(plot.getId().toString())))
         );
         final Settings.Auto_Clear doneRequirements = Settings.AUTO_CLEAR.get("done");
-        if (PlotSquared.platform().expireManager() == null || doneRequirements == null) {
+        if (PlotSquared.platform().expireManager() == null || doneRequirements == null || player.hasPermission(Permission.PERMISSION_ADMIN_COMMAND_DONE)) {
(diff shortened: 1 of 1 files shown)


Core/src/main/java/com/plotsquared/core/command/Done.java
```diff
@@ -94,7 +94,7 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
                 TagResolver.resolver("plot", Tag.inserting(Component.text(plot.getId().toString())))
         );
         final Settings.Auto_Clear doneRequirements = Settings.AUTO_CLEAR.get("done");
-        if (PlotSquared.platform().expireManager() == null || doneRequirements == null) {
+        if (doneRequirements == null || player.hasPermission(Permission.PERMISSION_ADMIN_COMMAND_DONE)) {
```
dordsor21, line 97: Why is null check on the expire manager being removed??

MrJoshuaT: Because `PlotPlatform.expireManager` is marked as NotNull, so the check should not be required

MrJoshuaT: I've put the check back in as it seems sometimes not checked to be null and sometimes is

dordsor21 (approved):

SirYwell (approved):

PR #4268, opened 2023-12-14
PostPlotClearEvent added

David-M-GitHub:
## Overview
PostPlotClearEvent added

Fixes https://github.com/IntellectualSites/PlotSquared/issues/4288


dordsor21:
I think it could be worth making this a player event, as the clearing player could be a moderator/admin, not just the plot owner for example

Core/src/main/java/com/plotsquared/core/events/post/PostPlotClearEvent.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.plotsquared.core.events.CancellablePlotEvent;
+import com.plotsquared.core.events.PlotEvent;
+import com.plotsquared.core.events.Result;
+import com.plotsquared.core.plot.Plot;
+import com.plotsquared.core.plot.PlotId;
+
+/**
+ * Called after a {@link Plot} was deleted.
+ */
```
PierreSchwang, line 29: `@since TODO`

Core/src/main/java/com/plotsquared/core/events/post/PostPlotClearEvent.java
```diff
@@ -0,0 +1,66 @@
...
+public class PostPlotClearEvent extends PlotEvent implements CancellablePlotEvent {
```
PierreSchwang, line 30: No reason to make this event cancellable. There is nothing to cancel.

Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
```diff
@@ -178,6 +178,12 @@ public PlotClearEvent callClear(Plot plot) {
         return event;
     }
 
+    public PlotClearEvent callPostClear(Plot plot) {
+        PlotClearEvent event = new PlotClearEvent(plot);
```
PierreSchwang, line 182: Wrong event called.

Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
```diff
@@ -178,6 +178,12 @@ public PlotClearEvent callClear(Plot plot) {
         return event;
     }
 
+    public PlotClearEvent callPostClear(Plot plot) {
```
PierreSchwang, line 181: Wrong Event returned

PierreSchwang, line 181: Rather name it `callPostPlotClear`

PierreSchwang (changes requested):
And don't delete the checklist. It's there for a reason (e.g. describing the usage of the since annotation)

Core/src/main/java/com/plotsquared/core/events/post/PostPlotClearEvent.java
```diff
@@ -0,0 +1,56 @@
...
+import com.plotsquared.core.events.CancellablePlotEvent;
+import com.plotsquared.core.events.PlotEvent;
+import com.plotsquared.core.events.Result;
+import com.plotsquared.core.plot.Plot;
+import com.plotsquared.core.plot.PlotId;
+
+/**
+ * Called after a {@link Plot} was cleared.
+ *
+ * @since 7.2.1
```
dordsor21, line 30: This should be `TODO` as we have a task that will go through and replace them automagically

dordsor21:
I would still like this to be extend PlotPlayerEvent as the player executing the clear would be useful to know

Core/src/main/java/com/plotsquared/core/events/post/PostPlotClearEvent.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getWorld() {
+        return getPlot().getWorldName();
+    }
+
+    /**
+     * Get the PlotPlayer.
+     *
+     * @return PlotPlayer
+     */
+    public PlotPlayer<?> getPlotPlayer() {
```
dordsor21, line 67: These methods do not need to be here, no need to add extra methods that the API user can do themselves

Core/src/main/java/com/plotsquared/core/events/post/PostPlotClearEvent.java
```diff
@@ -0,0 +1,62 @@
...
+    public PostPlotClearEvent(final PlotPlayer<?> plotPlayer, final Plot plot) {
+        super(plotPlayer, plot);
+    }
+
+    /**
+     * Get the PlotId.
+     *
+     * @return PlotId
+     */
+    public PlotId getPlotId() {
```
dordsor21, line 49: These two methods should still not be here tbh

PierreSchwang (approved):

NotMyFault (approved):

PR #4295, opened 2024-01-17
Fix home command reset

### LINKED ISSUE #4294: /plot sethome none doesnt reset to default location
### Server Implementation

Paper

### Server Version

1.20

### Describe the bug

Resetting the plothome with /plot sethome none doesnt reset the home to the default location

### To Reproduce

- claim a plot
- set plothome (/p sethome)
- reset plothome (/p sethome none)

### Expected behaviour

Home position should be reset to default location.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/a7fcb735f056463b9aca86f128562df5

### PlotSquared Version

7.2.1-Premium

### Checklist

### Anything else?

_No response_

OneLiteFeather:
## Overview
This PR should fix the missing reset happening while /p sethome none was executed. I had help with @TheMeinerLP which found the underlying solution together with me, as I am not a great developer with basic sql knowledge yet.

Fixes #4294 

## Description
- We fixed the setposition method with the default value and added a comment too to give other developers a hint to look it up on the statement. Also we added a location not null check because the issue lies within the equals implementation.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/Plot.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/Plot.java
@@ -1483,7 +1483,7 @@ public class Plot {
      */
     public void setHome(BlockLoc location) {
         Plot plot = this.getBasePlot(false);
-        if (BlockLoc.ZERO.equals(location) || BlockLoc.MINY.equals(location)) {
+        if (location != null && (BlockLoc.ZERO.equals(location) || BlockLoc.MINY.equals(location))) {
             return;
         }
         plot.getSettings().setPosition(location);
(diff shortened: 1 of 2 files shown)


Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -1483,7 +1483,7 @@ private Location toHomeLocation(Location bottom, BlockLoc relativeHome) {
      */
     public void setHome(BlockLoc location) {
         Plot plot = this.getBasePlot(false);
-        if (BlockLoc.ZERO.equals(location) || BlockLoc.MINY.equals(location)) {
+        if (location != null && (BlockLoc.ZERO.equals(location) || BlockLoc.MINY.equals(location))) {
```
PierreSchwang, line 1486: The null check should not be required as we don't call equals on the nullable object itself

PierreSchwang (approved):
The one comment, which is not a dealbreaker but not really required.

No idea who introduced the default value as "DEFAULT" instead of just making the column nullable?

NotMyFault (approved):

PR #4336, opened 2024-02-13
fix: allow trusted players to remove books from lecterns while lectern-read-book flag is true

### LINKED ISSUE #4323: lectern-read-book flag prevents the removal of a book from lectern for all players
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

According to the wiki, the new `lectern-read-book` flag is intended to prevent players from removing books from the lectern ([Wiki Plot flag link](https://intellectualsites.gitbook.io/plotsquared/features/plot-flags#:~:text=True-,lectern%2Dread%2Dbook,-Prevent%20players%20taking)).

However, it also prevents trust/add players and plot owners from removing the book from the lectern.

As soon as a book has been added, it remains permanently in the lectern and the book can no longer be removed using the button. The only solution is to break the lectern.

Apparently the "Take Book" button/event is intercepted and terminated for all players, regardless of their authorization.

### To Reproduce

1. set `lectern-read-book flag` to `true`
2. place lectern
3. place book into lectern
4. open book in lectern
5. try to remove the book with the "Take Book" button

### Expected behaviour

I would expect that as an add/trust/owner player you can still take the book out of the lectern and this flag only applies to other players who don't have permission on the plot.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/4da2f0cb4f48480889ee49fa3a6da29e

### PlotSquared Version

7.3.2-Premium

### Checklist

### Anything else?

_No response_

Tamikaschu:
Fixes #4323

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -1921,7 +1921,9 @@ public class PlayerEventListener implements Listener {
 
     @EventHandler
     public void onPlayerTakeLecternBook(PlayerTakeLecternBookEvent event) {
-        Location location = BukkitUtil.adapt(event.getPlayer().getLocation());
+        Player player = event.getPlayer();
+        BukkitPlayer pp = BukkitUtil.adapt(player);
+        Location location = pp.getLocation();
         PlotArea area = location.getPlotArea();
         if (area == null) {
             return;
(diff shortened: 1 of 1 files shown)


dordsor21:
I think this should probably be a separate flag or something tbh, or come under a different flag as it's altering the block

Tamikaschu:
> I think this should probably be a separate flag or something tbh, or come under a different flag as it's altering the block

The flags description is currently saying: "Prevent players taking books from lecterns.". Entering and actually reading the books inside a lectern has to be done by setting the use flag for lecterns. I think that this is more of a name issue right now or what else comes up to your mind?

dordsor21 (approved):

NotMyFault (approved):

PR #4352, opened 2024-02-20
feat: add a getNextHighesBlock method to the plotspawn teleportation

### LINKED ISSUE #4351: Refinement to the Plotspawn Teleportation

Add a `getNextHighesBlock` method, which checks for the next air (safe) block when teleporting to a blocked (unsafe) plotspawn instead of the highest possible block on the y coordinate.

Otherwise players wont be able to set their spawn on carpets, slabs, etc. when a block is somewhere above them.

Set the plotspawn one block higher than the carpet, slab, etc. or get used to be teleported to an unwanted location

### Anything else?

_No response_

Tamikaschu:
Fixes #4351


Tamikaschu:
@OneLiteFeather Sorry, the Bugfix label applied again :(

PierreSchwang:
What I'm thinking about while reviewing - shouldn't a check be made, that there are at least 2 blocks of space for the player being able to stand up? In the fallback case we can't make sure that's possible, but that's why it's the fallback.

Tamikaschu:
> What I'm thinking about while reviewing - shouldn't a check be made, that there are at least 2 blocks of space for the player being able to stand up? In the fallback case we can't make sure that's possible, but that's why it's the fallback.

Technically the top 2 Blocks should be checked yes. The player would Spawn in the Swimming Animation right now if there is a full block above. Currently it would also allow the player spawn in such spawns (see screenshots), which should also be possible, since a player can sneak under a half slab/stand on carpet. 
image
image

PierreSchwang:
I guess carpets are totally alright - so about 1.85 blocks space or so?

Tamikaschu:
> I guess carpets are totally alright - so about 1.85 blocks space or so?

Carpets on the Plotspawn are counted as a blocked spawn, since you cannot really differ between half and full blocks. If there would be a block somewhere above (e.g. 100 blocks above) you would get teleported to that one. This is why i created the issue in first place and suggested replacing the getHighestBlock with a getNextHighestBlock method to make it more user friendly. I get that checking for the next 2 air blocks would be a more safer solution, still it would block the spawns like in the pictures above, even though there would be enough space for a player to fit in.

OneLiteFeather:
What about adding this "safety gap" as an configurable option too? In 1.21 we might get in trouble with entities which can change sizes, also for players. 2 blocks would be enough for the most servers I guess but those playing around with these entity sizes would still get stuck in the blocks. You might want to do a different PR regarding that for 1.21 later or add it here.

Tamikaschu:
> What about adding this "safety gap" as an configurable option too? In 1.21 we might get in trouble with entities which can change sizes, also for players. 2 blocks would be enough for the most servers I guess but those playing around with these entity sizes would still get stuck in the blocks. You might want to do a different PR regarding that for 1.21 later or add it here.

You would always have to spawn in the Player at a BlockLocation. If the plotspawn should support the players size you would always have to round up the blocks which need to be air in order to spawn in the player safe. Adding a safety gab wouldnt be possible that easy since the plot location only supports integer values.

Tamikaschu:
Adding support to different sizes would probably also require a different way of checking if the plotspawn if blocked due to not only a possible change in height but also in width. Therefore some more adjustments have to be taken into account.

Tamikaschu:
Closing due to yet impossible expectations in adjustments

PierreSchwang:
I've would've been totally fine without a two-block-height check tbf

PR #4356, opened 2024-02-22
fix-count-entities

### LINKED ISSUE #4339: Plot limit can be overwritten.
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

The limitation can be circumvented by having the entity on a merged property where the road previously existed and walking around, so the plot cap is always recalculated and the transition to plot 1 to 2 where the road was ignored.

Fawe: 2.9.1-SNAPSHOT-659;c93ec87

### To Reproduce

1. To reproduce the whole thing. Do you need at least 2 plots that merged.
2. A villager should be spawned where the road used to be.
3. Hit this Villager so that it moves around a bit as the counts are then recalculated.
4. As soon as step 3 happens, another mob or misc can be spawned/set.
5. The whole thing can be done permanently without limits.

Detailed instructions can be seen in the video above.

### Expected behaviour

The limitation should be maintained and not be recalculated again and again as soon as a mob runs over the merged property.
In general it should not be possible for the cap to be overwritten.

### Screenshots / Videos

How to bypass cap:
https://www.youtube.com/watch?v=KCqGM4RalOE&feature=youtu.be

ArmorStands are ignored by caps and can be placed infinitely.
https://www.youtube.com/watch?v=0YuhzfdSx0I

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/756733ea33a144f7b23cf2e472a6b4dc

### PlotSquared Version

7.3.4-Premium

### Checklist

### Anything else?

_No response_

ch4ika:
Fixes #4339

## Description
I don't know what the code did exactly before, but it definitely didn't work for merged plots.
This change works and ensures that the caps are added up correctly.

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntityEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntityEventListener.java
@@ -195,7 +195,7 @@ public class EntityEventListener implements Listener {
             }
             return;
         }
-        if (BukkitEntityUtil.checkEntity(entity, plot.getBasePlot(false))) {
+        if (BukkitEntityUtil.checkEntity(entity, plot)) {
             event.setCancelled(true);
         }
     }
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
@@ -41,9 +41,7 @@ import com.sk89q.worldedit.bukkit.BukkitWorld;
 import com.sk89q.worldedit.regions.CuboidRegion;
 import com.sk89q.worldedit.world.block.BaseBlock;
 import com.sk89q.worldedit.world.block.BlockTypes;
-import io.papermc.lib.PaperLib;
 import org.bukkit.Bukkit;
-import org.bukkit.Chunk;
 import org.bukkit.World;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
@@ -51,9 +49,7 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 
 import static com.plotsquared.core.util.entity.EntityCategories.CAP_ANIMAL;
 import static com.plotsquared.core.util.entity.EntityCategories.CAP_ENTITY;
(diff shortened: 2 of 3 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
```diff
@@ -88,73 +84,29 @@ public boolean handleClear(
...
-                        if (X > bx && X < tx && Z > bz && Z < tz) {
-                            count(count, entity);
-                        } else {
-                            Plot other = area.getPlot(BukkitUtil.adapt(location));
-                            if (plot.equals(other)) {
-                                count(count, entity);
+        for (final CuboidRegion region : plot.getRegions()) {
+            for (int x = region.getMinimumPoint().getX() >> 4; x <= region.getMaximumPoint().getX() >> 4; x++) {
+                for (int z = region.getMinimumPoint().getZ() >> 4; z <= region.getMaximumPoint().getZ() >> 4; z++) {
+                    world.getChunkAtAsync(x, z).thenAccept(chunk -> {
```
SirYwell, line 101: loading chunks async will break things; `count` is returned before the chunk is loaded (that was flawed before too, I guess). For very large plots, this might be important, but that means we either need to come up with something clever or load all chunks before starting counting.

Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -1220,15 +1220,13 @@ public boolean removeFlag(final @NonNull PlotFlag<?, ?> flag) {
...
-            count[CAP_VEHICLE] += result[CAP_VEHICLE];
-            count[CAP_MISC] += result[CAP_MISC];
-        }
+        int[] result = this.regionManager.countEntities(this);
+        count[CAP_ENTITY] += result[CAP_ENTITY];
+        count[CAP_ANIMAL] += result[CAP_ANIMAL];
+        count[CAP_MONSTER] += result[CAP_MONSTER];
+        count[CAP_MOB] += result[CAP_MOB];
+        count[CAP_VEHICLE] += result[CAP_VEHICLE];
+        count[CAP_MISC] += result[CAP_MISC];
```
SirYwell, line 1229: It doesn't make sense to copy the array like that. Just return it directly.

Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
```diff
@@ -88,73 +84,29 @@ public boolean handleClear(
...
+                        final Entity[] entities = chunk.getEntities();
```
SirYwell, line 102: Not all entities in the chunk also have to be in the plot. This needs to be filtered again.

SirYwell (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
ch4ika: I thought about it and played around, but unfortunately I can't really think of a simple idea that would allow for performance-friendly loading for larger plots without making any really big changes.

SirYwell: I think the short-term best solution is to just ignore unloaded chunks. That will result in the same behavior as the current code if I didn't miss anything. It also isn't a problem for a vast majority of servers I guess.

Long-term, we probably need to rethink the concept or come up with a solution to store the entity count of unloaded chunks somehow.

Removing entities later likely is a bad idea, as we would need to track how the entity was created (i.e. via spawn eggs, you'd need to give the spawn egg back to the player, etc...).

ch4ika: Thanks for the feedback. As an alternative, I'm also thinking of caching the whole thing somehow.

Because the solution with checkAsync works, but as you already said, too many additional checks would have to be made and the events would not really be cancelable.

The whole thing is breaking my head. If anyone wants to give that, I would be grateful. Otherwise, I would try it again as soon as I have time.

SirYwell: You could probably just open an issue for it so we can keep track of it. I don't think it's worth to spend too much time on it right now, the fix for merged plots is far more important.

ch4ika: The problem is I had already made several tickets about it, but @OneLiteFeather kept marking them as invalid because she didn't understand the problem and wasn't open to discussing it.

OneLiteFeather: We are open for discussions but not mixing up contexts or creating incorrect issues. Also we do no private support including voicechat and discord dms.

ch4ika: Anyway, you can now be sure that this problem exists and that you can reproduce it without any problems. I'll write another proper ticket in a few days to record it.

SirYwell: I created an issue to keep track of it: #4362

Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntityEventListener.java
```diff
@@ -195,9 +196,12 @@ public void creatureSpawnEvent(CreatureSpawnEvent event) {
             }
             return;
         }
-        if (BukkitEntityUtil.checkEntity(entity, plot.getBasePlot(false))) {
-            event.setCancelled(true);
-        }
+
+        BukkitEntityUtil.checkEntityAsync(entity, plot.getBasePlot(false)).thenAcceptAsync(toRemove -> {
+            if(toRemove) {
```
dordsor21, line 201: code style

Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitRegionManager.java
```diff
@@ -88,79 +87,45 @@ public boolean handleClear(
...
-                            count(count, entity);
-                        } else {
-                            Plot other = area.getPlot(BukkitUtil.adapt(location));
-                            if (plot.equals(other)) {
-                                count(count, entity);
-                            }
+        for (final CuboidRegion region : plot.getRegions()) {
+            for (int x = region.getMinimumPoint().getX() >> 4; x <= region.getMaximumPoint().getX() >> 4; x++) {
+                for (int z = region.getMinimumPoint().getZ() >> 4; z <= region.getMaximumPoint().getZ() >> 4; z++) {
+                    Chunk chunk = world.getChunkAt(x, z);
```
dordsor21, line 104: Can this method actually be called async?

Core/src/main/java/com/plotsquared/core/plot/Plot.java
```diff
@@ -1219,17 +1212,11 @@ public boolean removeFlag(final @NonNull PlotFlag<?, ?> flag) {
      * @see RegionManager#countEntities(Plot)
      */
     public int[] countEntities() {
```
dordsor21, line 1214: Should probably deprecate the "old" sync methods if moving to async

PR #4359, opened 2024-02-23
fix-silk-touch-with-instabreak

### LINKED ISSUE #4358: Silk Touch enchantment does not work with instabreak flag
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

The enchantment Silk Touch cannot be used with the instabreak flag

### To Reproduce

1. Enchant a tool with silk touch
2. Set instabreak flag to true
3. Break a block (e.g. stone) with the tool

### Expected behaviour

A clean stone block should drop instead of a cobblestone

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/861ed71f0e034f50b1e06a8219e498a8

### PlotSquared Version

7.3.5-Premium

### Checklist

### Anything else?

_No response_

ch4ika:
Fixes #4358

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -670,7 +670,7 @@ public class BlockEventListener implements Listener {
                 BlockBreakEvent call = new BlockBreakEvent(block, player);
                 Bukkit.getServer().getPluginManager().callEvent(call);
                 if (!call.isCancelled()) {
-                    event.getBlock().breakNaturally();
+                    event.getBlock().breakNaturally(event.getItemInHand());
                 }
             }
             // == rather than <= as we only care about the "ground level" not being destroyed


OneLiteFeather:
I tested your changes and your changes do not work the way you wanted. There is still no stone but cobblestone in a claimed plot with instabreak and with a silk touch pickaxe. I guess if you only give the method breakNaturally an itemstack without considering enchants it might won't work.

Tamikaschu:
> yes I already seen the docs but I tested your version in game and it is not working with the silk touch enchantment, it still drops only cobblestone. You will likely need to change the drops or check if there is a wrong method / check.

The changes work perfectly fine as shown in the [Video](https://www.youtube.com/watch?v=DOSPyFSMM9U)

OneLiteFeather:
Looks like I took the wrong branch, but what needs to be considered is that the drops will only happen if you have a tool in your hand, which changes the way instabreak works. Instabreaks actual behaviour is to break blocks with your bare hand too which is dropping cobblestone. But at the changes cobblestone won't drop with bare hands.

Tamikaschu:
> Looks like I took the wrong branch, but what needs to be considered is that the drops will only happen if you have a tool in your hand, which changes the way instabreak works. Instabreaks actual behaviour is to break blocks with your bare hand too which is dropping cobblestone. But at the changes cobblestone won't drop with bare hands.

"Set to `true` to allow blocks to be instantaneously broken in survival mode"
The flags description doesn't say anything about the intention of the flag being instantaneously breaking blocks with your bare hand. Therefore adapting the survival mechanic where u have to use the specific tool is the only logic conclusion. If someone really wanted to use this flag for the survival mode, which the flag is made for, it would break the game, since players would be able to obtain unbreakable blocks with hardness -1 (Bedrock).

SirYwell:
As this changes existing behavior, I'd like @IntellectualSites/plotsquared-team to weigh in too. I'm generally fine with it, as it makes sense this way. But it might make sense to put it behind a configurable option (in settings.yml) so we don't break any existing setups. (I guess we could need a new section for flags there then)

nicolube:
@ch4ika Pls add an Option so that it can be merged

NotMyFault:
> As this changes existing behavior, I'd like @IntellectualSites/plotsquared-team to weigh in too. I'm generally fine with it, as it makes sense this way. But it might make sense to put it behind a configurable option (in settings.yml) so we don't break any existing setups. (I guess we could need a new section for flags there then)

This is a reasonable explanation. I'll approve once a config option has been added.

PR #4369, opened 2024-03-06
Improve teleport behavior for home command

### LINKED ISSUE #4351: Refinement to the Plotspawn Teleportation

TheMeinerLP:
## Overview
Fixes #4351

## Description
This PR adds another setting called `sized-based` in the teleport section. This setting activates an additional check at the home command in which the block size is queried. If this is smaller than a slab, the player will be teleported in between instead of on top of it.

```[tasklist]
### Submitter Checklist
```

### Additonal informations
Thanks to @Tamikaschu for the base idea and the initial pr of #4352

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitUtil.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitUtil.java
@@ -262,6 +262,11 @@ public class BukkitUtil extends WorldUtil {
     }
 
     @Override
+    public boolean isSmallBlock(Location location) {
+        return adapt(location).getBlock().getBoundingBox().getHeight() < 0.25;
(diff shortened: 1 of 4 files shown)


TheMeinerLP:
Tested cases:
image
image
image

dordsor21 (approved):

PierreSchwang (approved):

PR #4302, opened 2024-01-20
Use MUSIC soundcategory for plot music flag

PierreSchwang:
## Overview
Fixes #4301

```[tasklist]
### Submitter Checklist
- [x] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [x] Ensure that the pull request title represents the desired changelog entry.
- [x] New public fields and methods are annotated with `@since TODO`.
- [x] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md).
```

PR #4391, opened 2024-04-07
fix plot music volume to span full plot

### LINKED ISSUE #4316: music flag not working.
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

Even if you use the flag set music ~ command, background music is not set.

### To Reproduce

.

### Expected behaviour

.

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Plot Debugpaste

.

### PlotSquared Version

PlotSquared-7.3.2-Premium

### Checklist

### Anything else?

_No response_

PierreSchwang:
## Overview
Fixes #4316

## Description
I missed a Float.MAX_VALUE in https://github.com/IntellectualSites/PlotSquared/pull/4302 - which this PR is a follow-up to and addresses the mistake.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
@@ -326,7 +326,7 @@ public class BukkitPlayer extends PlotPlayer<Player> {
             return;
         }
         this.player.playSound(BukkitUtil.adapt(location), Sound.valueOf(BukkitAdapter.adapt(id).name()),
-                SoundCategory.MUSIC, 1f, 1f
+                SoundCategory.MUSIC, Float.MAX_VALUE, 1f
         );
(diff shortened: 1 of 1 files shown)


TheMeinerLP (approved):
Makes sense to me, LGTM.
Would it make sense to simply add this to the config?

PierreSchwang:
> Makes sense to me, LGTM. Would it make sense to simply add this to the config?

No, that's nothing to be configured. The volume parameter set to it's max value just makes sure the music is played across the whole plot (even across plots if other players would be the recipients). Otherwise, you could here the music only in the middle of the plot with a margin of about 5-10 blocks. It's intentional to utilize the max range to make sure megaplots are also working accordingly. making this configurable or based of the plot size adds no benefit but rather confusion

SirYwell (approved):

PR #4395, opened 2024-04-14
Make builds reproducible

NotMyFault:
[reproducible-builds.org](https://reproducible-builds.org/) explains the reasons why reproducible builds are desirable. As of gradle 3.4, we can make use of a [dedicated](https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives) task.

Additionally, the PR suppresses the hidden timestamp in javadocs' html code, ensuring this PR applies to -javadoc.jar's too.
Using diffoscope, the change proposed can be validated to produce identical jars.

### DIFF
--- a/Bukkit/build.gradle.kts
+++ b/Bukkit/build.gradle.kts
@@ -113,5 +113,6 @@ tasks {
         opt.encoding("UTF-8")
         opt.keyWords()
         opt.addStringOption("-since", isRelease)
+        opt.noTimestamp()
     }
 }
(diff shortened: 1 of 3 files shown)


SirYwell:
This is a good step.

From my understanding, this does not guarantee reproducibility e.g. when using a different JDK (version) to compile, but I guess that gets difficult unless we want to provide something like a nix flake.

NotMyFault:
The latter isn't really a "hard goal" for https://reproducible-builds.org/, but this approach gets us closely to reproducible central builds, i.e., but having a solid order of classes and omitting the timestamp. 

This would get us closely into the right direction (for a MC plugin overkill but I'd like to play with it).

SirYwell (approved):

PR #4418, opened 2024-05-04
perf: Avoid expensive Plot#getOwner calls in Plot#getOwners

Vrganj:
## Description
Refactored Plot#getOwners and reduced the number of Plot#getOwner calls because
they go through a half dozen flag containers to check whether the plot is server-owned.

I'd like input on the first null check. The behavior is the same as the old implementation,
but it seems a little illogical. If there are 2 plots, one without an owner - getOwners would
return an empty set if called on the ownerless one, and 1 element on the other, creating
inconsistency.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/plot/Plot.java
+++ b/Core/src/main/java/com/plotsquared/core/plot/Plot.java
@@ -643,35 +643,25 @@ public class Plot {
     }
 
     /**
-     * Gets a immutable set of owner UUIDs for a plot (supports multi-owner mega-plots).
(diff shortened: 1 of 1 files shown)


SirYwell:
I assume the null check for the connected plots is rather an assert. The owner is only null if the plot is not claimed; and if it isn't claimed but merged with other plots at the same time, something is wrong already.

Vrganj:
Ah, then I suppose one of null checks can be removed.

SirYwell (approved):

NotMyFault (approved):

PR #4420, opened 2024-05-05
perf: Faster world lookup

Vrganj:
## Description
Make world lookup faster using an IdentityHashMap and computeIfAbsent.

```[tasklist]
### Submitter Checklist
```


NotMyFault (approved):

dordsor21:
I do not think we want to be forcing BukkitWorld objects to be kept in memory, as if a world is unloaded and loaded again we'll have garbage objects kept in the map that won't be removed. Elsewhere (somewhere in P2 and/or FAWE) we reference worlds with a WeakReference to ensure we do not do this.

dordsor21:
See #4439

dordsor21:
Gonna close this as stale

PR #4429, opened 2024-05-21
Replace isEnabled check with null check to fix vault interactions

Test-Account666:
## Overview
In some cases, the `isEnabled` check will cause the NullEconHandler to be used - Even though everything will work out.

## Description
I noticed that PlotSquared will use the NullEconHandler in some scenarios.
One of them is when the plugin providing the Economy is depending on Vault and PlotSquared.

I fixed this by replacing the Vault `isEnabled` check with a null check.

The VaultPermissionHandler was also affected, so I also replaced the `isEnabled` check with a null check.

Note that I kept the Vault `isEnabled` check in the `ServerListener`, since Vault is going to be enabled at this point.

```[tasklist]
### Submitter Checklist
(At least I think I did it correctly?)
(There are no new public fields and methods)
```


Test-Account666:
Nevermind. It seems like I haven't tested properly.

Merging this would cause permissions to break completely.

PR #4432, opened 2024-05-21
Replace `isPluginEnabled` check with null check. Fixes vault economy interactions.

Test-Account666:
## Overview
In some cases, the `isPluginEnabled` check will cause the NullEconHandler to be used - Even though everything will work out.

## Description
I noticed that PlotSquared will use the `NullEconHandler` in some scenarios.
One of them is when the plugin providing the Economy is depending on Vault and PlotSquared.

I fixed this by replacing the Vault `isPluginEnabled` check with a null check.

The VaultPermissionHandler is also affected, although I don't know what's causing the issue here.
I did not replace the `isPluginEnabled` this time, as it breaks permissions otherwise (See previous pull request).

Note that I kept the Vault `isPluginEnabled`  check in the `ServerListener`, since Vault is going to be enabled at this point.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/inject/BukkitModule.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/inject/BukkitModule.java
@@ -133,7 +133,7 @@ public class BukkitModule extends AbstractModule {
     @Provides
     @Singleton
     @NonNull EconHandler provideEconHandler() {
-        if (!Settings.Enabled_Components.ECONOMY || !Bukkit.getPluginManager().isPluginEnabled("Vault")) {
+        if (!Settings.Enabled_Components.ECONOMY || Bukkit.getPluginManager().getPlugin("Vault") == null) {
             return EconHandler.nullEconHandler();
         }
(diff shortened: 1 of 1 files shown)


TheMeinerLP:
Have you tested ?

Test-Account666:
> Have you tested ?

This time, I tested the merge command, yes.

Without enough money, the merge will fail, as expected.

With enough money, the merge will succeed and withdraw the correct amount.

I also used the unlink command and tried again, for good measure.

If I should be doing more than that, please tell me :)

dordsor21:
What exactly breaks by making the same change elsewhere?

Test-Account666:
> What exactly breaks by making the same change elsewhere, e.g. for permission handler?

I actually didn't investigate to why this happens.

Permission checking straight up doesn't work anymore.

PlotSquared thinks nobody has permission to do anything

dordsor21:
Hmm okay, so I suppose the vault permissions module is just never used then, which is why permissions continue to work?

PR #4433, opened 2024-05-22
Fix sending multiple messages of notify-enter and notify-leave flag when joining from foreign worlds

### LINKED ISSUE #3424: `notify-enter` and `notify-leave` send three messages when teleporting to plot from different world
### Server Implementation

Paper

### Server Version

1.17.1

### Describe the bug

On a Plot that has both `notify-enter` and `notify-leave` set to true, three messages instead of one.
`notify-leave` seems to remove a player from a plot again and then teleporting them back on. `notify-enter` does not have that problem.

### To Reproduce

1. Set both `notify-enter` and `notify-leave` to true on a plot.
2. teleport a player to another world! (player shouldn't have bypass permissions)
3. teleport player back on the plot.
4. three messages appear, instead of one.
5. repeat same process with only `notify-enter`, and everything works as intended.

### Expected behaviour

only the `notify-enter` message should be send.

### Screenshots / Videos

The three messages when teleporting player from different world to the plot:
image

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/4a2b6c24924845bd82fd5bc1258fa639

### PlotSquared Version

PlotSquared version 6.2.2-Premium

### Checklist

### Anything else?

_No response_

Fameless9:
## Overview

Fixes #3424

## Description

This PR combines the WorldChange and Teleport listeners. The teleport listener is triggered when a player changes worlds anyway, by combining them, plotEntry and plotExit are only called once when a player changes worlds and teleports at the same time, fixing the issue.

### Before
https://github.com/IntellectualSites/PlotSquared/assets/131914029/2be0c4dd-7222-444b-926e-97a2cbcf0d5f

### After
https://github.com/IntellectualSites/PlotSquared/assets/131914029/fb63ee4c-f35f-49ed-9c3d-648e8f325b14

### Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -122,7 +122,6 @@ import org.bukkit.event.player.AsyncPlayerChatEvent;
 import org.bukkit.event.player.AsyncPlayerPreLoginEvent;
 import org.bukkit.event.player.PlayerBucketEmptyEvent;
 import org.bukkit.event.player.PlayerBucketFillEvent;
-import org.bukkit.event.player.PlayerChangedWorldEvent;
 import org.bukkit.event.player.PlayerCommandPreprocessEvent;
 import org.bukkit.event.player.PlayerDropItemEvent;
 import org.bukkit.event.player.PlayerEvent;
@@ -533,7 +532,9 @@ public class PlayerEventListener implements Listener {
                     // to is identical to the plot's home location, and untrusted-visit is true
                     // i.e. untrusted-visit can override deny-teleport
                     // this is acceptable, because otherwise it wouldn't make sense to have both flags set
-                    if (!result && !(plot.getFlag(UntrustedVisitFlag.class) && plot
+                    if (result) {
+                        plotListener.plotEntry(pp, plot);
+                    } else if (!(plot.getFlag(UntrustedVisitFlag.class) && plot
                             .getHomeSynchronous()
                             .equals(BukkitUtil.adaptComplete(to)))) {
                         pp.sendMessage(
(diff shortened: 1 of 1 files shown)


PierreSchwang:
I can't really review on my phone - but the thing I've noticed is, that the metadata mutation was deleted and not re-added, right?

Fameless9:
The teleport listener does the same thing as the world change listener. It removes both lastPlotAccess and lastLocationAccess. One thing I don't quite get is the PERMISSION_WORLDEDIT_BYPASS part. I haven't implemented this into the teleport listener as I don't really understand what it does. Please let me know and I'll add that to the listener if necessary.

dordsor21:
> The teleport listener does the same thing as the world change listener. It removes both lastPlotAccess and lastLocationAccess. One thing I don't quite get is the PERMISSION_WORLDEDIT_BYPASS part. I haven't implemented this into the teleport listener as I don't really understand what it does. Please let me know and I'll add that to the listener if necessary.

Permissions can be per world so P2 needs to check that the player has WorkdEdit bypass permissions (or not) upon world change

Fameless9:
> > The teleport listener does the same thing as the world change listener. It removes both lastPlotAccess and lastLocationAccess. One thing I don't quite get is the PERMISSION_WORLDEDIT_BYPASS part. I haven't implemented this into the teleport listener as I don't really understand what it does. Please let me know and I'll add that to the listener if necessary.
> 
> Permissions can be per world so P2 needs to check that the player has WorkdEdit bypass permissions (or not) upon world change

Thanks a lot for clarifying. I'll add that back.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -533,7 +533,9 @@ public void onTeleport(PlayerTeleportEvent event) {
                     // to is identical to the plot's home location, and untrusted-visit is true
                     // i.e. untrusted-visit can override deny-teleport
                     // this is acceptable, because otherwise it wouldn't make sense to have both flags set
-                    if (!result && !(plot.getFlag(UntrustedVisitFlag.class) && plot
+                    if (result) {
+                        plotListener.plotEntry(pp, plot);
```
dordsor21, line 537: Why is this only being added here? Do we not want to call `plotEntry` if deny teleport doesn't allow entry, but the untrusted visit flag does?

dordsor21 (approved):

NotMyFault (approved):

PR #4515, opened 2024-10-09
Do not check interactions outside plot areas

### LINKED ISSUE #4513: Console Error
### Server Implementation

Paper

### Server Version

1.21.1

### Describe the bug

a Error in the Console

### To Reproduce

random
i dont know, how can reproduce the error

### Expected behaviour

without error

### Screenshots / Videos

_No response_

### Error log (if applicable)

https://paste.gg/p/anonymous/4bf73600bbdd43bcbaf58c95382e466d

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/5aab3da3a0bf4ddd8fa373c84fa00281

### PlotSquared Version

7.3.11 Premium

### Checklist

### Anything else?

nothing

SirYwell:
## Overview

Fixes #4513

## Description

I don't exactly know how to reproduce the exception, but it makes sense to be more defensive here, avoiding NPEs in situations the target location is not in a plot area.

```[tasklist]
### Submitter Checklist
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
+++ b/Core/src/main/java/com/plotsquared/core/util/EventDispatcher.java
@@ -371,7 +371,10 @@ public class EventDispatcher {
             Location location, BlockType blockType, boolean notifyPerms
     ) {
         PlotArea area = location.getPlotArea();
-        assert area != null;
+        // the interaction target location might be outside a plot area
+        if (area == null) {
+            return true;
+        }
         if (!area.buildRangeContainsY(location.getY()) && !player.hasPermission(Permission.PERMISSION_ADMIN_BUILD_HEIGHT_LIMIT)) {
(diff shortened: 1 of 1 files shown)


PierreSchwang:
I do kind of suspect the issue being that we use getPlotAreaAbs in the listener, whereas in the class where the area is null we use getPlotArea - but I can't really look into how they function differently rn

SirYwell:
I think the former is just on PlotPlayer rather than Location, but does the same actually.

PierreSchwang (approved):
Oh yeah, I see - in the listener the are of the player is checked (where the player is actually standing) whereas the location param of `checkPlayerBlockEvent` is the location of the block triggering the event. So those are most certainly always different - which is no issue in classical plot worlds ig

but the fix should make sense i guess

NotMyFault (approved):

PR #4522, opened 2024-10-13
Execute chunk loading operations outside the main thread

szumielxd:
## Description

Now, main command will wait with further execution until plot center location is asynchronously obtained. This will free the main thread while reaching plot in unloaded chunk

### Submitter Checklist


PierreSchwang (approved):
Can't see any obvious errors, and seems to be working ingame. My only nitpicks currently are
- Use a Logger instead of `Throwable#printStackTrace`
- I think we are using javax Annotations, instead of Jetbrains Annotations (not 100% sure on that though - another review is required either way)

dordsor21:
I would address the above things Pierre raised before merge - we should avoid merging code "smells"

szumielxd:
To be honest, It's hard to get any information about correct code-styling. I can easly find a file with both - NotNull and NonNull annotations. But I'm more than happy to modify my PR to use javax that's the correct one.

dordsor21:
Yeah it should switch to javax. Any issues already in the code can be fixed as and when but I'm opposed to introducing new  "problems" if we don't need to

szumielxd:
Now, everything should be fine.
Also, I'd like to suggest creating separate `contributing.md` file for this project to include project-specific guidelines. It's nearly 2 months since first comment about bad code practics that could be easly avoided

Core/src/main/java/com/plotsquared/core/command/MainCommand.java
```diff
@@ -147,8 +152,7 @@ public static MainCommand getInstance() {
                 try {
                     injector.getInstance(command);
                 } catch (final Exception e) {
-                    LOGGER.error("Failed to register command {}", command.getCanonicalName());
-                    e.printStackTrace();
+                    LOGGER.error(() -> "Failed to register command %s".formatted(command.getCanonicalName()), e);
```
dordsor21, line 155: Why are we changing away from the message formatting used by log4j?

dordsor21 (approved):

PR #4562, opened 2024-12-30
Release as java 21

NotMyFault:
As the title says, equally to WorldEdit.


SirYwell:
Does 1.19.4 work with Java 21?

PierreSchwang:
> The placeholders will not work if any of the plugin classes is compiled with Java 21, since it's not supported by the website for now

https://www.spigotmc.org/wiki/premium-resource-placeholders-identifiers/

PR #4569, opened 2025-01-11
feat: Add DisableBlockUpdateFlag

AerWyn81:
### Challenge
Using the current `disable-physics` flag doesn't suit all needs. 
I'm missing a flag that prevents blocks around one that I'm updating from being updated. 

There are several examples:
- I put a flower on a block, I break the block, I want the flower to stay in place,
- I modify a wall with the debug stick, I break a block, I don't want the wall to update,
- I modify a stair with the debug stick, I break the block, I don't want the stair to change direction.

### Resolution

In this PR, I've added the `disable-block-update` flag to disable block updating when blocks are updated. The default value is `false` because there shouldn't be any change in the update behaviour on a plot without a flag.

Two points:
- I haven't added the existing `disable-physics` flag because, as I see it, the two flags have their own responsibilities, one affecting the gravity of the blocks and the other the update.
- the flag doesn't block the updating of a block when snow is placed on it so that the block is covered in snow. I thought it should remain vanilla.

### Submitter Checklist


dordsor21:
I don't know if getting this specific with block physics should be in P2? I feel like the condition for not updating blocks is quite arbitrary, and doesn't really serve much of a purpose to the plot itself, and is rather something that should be player-specific based on that player's preferences

AerWyn81:
Hi @dordsor21, thanks for your response

I can understand your point of view in a context other than creative. 
This flag (which may or may not be activated) is of great interest in a creative context. However, what I don't understand is that PlotSquared already has a similar flag: disable-physics (related to gravity). Note that the proposal for the current flag does not contradict this element. After that, the principle of flag selection is closely linked to the choice of plot owner. It's a flag that was very much in demand in a creative context. At the moment, the flag is running well and is being used by players on a creative server with a snapshot version of PlotSquared, so the interest is there. It's useful for the plot because it blocks the updating of blocks that would otherwise remain static on builds

Core/src/main/resources/lang/messages_en.json
```diff
@@ -556,6 +556,7 @@
   "flags.flag_description_concrete_harden": "<gray>Set to `false` to disable concrete powder forming to concrete with water.</gray>",
   "flags.flag_description_device_interact": "<gray>Set to `true` to allow devices to be interacted with in the plot.</gray>",
   "flags.flag_description_disable_physics": "<gray>Set to `true` to disable block physics in the plot.</gray>",
+  "flags.flag_description_disable_block_update": "<gray>Set to `true` to disable block update in the plot.</gray>",
```
dordsor21, line 559:
```suggestion
  "flags.flag_description_disable_block_update": "<gray>Set to `true` to disable block updates in the plot, e.g. breaking of flowers when the block below is removed.</gray>",
```

Bukkit/src/main/java/com/plotsquared/bukkit/listener/HighFreqBlockEventListener.java
```diff
@@ -167,6 +170,18 @@ public void onPhysicsEvent(BlockPhysicsEvent event) {
             plot.debug("Prevented block physics and resent block change because disable-physics = true");
             return;
         }
+        if (plot.getFlag(DisableBlockUpdateFlag.class)) {
+            if (!event.getChangedType().hasGravity()) {
+                if (sourceBlock.getType().isAir() || block.getType().isAir()) {
+                    return;
+                }
+                if (!sourceBlock.getType().name().toLowerCase(Locale.ROOT).contains("snow")
+                        || !block.getLocation().add(0, -1, 0).getBlock().getType().equals(Material.GRASS_BLOCK)) {
```
dordsor21, line 179: I still think this is a massively specific implementation. It should be an "all or nothing" which is why I question its inclusion; this might be how _you_ want it, but it's not necessarily how others might want it to be implemented. This is why I suggest it might make better sense in a separate plugin where it is configurable

RedstoneFuture: For which practical creative construction use-case was this implemented? I would also say that “no block update” does not need any filter.

PR #4579, opened 2025-01-21
1.21.4 update

UndeadStudios:
nvm


dordsor21:
This breaks P2 support on all previous versions

UndeadStudios:
> This breaks P2 support on all previous versions

true didnt think about that just wanted 1.21.4 support

PierreSchwang:
Using / comparing the id or namespaced key should work ig. Should be put in constants though

PierreSchwang:
#4582

PR #4582, opened 2025-01-25
1.21.4

PierreSchwang:
## Overview
Not tested yet, checked IDs against minecraft wiki and seems to work out.

```[tasklist]
### Submitter Checklist
```


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -376,7 +377,8 @@ public void onEffect(@NonNull EntityPotionEffectEvent event) {
 
     @EventHandler
     public void onVehicleEntityCollision(VehicleEntityCollisionEvent e) {
-        if (e.getVehicle().getType() == EntityType.BOAT) {
+        final Class<? extends Entity> clazz = e.getVehicle().getType().getEntityClass();
+        if (clazz != null && (Boat.class.isAssignableFrom(clazz) || ChestBoat.class.isAssignableFrom(clazz))) {
```
PierreSchwang, line 381: always get the `isAssignableFrom` order confused - `clazz` is basically something like `OakBoat`, which implements `Boat` etc

NotMyFault (approved):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -376,7 +376,7 @@ public void onEffect(@NonNull EntityPotionEffectEvent event) {
 
     @EventHandler
     public void onVehicleEntityCollision(VehicleEntityCollisionEvent e) {
-        if (e.getVehicle().getType() == EntityType.BOAT) {
+        if (e.getEntity() instanceof Boat) {
```
SirYwell, line 379: That looks odd, I assume this should still be `e.getVehicle()`?

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 # Platform expectations
-paper = "1.20.4-R0.1-SNAPSHOT"
+paper = "1.21.4-R0.1-SNAPSHOT"
```
SirYwell, line 3: Will that cause binary incompatibilities with older versions?

SirYwell (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
PierreSchwang: yup, mb

gradle/libs.versions.toml
PierreSchwang: no idea, could just go back to 1.20 there (was primarily required to see the errors / changes)

dordsor21:
Do all these new clases/fields exist in 1.19.4? We still technically target it for support in P2 I believe?

SirYwell (approved):

dordsor21 (approved):

PierreSchwang:
I think I should look at the PR again, as there are seemingly string constants for entity types in the code, which obviously weren't caught by the newer paper api

PierreSchwang:
I'm not gonna touch the ReplicatingEntityWrapper (required for plot clear as it seems?). That would need a whole refactor for maaany things (and potentially a migration to just storing the entities as NBT - but that's something that has to be discussed).

PierreSchwang:
> > I'm not gonna touch the ReplicatingEntityWrapper (required for plot clear as it seems?). That would need a whole refactor for maaany things (and potentially a migration to just storing the entities as NBT - but that's something that has to be discussed).
> 
> Does it need a refactor though? What needs refactoring?

Well, without looking further, the first things I noticed:
- many entity types are not supported / added
- only `BOAT` is currently added (not the individual ones, like OAK_BOAT etc), setting the dataByte to the ordinal value of getWoodType of TreeVariant (which does not support all supported wood types)
- Chest-Boats are not supported at all (no inventory, no dataByte)
- Rafts are missing

I guess I could add the named ones, but it would be really great to get to java 21 to use type pattern matching in the switch-cases (though, blocked by SpigotMC atm)

dordsor21:
> > > I'm not gonna touch the ReplicatingEntityWrapper (required for plot clear as it seems?). That would need a whole refactor for maaany things (and potentially a migration to just storing the entities as NBT - but that's something that has to be discussed).
> > 
> > 
> > Does it need a refactor though? What needs refactoring?
> 
> Well, without looking further, the first things I noticed:
> 
> * many entity types are not supported / added
> * only `BOAT` is currently added (not the individual ones, like OAK_BOAT etc), setting the dataByte to the ordinal value of getWoodType of TreeVariant (which does not support all supported wood types)
> * Chest-Boats are not supported at all (no inventory, no dataByte)
> * Rafts are missing
> 
> I guess I could add the named ones, but it would be really great to get to java 21 to use type pattern matching in the switch-cases (though, blocked by SpigotMC atm)

Perhaps just minimal work for 1.21.4 and we can revisit this later? I don't see much issue with using strings - multi-version support is tricky otherwise.

dordsor21:
Looks good enough for now to me

PR #4593, opened 2025-02-13
fix: plot entry trigger on teleport

Tjorven-Liebe:
## Overview
The teleportevent will not longer trigger the plot-entry method in PlotListener, if currentplot = lastplot.

## Description
This will change the way the entry is handled on teleport. If the current plot is the last plot, the teleportevent will not longer trigger the plot-entry method in PlotListener.


dordsor21 (approved):

RedstoneFuture:
Yes, good point. Thank you.
Currently, the `greeting` flag is really executed with every teleport event (e.g. WorldEdit compass teleport), which is not useful.

SirYwell:
Does this also handle merged plots?

dordsor21:
> Does this also handle merged plots?

Probably not, should be a `getBasePlot` I'd say

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -593,7 +593,7 @@ public void onTeleport(PlayerTeleportEvent event) {
                     return;
                 }
                 Plot plot = area.getPlot(location);
-                if (plot != null) {
+                if (plot != null && !plot.equals(lastPlot)) {
```
dordsor21, line 596: Should compare base plots here

dordsor21 (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
Tjorven-Liebe: https://github.com/user-attachments/assets/180c2284-c97a-4fba-8b45-9f8aeea21e38

Tjorven-Liebe:
> > Does this also handle merged plots?
> 
> Probably not, should be a `getBasePlot` I'd say

isn't base plot the main plot of a merge. I've tested my code, by me it works totally fine.

https://github.com/user-attachments/assets/180c2284-c97a-4fba-8b45-9f8aeea21e38

dordsor21 (approved):
the PlotArea#getPlot calls getBasePlot itself 👍

dordsor21 (approved):

PR #4600, opened 2025-02-22
feat/chore: optimize plot queries and visit command

PierreSchwang:
## Description
- Allow streaming of plots in PlotQuery to early return / short-circuiting where applicable
- Don't duplicate queries in `/visit` to remove overhead

```[tasklist]
### Submitter Checklist
- [ ] Make sure you are opening from a topic branch (**/feature/fix/docs/ branch** (right side)) and not your main branch.
- [ ] I read and followed the [contribution guidelines](https://github.com/IntellectualSites/.github/blob/main/CONTRIBUTING.md).
```

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/util/query/OwnersIncludeFilter.java
+++ b/Core/src/main/java/com/plotsquared/core/util/query/OwnersIncludeFilter.java
@@ -23,17 +23,11 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 
 import java.util.UUID;
 
-class OwnersIncludeFilter implements PlotFilter {
-
-    private final UUID owner;
-
-    OwnersIncludeFilter(final @NonNull UUID owner) {
-        this.owner = owner;
-    }
+record OwnersIncludeFilter(UUID owner) implements PlotFilter {
 
     @Override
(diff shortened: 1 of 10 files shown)


Core/src/main/java/com/plotsquared/core/util/query/OwnersIncludeFilter.java
```diff
@@ -23,17 +23,11 @@
...
-
-    OwnersIncludeFilter(final @NonNull UUID owner) {
-        this.owner = owner;
-    }
+record OwnersIncludeFilter(UUID owner) implements PlotFilter {
 
     @Override
     public boolean accepts(final @NonNull Plot plot) {
-        return plot.isBasePlot() && plot.getOwners().size() > 0 && plot.getOwners().contains(owner);
+        return plot.isBasePlot() && plot.isOwner(owner);
```
PierreSchwang, line 30: the current `Plot#getOwners#contains` takes server-plots into account (so, if the server uuid is passed into this filter, and a plot is server owned, this filter would return true). `Plot#isOwner` uses `Plot#getOwnerAbs` and therefor would not produce the same result. Haven't seen an occurrence of a server plot being queried - not sure if this should just be changed either tbh

PR #4654, opened 2025-05-19
Add entity cap check of armor stands & end crystals into player interact listener

FlorianMichael:
## Overview
When cancelled in the creature spawn event and placed by a player, the player will lose the armor stand / end crystal item from their inventory. This PR aims to fix this by moving the verification whether the entity should be cancelled up to the spawn egg right click, this way the item is not touched.

I tried my best to verify this is not breaking anything; I'm not sure if the original code in EntitySpawnListener should be kept? Reviews appreciated.

### Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntitySpawnListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/EntitySpawnListener.java
@@ -18,6 +18,7 @@
  */
 package com.plotsquared.bukkit.listener;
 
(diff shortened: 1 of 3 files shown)


SirYwell:
I assume moving this check isn't okay. Spawning should also be prevented when caused by e.g. dispensers, throwing eggs, etc.

FlorianMichael:
@SirYwell Hm yeah, I forgot that. Added the original code back now, this way it's cancelled on interaction level when the player spawns the entity and otherwise on creature spawn level.

dordsor21 (approved):

SirYwell (approved):

PR #4683, opened 2025-06-28
fix: `/plot grant add` doesn't send success message reliably

### LINKED ISSUE #4383: `/plot grant add <name>` doesn't send success message
https://github.com/IntellectualSites/PlotSquared/blob/ff83868cbc20f082362bfcb487b62c865a2926e9/Core/src/main/java/com/plotsquared/core/command/Grant.java#L133-L135
sends the success message, but commonly, you don't reach the `else` block, unless `pp == null`, which isn't the case, normally.
We could move the message to the line below
https://github.com/IntellectualSites/PlotSquared/blob/ff83868cbc20f082362bfcb487b62c865a2926e9/Core/src/main/java/com/plotsquared/core/command/Grant.java#L103
where the actual grant is added to the player.

Lion-King287:
## Overview

Fixes #4383 

## Description
Moves the `grants.added` success message in the `Grant.java` command handler so that it is reliably sent **after** the plot has been successfully granted.

### Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Grant.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Grant.java
@@ -101,6 +101,10 @@ public class Grant extends Command {
                                     );
                                 } else {
                                     access.set(access.get().orElse(0) + 1);
+                                    player.sendMessage(
+                                            TranslatableCaption.of("grants.added"),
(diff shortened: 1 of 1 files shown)


Lion-King287:
> I feel like the message should be duplicated, not moved - no? As of now, the message would be sent if the target player is online, but not anymore if they are offline. The confirmation should be sent to the command sender in either case.

I really didn't think of that, it's fixed now. Thank you!

PR #4698, opened 2025-07-18
fix: correct tab completion for `/plot flag set <flag> <value>`

### LINKED ISSUE #3028: `/plot f set <flag> <value>` uses improper tab completions
The tab completion should stop at the value provided

jvstinxz:
## Overview
Fixes #3028 

## Description
This PR fixes the incorrect tab completion behavior for the command `/plot flag set <flag> <value>`

### Submitter Checklist

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
+++ b/Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
@@ -316,7 +316,7 @@ public final class FlagCommand extends Command {
             } catch (final Exception ignored) {
             }
         }
-        return tabOf(player, args, space);
+        return Collections.emptyList();
     }
 
     @CommandDeclaration(command = "set",


PierreSchwang (approved):
works

PierreSchwang:
the tabcompletion should probably get a small refactor - if you want to, you can adjust those small things:
- don't call Arrays.asList for every tab completion (could be a Set which is initialized once - or compare constants directly)
- potentially check args.length before comparing against the list (not sure if the JVM optimizes that either way)
- the abbreviations don't seem to tab-complete (e.g. `/p f set` -> `/p f s`) - all possible abbreviations are listed on line 252

jvstinxz:
Still new to this so I just did the third suggestion, should be fine like that.

Core/src/main/java/com/plotsquared/core/command/FlagCommand.java
```diff
@@ -270,14 +270,14 @@ public Collection<Command> tab(
                     .filter(value -> value.startsWith(args[0].toLowerCase(Locale.ENGLISH)))
                     .map(value -> new Command(null, false, value, "", RequiredType.NONE, null) {
                     }).collect(Collectors.toList());
-        } else if (Arrays.asList("set", "add", "remove", "delete", "info")
+        } else if (Arrays.asList("set", "s", "add", "a", "remove", "r", "delete", "info", "i")
```
dordsor21, line 273: Yeah I'd like to see these extracted to static final `Collections.immutableSet` at the class level - this would be best practices for Java

PR #4728, opened 2025-08-10
fix: special handle thrown eggs for projectile hit checks

FlorianMichael:
## Overview
Throwing eggs on other plots will spawn chickens even though the player is not trusted on that plot.

## Description
Cancelling the ProjectileHitEvent will not cover for thrown eggs spawning chickens due to how Bukkit has these events structured. The provided patch calls the same code used for normal projectiles in the specialized egg event to prevent unwanted chickens from spawning on other plots.

### Submitter Checklist


SirYwell (approved):
Looks good, thanks.

dordsor21:
Is this the event thrown on the egg throw or egg hit? I.e. does this block a player throwing it across a road for example

FlorianMichael:
Also: It would be difficult to prevent the player from actually throwing the projectile, as you would have to calculate the trajectory of the projectile to determine whether it will land on a plot for which the thrower has permissions on.

PR #4736, opened 2025-08-23
fix: correct height that FAWE ends up using for schematics

### LINKED ISSUE #4404: Freshly create plot schematic is regenerated (cleared) one block too low
### Server Implementation

Paper

### Server Version

1.20.4

### Describe the bug

When clearing a plot in an area utilizing plot schematics, the schematic will be pasted one block too low.
World-Generation itself works (e.g. //regen works).

The culprit seems to be somewhere here (used by the FAWE Delegate to determine minY for plot schematics):
https://github.com/IntellectualSites/PlotSquared/blob/a6412581a6bef54b816c078c690e6815495377dd/Core/src/main/java/com/plotsquared/core/generator/HybridPlotWorld.java#L558-L560

When replacing `SCHEM_Y + plotY` with `getMinBuildHeight()` (the same reference point the world generator uses, as visible in HybridGen: https://github.com/IntellectualSites/PlotSquared/blob/main/Core/src/main/java/com/plotsquared/core/generator/HybridGen.java#L76) clear works. My only idea for that being is, that one of these two variables is normalized somewhere - possibly against the road height.
I don't just want to PR my fix, as I'm not sure what the reasoning behind the change was (@dordsor21 ?) and if that does not have any unexpected side effects.

### To Reproduce

1. Create plot schematic as described in wiki (`/plot schematic save`, move to appropriate schematic folder)
2. World will be generated accordingly
3. Claim Plot and run `/plot clear`
4. Schematic will be one block too low

### Expected behaviour

The schematic should be pasted from the correct y-level

### Screenshots / Videos

_No response_

### Error log (if applicable)

_No response_

### Fawe Debugpaste

-

### Fawe Version

-

### Checklist

- [X] I have included a Fawe debugpaste.
- [X] I am using the newest build from https://ci.athion.net/job/FastAsyncWorldEdit/ and the issue still persists.

### Anything else?

_No response_

dordsor21:
- this did not end up having an impact on P2 code as it always used getMinBuildHeight when PASTE_ON_TOP was false anyway
 - there should not be any side effects to this as any uses this specific SCHEM_Y value are effectively normalised against it
 - fixes #4404

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/generator/ClassicPlotWorld.java
+++ b/Core/src/main/java/com/plotsquared/core/generator/ClassicPlotWorld.java
@@ -164,6 +164,9 @@ public abstract class ClassicPlotWorld extends SquarePlotWorld {
                 : getMinBuildHeight();
     }
 
+    /**
+     * Get the lowest height of plot, road, and wall. Accounts for {@link Settings.Schematics#USE_WALL_IN_ROAD_SCHEM_HEIGHT}
+     */
     int schematicStartHeight() {
         int plotRoadMin = Math.min(PLOT_HEIGHT, ROAD_HEIGHT);
         if (!Settings.Schematics.USE_WALL_IN_ROAD_SCHEM_HEIGHT) {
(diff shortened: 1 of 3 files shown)


TheMeinerLP:
Reminder: Apply patch file
[fix__update_minimum_Y_coordinate_calculation_to_use_getMinGenHeight.patch](https://github.com/user-attachments/files/22186589/fix__update_minimum_Y_coordinate_calculation_to_use_getMinGenHeight.patch)

NotMyFault (approved):
Untested, but looks reasonable

NotMyFault (approved):

PR #4747, opened 2025-08-31
Remove redundant permission checks for Flag List Command

gamecrash534:
## Overview

This PR relates to an issue where a player cannot list all plot flags available if he isn't on a plot where he has permissions to change them.

## Description

As said in the overview, this PR aims to solve the issue of a player not being able to use `/plot flag list` if he does not meet the requirement of standing on a plot where he has permission to change these flags.  The idea behind this issue being that a player should be able to view the list of flags regardless of permission to change them - as one might be on a plot of another player and wants to find the correct flag to do XY, e.g. to help him out. This eliminates the need to visit / google the documentation or go to a own plot just to enter this command and view the available flags. It still, though, needs the `plots.flag.list`-permission to be executed.

### Submitter Checklist


SirYwell:
From my understanding, the `plots.flag.list` permission allows to use the command on *all* plots, while without the permission, using it isn't even allowed on the own plot, right?
In that case, I think we should rather add another permission (e.g. `plots.flag.list.other`). Without that permission, `checkRequirements` should still be called. This way, it is possible to control whether the user can run the command on other plots too.

dordsor21 (approved):

NotMyFault (approved):

PR #4758, opened 2025-09-17
fix: binary incompatibility on 1.21.1

PierreSchwang:
## Overview
EssentialsX pulled in paper as a dependency , which resulted in paper 1.21.8 being on the classpath. Gradle / Java compiled against that version, which resulted in PlotSquared not running on older versions anymore (like 1.21.1, etc). Reason being the change of the Biome enum to an interface, which results in different instructions in the bytecode, which makes PS not compatible with the enum versions anymore.

Strictly set the paper version globally and drop bukkit and spigot globally. That should prevent that happening in the future as well.

### Submitter Checklist

### DIFF
--- a/Bukkit/build.gradle.kts
+++ b/Bukkit/build.gradle.kts
@@ -17,6 +17,19 @@ repositories {
     }
 }
 
+// Make sure we control the exact version of paper being included, while dropping spigot + bukkit
+configurations.all {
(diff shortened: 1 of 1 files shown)


SirYwell (approved):
Looks good.

NotMyFault (approved):
Untested, but looks reasonable.

PR #4801, opened 2025-12-02
Remove BlockDamageEvent

larroxtv:


PierreSchwang:
So the instabreak flag still functions? Players still get a message, that they can't break blocks without having to wait for the block break animation to finish (e.g. with obsidian)?

I doubt simply removing the event (or rather the listener) is the correct approach to that.

PR #4818, opened 2026-01-22
Make sure no blocks are placed when deleting a plot

### LINKED ISSUE #4794: Blocks placed during /plot delete confirmation remain after plot is deleted
### Server Implementation

Paper

### Server Version

1.21.3

### Describe the bug

When deleting a plot, any blocks that are placed during the `/p confirm` phase remain on the plot even after it is deleted or becomes unclaimed.
These blocks are not removed together with the plot deletion process.

### To Reproduce

1. Initiate a plot deletion (`/plot delete`).
2. While waiting for the confirmation, place several blocks on the plot.
3. Execute `/plot confirm`.
4. The plot is deleted / unclaimed, but the blocks placed during the confirmation phase remain instead of being removed.

### Expected behaviour

All blocks placed on the plot (including those placed during the deletion confirmation phase) should be removed when the plot is deleted or unclaimed.

### Screenshots / Videos

https://github.com/user-attachments/assets/596986e1-f635-4827-aa56-acce8f986757

### Error log (if applicable)

_No response_

### Plot Debugpaste

https://athion.net/ISPaster/paste/view/04b06c6d51d3420bab74485da6499af0

### PlotSquared Version

PlotSquared version 7.5.10-SNAPSHOT

### Checklist

- [x] I have included a Plot debugpaste.
- [x] I am using the newest build from https://www.spigotmc.org/resources/77506/ and the issue still persists.

### Anything else?

_No response_

florianreuth:
## Overview
Fixes https://github.com/IntellectualSites/PlotSquared/issues/4794

## Description
When starting the deletion process, a plot meta is set. The plot meta will be unset once the plot deletion is finished and during it being set block place/break events will be cancelled.

### Submitter Checklist


Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -135,6 +144,12 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
         };
         if (hasConfirmation(player)) {
             CmdConfirm.addPending(player, getCommandString() + ' ' + plot.getId(), run);
+            // Schedule cleanup task for when confirmation times out
+            TaskManager.runTaskLater(() -> {
+                for (Plot connectedPlot : plots) {
+                    connectedPlot.deleteMeta("pendingDelete");
+                }
+            }, TaskTime.seconds(Settings.Confirmation.CONFIRMATION_TIMEOUT_SECONDS + 1));
```
dordsor21: This is potentially a problem in case the delete takes a long time for some reason. Probably need to refactor a bit to cancel this task upon confirmation.

Also probably need to ensure that the meta is removed on error during delete as well

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -92,8 +93,14 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
         final int currentPlots = Settings.Limit.GLOBAL ?
                 player.getPlotCount() :
                 player.getPlotCount(plot.getWorldName());
+
+        final AtomicBoolean confirmed = new AtomicBoolean(false);
```
dordsor21: No need for `final` modifiers - effective final is good enough

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -139,15 +146,23 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
...
             } else {
+                for (Plot connectedPlot : plots) {
+                    connectedPlot.deleteMeta("pendingDelete");
+                }
                 player.sendMessage(TranslatableCaption.of("errors.wait_for_timer"));
             }
         };
         if (hasConfirmation(player)) {
+            for (Plot connectedPlot : plots) {
+                connectedPlot.setMeta("pendingDelete", true);
```
dordsor21: No need to stop players doing anything between `/plot delete` and confirmation of deletion

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -92,8 +93,14 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
         final int currentPlots = Settings.Limit.GLOBAL ?
                 player.getPlotCount() :
                 player.getPlotCount(plot.getWorldName());
+
+        final AtomicBoolean confirmed = new AtomicBoolean(false);
         Runnable run = () -> {
+            confirmed.set(true);
             if (plot.getRunning() > 0) {
+                for (Plot connectedPlot : plots) {
+                    connectedPlot.deleteMeta("pendingDelete");
```
dordsor21: We should also be ensuring that the meta is removed if an exception is thrown somewhere (try-finally).

dordsor21 (changes requested):

Core/src/main/java/com/plotsquared/core/command/Delete.java
dordsor21: a try inside the callback itself is fine. Exceptions obviously shouldn't happen but that doesn't mean we shouldn't handle it gracefully. Plot meta is persistent so on a restart the p[lot still still be uneditable. To ensure we're not removing meta twice, etc. we can always add a boolean somewhere that's set when the meta is removed (and rather than a finally just except and rethrow. It's a little ugly but it's important we don't lock a plot out :) )

florianreuth: Okay, fixed now

NotMyFault:
@dordsor21 might taking a look?

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -91,11 +91,19 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
...
+                }
                 player.sendMessage(TranslatableCaption.of("errors.wait_for_timer"));
                 return;
             }
+
+            for (Plot connectedPlot : plots) {
+                connectedPlot.setMeta("pendingDelete", true);
+            }
```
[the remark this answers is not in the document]
dordsor21: That's fine

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -91,11 +91,19 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
         final int currentPlots = Settings.Limit.GLOBAL ?
                 player.getPlotCount() :
                 player.getPlotCount(plot.getWorldName());
+
         Runnable run = () -> {
             if (plot.getRunning() > 0) {
+                for (Plot connectedPlot : plots) {
+                    connectedPlot.deleteMeta("pendingDelete");
+                }
```
dordsor21, line 99: theoretically a player could run delete twice and remove the meta in this case

Core/src/main/java/com/plotsquared/core/command/Delete.java
```diff
@@ -104,32 +112,47 @@ public boolean onCommand(final PlotPlayer<?> player, String[] args) {
...
+                            ),
+                            TagResolver.resolver("world", Tag.inserting(Component.text(plotArea.getWorldName()))),
+                            TagResolver.resolver("plot", Tag.inserting(Component.text(plot.getId().toString())))
+                    );
+                    eventDispatcher.callPostDelete(plot);
+                } catch (Throwable e) {
+                    // ... or if something went wrong
+                    for (Plot connectedPlot : plots) {
+                        connectedPlot.deleteMeta("pendingDelete");
+                    }
```
dordsor21, line 146: We remove the meta at the start of the try block, so this is needless. Error handling I meant in the deletion method itself - we need to make sure we're recovering properly (if this requires architectural changes/changes the method signatures so be it)

PR #4901, opened 2026-07-15
fix: prevent NullPointerException in world setup when getWorld returns null

bastigfz:
### Description

Fixes a `NullPointerException` that occurs during `/plot setup` when `BukkitUtil.getSpawn()` is called for a world that has not yet been registered with Bukkit.

### Root Cause

`BukkitUtil.getSpawn()` called `getWorld(world).getSpawnLocation()` without a null check. During the plot setup process, the world may not yet be available via `Bukkit.getWorld()`, causing:

```
java.lang.NullPointerException: Cannot invoke "org.bukkit.World.getSpawnLocation()"
because the return value of "com.plotsquared.bukkit.util.BukkitUtil.getWorld(String)" is null
  at BukkitUtil.getSpawn(BukkitUtil.java:318)
  at CommonSetupSteps.java:231
```

### Fix

Added a null check in `BukkitUtil.getSpawn()`, returning a safe fallback spawn location at `(0, 64, 0)` when the world is not yet available. This is consistent with the existing null-safe pattern already used in `setSpawn()`.

### Related

- Reported in https://gist.github.com/byJoonas/98c4d355da6ee885bbd65d3a25e42ae0

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/player/BukkitPlayer.java
@@ -41,6 +41,7 @@ import com.sk89q.worldedit.world.item.ItemTypes;
 import io.papermc.lib.PaperLib;
 import net.kyori.adventure.audience.Audience;
 import org.bukkit.GameMode;
+import org.bukkit.World;
 import org.bukkit.Sound;
 import org.bukkit.SoundCategory;
 import org.bukkit.WeatherType;
@@ -229,8 +230,12 @@ public class BukkitPlayer extends PlotPlayer<Player> {
         if (!WorldUtil.isValidLocation(location)) {
             return;
         }
+        final World world = BukkitUtil.getWorld(location.getWorldName());
+        if (world == null) {
+            return;
+        }
         final org.bukkit.Location bukkitLocation =
-                new org.bukkit.Location(BukkitUtil.getWorld(location.getWorldName()), location.getX() + 0.5,
+                new org.bukkit.Location(world, location.getX() + 0.5,
                         location.getY(), location.getZ() + 0.5, location.getYaw(), location.getPitch()
                 );
         PaperLib.teleportAsync(player, bukkitLocation, getTeleportCause(cause));
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitUtil.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/util/BukkitUtil.java
@@ -26,6 +26,7 @@ import com.plotsquared.core.PlotSquared;
 import com.plotsquared.core.configuration.caption.Caption;
 import com.plotsquared.core.configuration.caption.LocaleHolder;
 import com.plotsquared.core.location.Location;
+import com.plotsquared.core.location.UncheckedWorldLocation;
 import com.plotsquared.core.player.PlotPlayer;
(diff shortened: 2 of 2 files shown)


SirYwell:
> During the plot setup process, the world may not yet be available via `Bukkit.getWorld()`

Why is that?

MattBDev:
I've never encountered this. I'm going to look into this before approving.

bastigfz:
> Why is that? 

@SirYwell 

The issue occurs because `BukkitSetupUtils.setupWorld()` calls `worldManager().handleWorldCreation(...)` which may create the world asynchronously (e.g. via Multiverse or Paper world creation). After that call, `Bukkit.getWorld(world)` can still be `null`, but `setupWorld()` returns the world name anyway:

```java
Objects.requireNonNull(PlotSquared.platform()).worldManager()
        .handleWorldCreation(builder.worldName(), builder.generatorName());

if (Bukkit.getWorld(world) != null) {
    return world;
}

return builder.worldName();
```

Back in `CHOOSE_WORLD_NAME`, `getSpawn(world)` is called immediately after `setupWorld()` returns, and at that point the world may not yet be registered with Bukkit, causing the NPE.

SirYwell:
> The issue occurs because `BukkitSetupUtils.setupWorld()` calls `worldManager().handleWorldCreation(...)` which may create the world asynchronously (e.g. via Multiverse or Paper world creation).

No, world creation is not asynchronous. Maybe your fork is at fault.

bastigfz:
> No, world creation is not asynchronous. Maybe your fork is at fault.

You are right, `Bukkit.createWorld()` is synchronous. I was wrong about async world creation.

Looking at the stack trace from the reporter again:
```
[16:52:05] NPE at BukkitUtil.getSpawn()
[16:52:05] Creating world plotworld...
[16:52:06] World plotworld created!
```

The NPE happens *before* the world is even created. This means `setupWorld()` returned the world name without `handleWorldCreation()` having been called, or the world creation failed silently.

I cannot fully explain the root cause from the code alone – it could be a Leaf-specific behavior, a version mismatch in 7.5.13-Premium, or something else. However, the defensive null-checks in `getSpawn()` and `BukkitPlayer.teleport()` are still valid: they prevent the crash regardless of *why* the world is not registered, and they follow the same null-safe pattern already used in `setSpawn()`.

SirYwell:
> > No, world creation is not asynchronous. Maybe your fork is at fault.
> 
> You are right, `Bukkit.createWorld()` is synchronous. I was wrong about async world creation.
> 
> Looking at the stack trace from the reporter again:
> 
> ```
> [16:52:05] NPE at BukkitUtil.getSpawn()
> [16:52:05] Creating world plotworld...
> [16:52:06] World plotworld created!
> ```
> 
> The NPE happens _before_ the world is even created. This means `setupWorld()` returned the world name without `handleWorldCreation()` having been called, or the world creation failed silently.
> 
> I cannot fully explain the root cause from the code alone – it could be a Leaf-specific behavior, a version mismatch in 7.5.13-Premium, or something else. However, the defensive null-checks in `getSpawn()` and `BukkitPlayer.teleport()` are still valid: they prevent the crash regardless of _why_ the world is not registered, and they follow the same null-safe pattern already used in `setSpawn()`.

This really sounds like a problem with the server software you are using.

NonSwag:
`#createWorld` will only ever return null if another world with the exact same UUID already exists, everything else throws an exception.
If the world did not exist before (which is generally the case for the setup command) `#createWorld`will never return null

PR #4910, opened 2026-07-29
Fixed support for new egg types in 1.21.5 and later versions

MattBDev:
## Overview
Fixes an issue where blue and brown eggs aren't treated the same as regular eggs. 

## Description
Includes a new variable to detect all eggs in game and treat them the same. 

### Submitter Checklist

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -92,6 +92,7 @@ import net.kyori.adventure.text.minimessage.tag.standard.StandardTags;
 import org.bukkit.Bukkit;
 import org.bukkit.FluidCollisionMode;
 import org.bukkit.Material;
+import org.bukkit.NamespacedKey;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.Sign;
@@ -1310,16 +1311,20 @@ public class PlayerEventListener implements Listener {
                         }
                     }
                 }
+                // 1.21.5 (Spring to Life) introduced brown_egg and blue_egg.
+                // This new variable detects all current and future eggs.
+                // It could be replaced with an Item Tag or the exact Material in the future.
+                boolean isEgg = Optional.ofNullable(Bukkit.getTag(org.bukkit.Tag.REGISTRY_ITEMS, NamespacedKey.minecraft("eggs"), Material.class))
+                        .map(tag -> tag.isTagged(type))
+                        .orElse(type.name().endsWith("EGG"));
                 if (PaperLib.isPaper()) {
-                    if (MaterialTags.SPAWN_EGGS.isTagged(type) || Material.EGG.equals(type)) {
+                    if (MaterialTags.SPAWN_EGGS.isTagged(type) || isEgg) {
                         eventType = PlayerBlockEventType.SPAWN_MOB;
                         break;
                     }
-                } else {
-                    if (type.toString().toLowerCase().endsWith("egg")) {
-                        eventType = PlayerBlockEventType.SPAWN_MOB;
-                        break;
-                    }
+                } else if (isEgg) {
+                    eventType = PlayerBlockEventType.SPAWN_MOB;
+                    break;
                 }
                 if (type.isEdible()) {
                     //Allow all players to eat while also allowing the block place event to be fired
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
@@ -92,6 +92,7 @@ import net.kyori.adventure.text.minimessage.tag.standard.StandardTags;
 import org.bukkit.Bukkit;
 import org.bukkit.FluidCollisionMode;
 import org.bukkit.Material;
+import org.bukkit.NamespacedKey;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.Sign;
@@ -155,6 +156,7 @@ import org.bukkit.plugin.Plugin;
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.util.Vector;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.lang.reflect.Field;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1310,16 +1310,18 @@ public void onInteract(PlayerInteractEvent event) {
...
+                // It could be replaced with an Item Tag or the exact Material in the future.
+                final boolean isEgg = type.toString().toLowerCase().endsWith("egg");
                 if (PaperLib.isPaper()) {
-                    if (MaterialTags.SPAWN_EGGS.isTagged(type) || Material.EGG.equals(type)) {
-                        eventType = PlayerBlockEventType.SPAWN_MOB;
-                        break;
-                    }
-                } else {
-                    if (type.toString().toLowerCase().endsWith("egg")) {
+                    if (MaterialTags.SPAWN_EGGS.isTagged(type) || isEgg) {
```
SirYwell, line 1327: Why do we need the `||` here?

MattBDev: Spawn eggs are not the same as regular eggs so the `||` here is valid and needed. However, I should have probably reversed this check so `isEgg` is checked BEFORE the `MaterialTag` since it will probably be true in almost all cases. I'm willing to change it for PR approval, but it honestly won't make much of a difference.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1310,16 +1310,18 @@ public void onInteract(PlayerInteractEvent event) {
                         }
                     }
                 }
+                // 1.21.5 (Spring to Life) introduced brown_egg and blue_egg.
+                // This new variable detects all current and future eggs.
+                // It could be replaced with an Item Tag or the exact Material in the future.
```
PierreSchwang, line 1322: Not wanting to be nit-picky, but why not now already, given the `eggs` tag was introduced in 1.21.5? Something like
```java
boolean isEgg = Optional.ofNullable(Bukkit.getTag(org.bukkit.Tag.REGISTRY_ITEMS, NamespacedKey.minecraft("eggs"), Material.class))
                        .map(tag -> tag.isTagged(type))
                        .orElseGet(() -> type.name().endsWith("EGG"));
```
with some caching?

MattBDev: You're not being nitpicky. I've been gone for awhile so I appreciate the constructive feedback. That is actually a more elegant solution and actually almost what I did. I guess I went the lazy way? 😅

PierreSchwang: A simple `Material.EGG.equals(type)` should suffice in the orElse part, wasn't thinking much

MattBDev: Actually, it would just be `.orElse(type.name().endsWIth("EGG")`

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1310,16 +1311,20 @@ public void onInteract(PlayerInteractEvent event) {
...
+                boolean isEgg = Optional.ofNullable(Bukkit.getTag(org.bukkit.Tag.REGISTRY_ITEMS, NamespacedKey.minecraft("eggs"), Material.class))
```
PierreSchwang: I *think* based on the JavaDocs of either `Bukkit.getTag` or something along the lines the result of said method should be cached (seems to be an expensive method, potentially).

PierreSchwang (changes requested):

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
MattBDev: Would you prefer to make the call to `Bukkit.getTag()` a static field when the PlayerEventListener class is initialized?

PierreSchwang: Yeah, that would be the best approach I'd say

NotMyFault:
@Copilot fix the merge conflict

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -155,6 +156,7 @@
 import org.bukkit.potion.PotionEffect;
 import org.bukkit.util.Vector;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.jetbrains.annotations.Nullable;
```
dordsor21, line 159: wrong annotation type

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -291,6 +293,11 @@ public class PlayerEventListener implements Listener {
     private final WorldEdit worldEdit;
     private final PlotAreaManager plotAreaManager;
     private final PlotListener plotListener;
+    private final org.bukkit.@Nullable Tag<Material> eggTag = Bukkit.getTag(
```
dordsor21, line 296: bukkit.nullable?!

Bukkit/src/main/java/com/plotsquared/bukkit/listener/PlayerEventListener.java
```diff
@@ -1310,16 +1317,20 @@ public void onInteract(PlayerInteractEvent event) {
                         }
                     }
                 }
+                // 1.21.5 (Spring to Life) introduced brown_egg and blue_egg.
+                // This new variable detects all current and future eggs.
+                // It could be replaced with an Item Tag or the exact Material in the future.
+                boolean isEgg = Optional.ofNullable(eggTag)
+                        .map(tag -> tag.isTagged(type))
+                        .orElse(type.name().endsWith("EGG"));
```
dordsor21, line 1325: does this not incorrectly match dragon_egg?

PierreSchwang: Should be simply replaced with Material.EGG.equals / ==

PR #4923, opened 2026-08-17
Add runtime deletion for single plot areas

### LINKED ISSUE #4603: remove single plot areas easily

You really should add a feature where we can delete single plot areas without having to restart the server. You’ve done it so we can make single plot areas but now you really should add that in at some point.

No

### Anything else?

PlotSquared is a great plugin, I knew what I was purchasing. But I feel like it should be easier for people with learning difficulties to make use of this plugin. Maybe try and make it so easy plot has a unique ID instead of it all being “1;1”

in a server like mine that is based around players buying plots for money, I wanted plot squared so I can manage the plots easier and maybe even automate them since I know this plugin is capable of that.

Please take my suggestion into consideration. Thank you

tantaf123:
## Overview

Fixes #4603

## Description

Adds `/plot area delete <area>` to safely remove a single plot area at runtime without requiring a server restart or manual editing of PlotSquared's configuration.

The command:

- Is implemented as a subcommand of `/plot area`.
- Only operates on partial areas containing the single `1;1` plot.
- Refuses to remove an area while it contains claimed plot data.
- Requires confirmation before removal.
- Revalidates the area after confirmation.
- Removes the area's entry from `worlds.yml`.
- Unregisters the area from the active `PlotAreaManager`.
- Does not modify or regenerate the existing Minecraft terrain.
- Uses PlotSquared's localized caption system for command messages rather than hardcoded locale-specific strings.

This keeps deletion narrowly scoped to single plot areas and leaves the existing broader `/plot area delete` behaviour unchanged.

### Submitter Checklist

- [x] I am opening from a topic branch and not my main branch.
- [x] I ensure that the pull request title represents the desired changelog entry.
- [x] New public fields and methods are annotated with `@since` TODO.
- [x] I read and followed the contribution guidelines.


PierreSchwang:
Without reviewing thoroughly, the command should be part of the "area" command (= subcommand), and localized messages should not be static but rather added to the locale file and referenced by a key (i18n)

tantaf123:
Small follow-up: I’ve updated the PR description with some additional implementation details and clarifications, including the safety/revalidation behaviour and scope of the command. Just mentioning it here in case the edited description was missed. Thanks!

tantaf123:
Hi! Just following up on this PR since it’s been a little while. I addressed the latest review feedback on 25 August, including the unsupported-area message, confirmation revalidation/null handling, and the locale grammar fix. As far as I can tell, there aren’t any outstanding changes left on my side. No rush, just wanted to check whether there’s anything else you’d like me to adjust before this can be reviewed further. Thanks!

PR #4924, opened 2026-08-19
Allow single plot areas in plot info by name

tantaf123:
## Overview

Single-plot areas internally use the plot ID `1;1`, which can make multiple named single areas difficult to distinguish when using `/plot info`.

This change allows named single-plot areas to be resolved by their area name when using `/plot info`.

For example, a single area named `apple_house` can be queried using:

`/plot info apple_house`

For named single-plot areas, the area name is also displayed instead of the generic `1;1` identifier in the unclaimed plot information message.

Normal plot IDs and aliases continue to use the existing resolution behaviour.

## Submitter Checklist
- [x] I am opening from a topic branch and not my main branch.
- [x] The pull request represents the desired changelog entry.
- [x] No new public fields or methods were added.
- [x] I read and followed the Contribution guidelines.

### DIFF
--- a/Core/src/main/java/com/plotsquared/core/command/Info.java
+++ b/Core/src/main/java/com/plotsquared/core/command/Info.java
@@ -18,6 +18,7 @@
  */
 package com.plotsquared.core.command;
 
+import com.plotsquared.core.PlotSquared;
 import com.plotsquared.core.configuration.Settings;
 import com.plotsquared.core.configuration.caption.Caption;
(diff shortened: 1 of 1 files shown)


MattBDev:
While I like the idea I wonder if this should be held off until v8 when @Citymonstret rewrites the database schema.

tantaf123:
That makes sense. My main goal here was to avoid changing the underlying
PlotId/database representation and only use the existing single-area name
as a human-facing identifier for /plot info. If v8 is the better place for
this though, I’m happy for it to wait.

PR #4930, opened 2026-09-01
fix: Prevent vine growth on the same block if `vine-grow` is `false`

CAG2Mark:
## Overview
Prevents vines from growing on the same block if the `vine-grow` flag is set to `false`. 

## Description
A single vine placed in a 1x1x1 hole can spread to all available faces in that hole, and this triggers `BlockGrowEvent` instead of `BlockSpreadEvent`. This PR checks for vine growth that is fired under `BlockGrowEvent`. Note that vines spreading to another block is still handled under `BlockSpreadEvent`, a subclass of `BlockGrowEvent`, and this is handled separately from `BlockFormEvent` (see line 415 in `BlockEventListener.java`).

### Submitter Checklist
  - Q: Do new listeners need this annotation?

### DIFF
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -410,6 +410,37 @@ public class BlockEventListener implements Listener {
     }
 
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onBlockGrow(BlockGrowEvent event) {
+        if (event instanceof BlockFormEvent) {
+            return; // handled below
+        }
+        Block block = event.getBlock();
+        Location location = BukkitUtil.adapt(block.getLocation());
+        if (location.isPlotRoad()) {
+            event.setCancelled(true);
+            return;
+        }
+        PlotArea area = location.getPlotArea();
+        if (area == null) {
+            return;
+        }
+        Plot plot = area.getOwnedPlot(location);
+        if (plot == null) {
+            return;
+        }
+        switch (event.getBlock().getType().toString()) {
+            case "VINE":
+                // Vines may grow on the same block if there are multiple available faces to grow on, for example
+                // when in an enclosed 1x1x1 hole, which triggers BlockGrowEvent instead of BlockSpreadEvent.
+                if (!plot.getFlag(VineGrowFlag.class)) {
+                    plot.debug("Vine could not grow because vine-grow = false");
+                    event.setCancelled(true);
+                }
+                break;
+        }
+    }
+
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
     public void onBlockForm(BlockFormEvent event) {
         if (event instanceof EntityBlockFormEvent) {
             return; // handled below
--- a/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
+++ b/Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
@@ -410,37 +410,6 @@ public class BlockEventListener implements Listener {
     }
 
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
-    public void onBlockGrow(BlockGrowEvent event) {
-        if (event instanceof BlockFormEvent) {
-            return; // handled below
-        }
-        Block block = event.getBlock();
-        Location location = BukkitUtil.adapt(block.getLocation());
-        if (location.isPlotRoad()) {
-            event.setCancelled(true);
-            return;
-        }
-        PlotArea area = location.getPlotArea();
-        if (area == null) {
-            return;
-        }
-        Plot plot = area.getOwnedPlot(location);
-        if (plot == null) {
-            return;
-        }
-        switch (event.getBlock().getType().toString()) {
-            case "VINE":
-                // Vines may grow on the same block if there are multiple available faces to grow on, for example
-                // when in an enclosed 1x1x1 hole, which triggers BlockGrowEvent instead of BlockSpreadEvent.
-                if (!plot.getFlag(VineGrowFlag.class)) {
-                    plot.debug("Vine could not grow because vine-grow = false");
-                    event.setCancelled(true);
-                }
-                break;
-        }
-    }
-
-    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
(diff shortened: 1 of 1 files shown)


Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -409,6 +409,37 @@ public void onCauldronEmpty(CauldronLevelChangeEvent event) {
         }
     }
 
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onBlockGrow(BlockGrowEvent event) {
+        if (event instanceof BlockFormEvent) {
+            return; // handled below
```
PierreSchwang: I would be fine with handling all growth types of vines under this event handler and removing the vine checks from `BlockSpreadEvent`. Feels cleaner than keeping vine spread logic in multiple places imo

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -409,6 +409,37 @@ public void onCauldronEmpty(CauldronLevelChangeEvent event) {
...
+        }
+        PlotArea area = location.getPlotArea();
+        if (area == null) {
+            return;
+        }
+        Plot plot = area.getOwnedPlot(location);
+        if (plot == null) {
+            return;
+        }
+        switch (event.getBlock().getType().toString()) {
```
PierreSchwang: If we only handle vines in this handler we could potentially think about moving the material check to the top (we don't really do that currently, but that checks is most certainly cheaper than getting the plot area and plot - especially for such a potentially high frequency event)

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -409,6 +409,37 @@ public void onCauldronEmpty(CauldronLevelChangeEvent event) {
...
+            case "VINE":
```
PierreSchwang: The current handler for the VineGrowthFlag handles more than just vines, so if the other handler gets removed (1st comment) this should contain all types (currently this makes sense as no other type should be able to grow in a single block space)

PierreSchwang: And is there any reason for using the enum name instead of the field? (would only make sense if there were any material renames in our specific supported version range)

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
```diff
@@ -409,6 +409,37 @@ public void onCauldronEmpty(CauldronLevelChangeEvent event) {
...
 
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
+    public void onBlockGrow(BlockGrowEvent event) {
+        if (event instanceof BlockFormEvent) {
+            return; // handled below
+        }
+        Block block = event.getBlock();
+        Location location = BukkitUtil.adapt(block.getLocation());
+        if (location.isPlotRoad()) {
+            event.setCancelled(true);
```
PierreSchwang: This should check road-flags, no? Unsure in this case tbh, as the default flag value is true I think, which would be unexpected for roads...

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: I don't know. I just followed what was done in the other handlers for consistency.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: All the other related handlers seem to do the same check as this, so I don't know.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: In principle, this check could be moved to the top for the other handlers too, but it might be a bit less clean

CAG2Mark:
Ah, I didn't see there is already a handler for `BlockGrowEvent`, oops...

EDIT: It seems `CropGrowFlag` prevents the described vine growth behaviour, see `onGrow`. I think this would count as a bug. Perhaps we should just rework `onGrow`?

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
MattBDev: Enum names were used due to compatibility issues with older versions of Minecraft. Given vines have been around for over 10 years, we don't need to use the string version of the enum.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: It seems this doesn't work. The BlockSpreadEvent doesn't appear to trigger handlers for BlockGrowEvent (despite the former being the latter's superclass). Vines would spread to other blocks in my testing when only handling BlockGrowEvent.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: No longer relevant as I've moved the code to `onGrow`, which was already used for crop handling.

Bukkit/src/main/java/com/plotsquared/bukkit/listener/BlockEventListener.java
CAG2Mark: No longer relevant as per my comment here: https://github.com/IntellectualSites/PlotSquared/pull/4930#discussion_r3940940044

CAG2Mark: Resolved in https://github.com/IntellectualSites/PlotSquared/pull/4930/changes/c00c1b9251cb3e7cd26a7e6fe10682f439554b60.

CAG2Mark:
Please see my comments as they contain some information. Still not sure what to do about road flags (and it seems many of the other handlers don't deal with this anyway, or maybe it is implicitly handled somehow, not sure)

